PR #456, opened 2020-04-07
map viewer with invalid map MA-698

pecholleyc:


frontend/src/components/explorer/MapViewer.vue
```diff
@@ -350,7 +339,7 @@ export default {
       }
 
       if (!this.checkValidRequest(this.currentDisplayedType, this.currentDisplayedName)) {
-        this.showMessage(messages.mapNotFound, 'info');
+        this.showMessage(`Invalid map ID "${this.currentDisplayedName}"`, 'info');
```
mihai-sysbio, line 342: For consistency, this message should use the same `danger` class as above, in which case there should be no need for `loadErrorTypeMessage`.

PR #460, opened 2020-04-07
widen content frame until widescren MA-772 

mihai-sysbio:
@e0 is this more in line with what you had in mind?


mihai-sysbio:
That's right, but that's the way Bulma behaves. I had already adjusted the `margin-left` for the logo, which I was reluctant to do because it would go against the default Bulma. Since it's already done on side, I've done the same for `margin-right` so it should be fine now.

PR #481, opened 2020-04-28
wrap neo4j queries in API container with expressjs MA-797

e0:


api/src/neo4j/driver.js
```diff
@@ -0,0 +1,12 @@
+import neo4j from 'neo4j-driver'
+
+// TODO: put uri in an environment variable
```
mihai-sysbio, line 3: This could be a checklist item on the Jira card, to improve time estimation.

frontend/src/store/modules/genes.js
```diff
@@ -5,13 +5,22 @@ const data = {
...
 
 const getters = {
-  geneName: state => state.gene.geneName,
+  geneName: state => state.gene.id,
 };
 
 const actions = {
   async getGeneData({ commit }, { model, id }) {
-    const gene = await genesApi.fetchGeneData(model, id);
+    console.warn(`TODO: use model: ${model}`);
```
mihai-sysbio, line 13: Because of linter settings, `console.log` throws an error in production mode so the Docker container doesn't build. This is allowed locally, but extra care should be given to remove them.

utils/mockData/.gitignore
```diff
@@ -0,0 +1,23 @@
...
+*.pid.lock
+
+# Dependency directories
+node_modules/
+
+# Yarn Integrity file
+.yarn-integrity
+
+# Mock data
+data/
```
mihai-sysbio, line 23: Most of these are already in the global `.gitignore` - it would be easier to add there the few missing ones.

mihai-sysbio:
Also, it would be nice to add a linter to the new `api`.

api/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM node:lts-slim
```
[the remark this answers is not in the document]
e0: Is there any preference for the alpine version? I just went with the same one that we had in frontend.

frontend/src/store/modules/genes.js
e0: Have added a Jira task about this now. Technically this frontend stuff shouldn't be in this branch and PR but I think we can ignore this for now.

e0:
@mihai-sysbio Thanks for the great feedback. I have added a few commits to address some of the issues. Please take another look when you have time.

api/Dockerfile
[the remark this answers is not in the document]
mihai-sysbio: It used to be that the `alpine` version was smaller. With less code implying a smaller attack surface, it was seen as more secure. With time, there have been some stability issues when people are using that with some `npm` packages but I don't think we are. So I'd wes should aim for `alpine` unless things break.

PR #488, opened 2020-06-10
refactor: separate networks between docker containers

mihai-sysbio:


e0:
I think this is a good idea. Is the intention to have a separate network for things that should not be accessible from outside? If that's the case I might personally prefer terminology like `public-network` and `private-network`.

mihai-sysbio:
It's more about separating which containers can talk to each other (between themselves), essentially having multiple `private-networks`.

e0:
Ah ok, I think I was thinking of consolidating the naming of the networks and maybe we can avoid the overwriting in the local docker compose file.

mihai-sysbio:
The `frontend-network` in `-local` needs to be added since it only exists in the local setup, ie not in production. However, it should work just fine to add it in the base config under `networks`, even though it's used only in `-local`.

e0:
Yea sorry, by overwriting I was thinking of the list of networks. I think I might still prefer something like a `public-network` for everything end-user facing and we can still have multiple private ones if needed. I think it is a bit easier to understand and also avoids adding a new network in the local config. That is just a personal preference though. Either way is fine.

PR #491, opened 2020-07-07
feat: new deployment setup including neo4j MA-807 MA-857

e0:


nginx/conf.d/nginx.conf
```diff
@@ -1,31 +1,22 @@
-worker_processes 4;
-events { worker_connections 1024; }
```
mihai-sysbio, line 2: The `worker_processes` should be defined, as the [default is just `1`](http://nginx.org/en/docs/ngx_core_module.html#worker_processes).
We might want the `worker_connections` to also be higher than the [default `512`](http://nginx.org/en/docs/ngx_core_module.html#worker_connections).

mihai-sysbio:
Here are some other notes:
- it feels the creation of the `nginx/conf.d` folder is not entirely necessary, and could be dealt with by changing the COPY command in the dockerfile
- in the refactoring of the nginx configuration, the IP filtering was lost. Of course, we don't need that for the `production` instance, but the Jenkinsfile was relying on that in order to set some filtering rules in the nginx configuration for the internal development instance.

api/src/index.js
```diff
@@ -1,8 +1,12 @@
 import express from 'express';
 import router from './endpoints/index';
+import { initializeSearchIndex } from './neo4j';
 
 const app = express();
 
+// runs only when index is not created
+(async() => initializeSearchIndex())();
```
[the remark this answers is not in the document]
e0: I agree, we already have a Jira issue for using the `cypher-shell` binary during the build stage. I have added your point as a checklist item for that issue now.

nginx/conf.d/nginx.conf
e0: Yes, I need to look more into this as well. I had these initially but it didn't work out of the box with the nginx-ctrtbot image we are using. Should be no problem to append these lines later though.

e0:
@mihai-sysbio These are great points. Regarding the first one this folder setup convention works well by default with the `nginx-certbot` docker image. I actually tried using `COPY` but didn't get that to work. It should be possible though.

Overall, as we discussed before, this PR is the minimal working setup for us to successfully deploy. I personally would prefer it if we can merge and close this accompanying Jira issue as soon as possible and open up new issues for each of these points. What do you think?

nginx/conf.d/nginx.conf
mihai-sysbio: I see, I've added a reminder for this in MA-851.

mihai-sysbio:
Good idea, all of the above points have been copied as checklist items in MA-851. I opted for keeping the old nginx conf around, to make it easier when continuing the refactoring work.

PR #496, opened 2020-07-17
feat: use cypher-shell to import data and add one-step deploy MA-859

e0:
This PR adds one-step build/deploy, including data import, for both the dev and production environments. This depends on setting the local file paths in `./env` file to the data files and data generator. The command `build-stack` or `deploy-stack <CONTEXT>`runs the data generator, which outputs data that is used by `cypher-shell` to populate the neo4j database during build.

As long as the paths to these folders are set, this setup should work for all members of the MetabolicAtlas organization. We can also make the two repos public so anyone can access them. Alternatively, we could host a copy of the import data publicly and add an option to use this data instead of running the data generator. Since this still requires publicly hosting the data, albeit in a different format, I think it would be preferable to make repos public. There is also a bit of extra maintenance to make sure the latest version of the data is available and could add more complexity in the future if multiple versions of data would be used. On the other hand, the advantage of publicly hosting the data makes the dev setup more independent and does not require one to have `node` and `yarn`. Please add your input on this when you have time.


docker-compose.yml
```diff
@@ -11,6 +11,13 @@ services:
     environment:
       - NEO4J_AUTH=${NEO4J_USERNAME}/${NEO4J_PASSWORD}
       - NEO4JLABS_PLUGINS=["apoc"]
```
mihai-sysbio, line 13: How about [referencing the `.env` file directly](https://docs.docker.com/compose/compose-file/#env_file)? There could potentially be distinct `local.env` and `prod.env`.

neo4j/Dockerfile
```diff
@@ -1,5 +1,10 @@
 FROM neo4j:4.1.0
 
-COPY ./plugins /plugins
-RUN rm -r /var/lib/neo4j/data
-COPY ./data /var/lib/neo4j/data
+ARG NEO4J_USERNAME
+ARG NEO4J_PASSWORD
```
mihai-sysbio, line 4: Are these needed? From what I'm seeing they're defined as env vars instead.

docker-compose-prod.yml
```diff
@@ -18,15 +18,9 @@ services:
...
-    command: neo4j
-    ports:
-      - "7687:7687"
 
   api:
     restart: unless-stopped
-    container_name: api
     build:
       context: ./api
       dockerfile: ./Dockerfile.prod
```
mihai-sysbio, line 26: Unless there are plans for different dockerfiles, maybe this can be moved to the base `docker-compose` .

docker-compose.yml
e0: The docker and docker-compose terminology for this is actually quite confusing. The variables coming from the `env_file` option are only for run-time so they would not be available in either docker-compose file or Dockerfile. Instead you have to use the `--env-file` command line option for them to be visible in the docker-compose file.

We have planned using different env files in MA-851.

neo4j/Dockerfile
e0: Continuing from above, providing the `--env-file` still does not make them available in the Dockerfile. Instead, one needs to defined the `ARGS`, which are only available during build (Dockerfile), see https://docs.docker.com/compose/compose-file/#args.

PR #498, opened 2020-07-21
refactor: production deployment MA-855

mihai-sysbio:


proj.sh
```diff
@@ -3,54 +3,51 @@ export PATH=$PATH:/usr/local/bin
...
 
 function deploy-stack {
   generate-data
   docker-compose -f docker-compose.yml -f docker-compose-prod.yml --context $1 up -d --build
 }
 
+function import-db {
+  generate-data --drop-indexes
+  source .env
+  docker exec -it neo4j bash -c "cypher-shell -u ${NEO4J_USERNAME} -p ${NEO4J_PASSWORD} --format plain 'MATCH (n) DETACH DELETE n;'"
```
e0, line 40: These cypher statements could be added to the data-generator as well, when `--drop-indexes` (which probably can be renamed to `--reset-db`) is provided.

mihai-sysbio: Great idea!

PR #501, opened 2020-07-27
fix: misc frontend issues MA-850

e0:


frontend/src/store/modules/models.js
```diff
@@ -10,6 +10,8 @@ const getters = {
     const modifiedModel = {
       ...model,
       email: model.authors[0].email,
+      apiName: model.short_name.split('-').map(s => s[0] + s.slice(1).toLowerCase()).join(''),
+      apiVersion: model.version.split('.').join('_'),
```
mihai-sysbio, line 14: Would there be a way to use some values from the db without applying changes in the frontend?

e0: For the version name we could map this in the backend instead. Or just use `Human-GEM` as the labels during import. For the version number, I'm not sure but I think dots handling is a little weird. I personally think mapping it at one place on the frontend is quite ok, at least for now.

PR #503, opened 2020-07-28
feat: add mouse gem MA-840

pecholleyc:
remove database_name in csv, change its value using short_name. Ideally the database_name field should be completely removed from the model objects


mihai-sysbio:
Would this be a good time to revise the definition of the models in the `api/src/neo4j` search ?
https://github.com/MetabolicAtlas/MetabolicAtlas/blob/feat/Mouse-GEM_MA-840/api/src/neo4j/queries/search.js#L258-L261

e0:
> Would this be a good time to revise the definition of the models in the `api/src/neo4j` search ?
> https://github.com/MetabolicAtlas/MetabolicAtlas/blob/feat/Mouse-GEM_MA-840/api/src/neo4j/queries/search.js#L258-L261

Sounds like a good idea. These lines were added before the models json existed in the backend. We can probably just reused the json here. Could also reuse the mapping logic from earlier with the `apiName`.

PR #506, opened 2020-09-01
neo4j in production MA-851

mihai-sysbio:


e0 (commented):
Looks good I think. The only thing is I’m worried that the cpu limits add up to be exactly 2.0. That’s all we currently have on the prod machine right? Wondering if we should make sure to reserve some for the rest of the system.

mihai-sysbio:
Yeah that was my intention - in the worse case scenario that one service does not take over. There aren't (m)any other processes running on this machine, and I believe the ftp resources won't be maxed out at the same time as the rest.
Where would you like to remove CPU allocation from?

e0:
I think it is still safer to reserve a bit for things like `ssh` and deployment. How about having the current `reservations` values for `limits` (and update the `reservation` values too accordingly) instead for `nginx`, `api`, and `neo4j`? That guarantees 0.2 for the rest of the system.

PR #507, opened 2020-09-01
add model to cypher queries and API MA-870

e0:


mihai-sysbio:
I didn't realise #508 was on top of this PR, please see my comments there. We should continue the discussion here.

api/src/endpoints/neo4j.js
```diff
@@ -12,65 +12,48 @@ import {
...
 
-neo4jRoutes.get('/:version/compartments/:id', async (req, res) => fetchWith(req, res, getCompartment));
-neo4jRoutes.get('/:version/compartments/:id/related-reactions', async (req, res) => fetchWith(req, res, getRelatedReactionsForCompartment));
+neo4jRoutes.get('/compartments/:id', async (req, res) => fetchWith(req, res, getCompartment));
+neo4jRoutes.get('/compartments/:id/related-reactions', async (req, res) => fetchWith(req, res, getRelatedReactionsForCompartment));
 
-neo4jRoutes.get('/:version/genes/:id', async (req, res) => fetchWith(req, res, getGene));
-neo4jRoutes.get('/:version/genes/:id/related-reactions', async (req, res) => fetchWith(req, res, getRelatedReactionsForGene));
+neo4jRoutes.get('/genes/:id', async (req, res) => fetchWith(req, res, getGene));
+neo4jRoutes.get('/genes/:id/related-reactions', async (req, res) => fetchWith(req, res, getRelatedReactionsForGene));
```
e0, line 39: For https://github.com/MetabolicAtlas/MetabolicAtlas/pull/508#discussion_r481796918: 

All of the endpoints would look similar in terms of query params. In this case it would be `/genes/:id/related-reactions?model=HumanGem`

While query params may not be the best choice for some endpoints for now, I'm still highly in favour of them for the the consistency and flexibility for future use cases.

api/src/neo4j/queries/compartment.js
```diff
@@ -1,40 +1,42 @@
 import querySingleResult from 'neo4j/queryHandlers/single';
 
-const getCompartment = async ({ id, version, full }) => {
-  const v = version;
+const getCompartment = async ({ id, model, version, full }) => {
+  const m = model ? `:${model}` : '';
+  const v = version ? `:V${version}` : '';
```
e0, line 5: For https://github.com/MetabolicAtlas/MetabolicAtlas/pull/508#discussion_r481799695:

Good point, it was bothering me too but I was thinking it's nice to clearly see exactly how their values in every function. But I don't think that should weigh over the advantage of having it defined in a single place. Will try to add a utility function for this and see how that feels.

api/src/neo4j/index.js
```diff
@@ -34,4 +32,5 @@ export {
   getInteractionPartners,
   getMapsListing,
   mapSearch,
+  search,
```
e0, line 35: For https://github.com/MetabolicAtlas/MetabolicAtlas/pull/508#discussion_r481805539:

I actually renamed the original `search` function into `_search` and added a new `search` function, which is: `const search = async (params) => params.model ? modelSearch(params) : globalSearch(params);
`. It helped to clean things up in the `api/src/endpoints/neo4j.js` file. So no change in functionality here, and the results are still different depending on if a model is provided.

api/src/endpoints/neo4j.js
mihai-sysbio: Sounds good!

mihai-sysbio (approved):

PR #508, opened 2020-09-01
remove transport reaction labels MA-874

e0:


mihai-sysbio (commented):
For most queries, having the model and version as params looks more powerful - it enables us to use the same API endpoint across multiple models. However, there are some queries where it looks potentially confusing, eg the `/relatedX` queries.

mihai-sysbio:
Neat! I'm curious to see if any of the query take longer now.

PR #509, opened 2020-09-02
gem search focus MA-785

e0:


frontend/src/components/explorer/gemBrowser/GemSearch.vue
```diff
@@ -112,7 +112,11 @@ export default {
     },
   },
   mounted() {
-    $('#search').focus();
+    if (this.searchTermString.length === 0) {
+      $('#search').focus();
+    } else {
+      this.blur();
+    }
```
mihai-sysbio, line 119: I think this lines might not be needed.

e0: The behaviour for this one is a bit weird, during testing, this was needed for when the search field is already in focus and the route changes to another page within the gem browser but only sometimes. There is probably a better way to solve the problem but that would require some more changes to how the search field works. So this was the fastest way to guarantee a good end-user experience without having to do too much refactoring at the moment.

PR #513, opened 2020-09-29
update the style of the map viewer layout MA-385 

mihai-sysbio:
In this PR the aim is to refactor/drastically improve the layout of the Map Viewer on both mobile and desktop views. Any remaining bugs, eg the 2D map not showing up on mobile, should be reported in MA-897 and resolved separately.

### DIFF
--- a/frontend/src/components/explorer/MapViewer.vue
+++ b/frontend/src/components/explorer/MapViewer.vue
@@ -27,7 +27,7 @@
              @click="$store.dispatch('maps/toggleDataOverlayPanelVisible')">
             {{ dataOverlayPanelVisible ? 'Hide' : 'Show' }} data overlay
           </a>
-          <MapsListing v-if="showingMapListing" :maps-listing="mapsListing" />
+          <MapsListing v-if="showingMapListing" />
         </div>
         <div v-if="!currentMap"
              class="column is-unselectable om-1 fixed-height-mobile">
@@ -121,9 +121,9 @@ export default {
       model: state => state.models.model,
       showing2D: state => state.maps.showing2D,
       dataOverlayPanelVisible: state => state.maps.dataOverlayPanelVisible,
+      mapsListing: state => state.maps.mapsListing,
     }),
     ...mapGetters({
-      mapsListing: 'maps/mapsListing',
       queryParams: 'maps/queryParams',
     }),
   },
--- a/frontend/src/components/explorer/MapViewer.vue
+++ b/frontend/src/components/explorer/MapViewer.vue
@@ -172,7 +172,7 @@ export default {
       return showing2D ? '2d' : '3d';
     },
     handleQueryParamsWatch(newQuery, oldQuery) {
-      if (!this.$route.params.map_id) {
+      if (newQuery && !this.$route.params.map_id) {
         const payload = [{}, null, `${this.$route.path}?dim=${newQuery.dim}`];
         history.replaceState(...payload); // eslint-disable-line no-restricted-globals
         return;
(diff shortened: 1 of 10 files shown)


frontend/src/components/explorer/MapViewer.vue
```diff
@@ -223,15 +231,15 @@ export default {
...
-    padding: 10px;
-    border-radius: 5px;
-    background: rgba(22, 22, 22, 0.8);
-  }
+      position: absolute;
+      z-index: 10;
+      padding: 10px;
+      border-radius: 5px;
+      background: rgba(22, 22, 22, 0.8);
+    }
```
e0, line 242: These indents seem like they are not needed.

frontend/src/components/explorer/MapViewer.vue
```diff
@@ -243,13 +251,46 @@ export default {
       }
     }
   }
+
+  @media screen and (max-width: $tablet) {
+    &.ordered-mobile {
+      display: flex;
+      flex-flow: column;
+    }
+    .om-1 {
```
e0, line 260: Just curious, what does the `om` stand for?

frontend/src/style/vars.scss
```diff
@@ -25,6 +25,6 @@ $grey: hsl(0, 0%, 48%);
 $grey-lighter: hsl(0, 0%, 86%);
 $black: hsl(0, 0%, 4%);
 
-// Custom variables for fullheight Map Viewer
-$navbar-height: 2.5rem;
+// Custom variables for fullheight Map Viewer, same as Bulma defaults
```
e0, line 28: Can we these lines if they are the same as Bulma defaults?

e0 (approved):
Have added some notes in https://chalmersysbiocsb.atlassian.net/browse/MA-897.

frontend/src/components/explorer/MapViewer.vue
```diff
@@ -161,6 +168,9 @@ export default {
     this.loadMapFromParams();
   },
   methods: {
+    dimensionalState(showing2D) {
+      return showing2D ? '2d' : '3d';
+    },
```
e0, line 173: This is a good refactor and I think we could even go one step further and have it in a Vuex getter. Then we wouldn't need to pass it as a param. Not necessary now though.

frontend/src/components/explorer/MapViewer.vue
```diff
@@ -11,24 +11,31 @@
...
+          <div id="dimensionToggle" class="buttons has-addons is-centered padding-mobile"
+               :title="`Switch to ${dimensionalState(!showing2D) }`"
+               @click="$store.dispatch('maps/toggleShowing2D')" tooltip="haha">
+            <button v-for="dim in [true, false]" :key="dim"
+                    class="button"
+                    :class="dim === showing2D ? 'is-selected is-primary has-text-weight-bold' : 'is-light'">
+              <span v-if="dim === showing2D" class="icon">
+                <i class="fa fa-check-square-o"></i>
+              </span>
+              <span v-if="dim !== showing2D">Switch to&nbsp;</span>
```
e0, line 23: This sometimes causes buttons to be in two rows but I think it's negligible for now.

mihai-sysbio: I think it's the `min-width` causing that, but removing that makes the text go outside the borders of the buttons; indeed, negligible for now.

e0: Yes I meant specifically the change in text. It might fit without the “Switch to” addition.

mihai-sysbio:
I have to admit, I really like this new map menu, and I'm happy to have the oldest unresolved issue behind us.

frontend/src/components/explorer/MapViewer.vue
mihai-sysbio: I tried, but it had a `width` that had to be overridden, so it made no difference. I kept that size for `min-width` instead.

PR #515, opened 2020-10-08
expand 3D viewer integration MA-559

e0:
There is likely still a few bugs with the 3D viewer (as well as the 2D viewer) integration. During review, please make a note of any problems you see either here or on Jira. Since this feature has already been in development for a while, let's try to focus on fixing only severe bugs (if any) in this current branch so we can try to merge this PR as soon as possible.


frontend/src/components/explorer/mapViewer/SidebarDataPanels.vue
```diff
@@ -155,19 +155,16 @@ export default {
...
           { name: 'name' },
-          { name: 'model_name', display: 'Model&nbsp;name' },
           { name: 'formula' },
           { name: 'compartment' },
-          { name: 'synonyms', display: 'Synonym(s)' },
         ],
         gene: [
           { name: 'name', display: 'Gene&nbsp;name' },
-          { name: 'alternate_name', display: 'Alt&nbsp;name' },
-          { name: 'synonyms', display: 'Synonym(s)' },
```
mihai-sysbio, line 166: May I ask what was the rationale behind removing the synonyms?

frontend/src/components/explorer/mapViewer/ThreeDviewer.vue
```diff
@@ -52,30 +91,137 @@ export default {
...
       // document.getElementById('viewer').addEventListener('select', e => console.debug('selected', e.detail));
     },
-    renderNetwork() {
+    getElementIdAndType(element) {
+      if (element.group === 'r') {
+        return [element.id, 'reaction'];
+      } if (element.group === 'e') {
+        return [element.id.split('-')[0], 'gene'];
+      }
+      return [element.id.split('-')[0], 'metabolite'];
```
mihai-sysbio, line 106: I think the introduction of the compartmentalized metabolite has cleared the need for merging the metabolite id with `-neo4jid`. And I think we wouldn't need that for the genes. So I would expect to not have to use `split` over there.

frontend/src/components/explorer/mapViewer/ThreeDviewer.vue
```diff
@@ -52,30 +91,137 @@ export default {
...
+        lz: z,
+      };
+      this.$store.dispatch('maps/setCoords', payload);
     },
     resetNetwork() {
       const viewer = document.getElementById('viewer3d');
       viewer.innerHTML = '';
     },
+    zoomIn() {
+      console.log('zoom in');
```
mihai-sysbio, line 204: Console logs break the production build, but I guess this meant to be fixed immediately?

mihai-sysbio (approved):
I believe the changes in `MapViewer.vue` led to a broken layout on mobile. While I'm supportive of refactoring, I would encourage the refactoring to be complete - it feels like merging in these changes _introduces_ bugs.

frontend/src/components/explorer/mapViewer/SidebarDataPanels.vue
e0: Good catch, it was causing problems and I meant to fix this actually but forgot about it.

frontend/src/components/explorer/mapViewer/ThreeDviewer.vue
e0: Good catch!

frontend/src/components/explorer/mapViewer/ThreeDviewer.vue
e0: Yes, this is bad practice but considering that there are so many bugs left for the map viewer I think we can ignore it for now.

mihai-sysbio: Alright, I've checked and as you say this issue is already noted under MA-897.

e0:
> I believe the changes in `MapViewer.vue` led to a broken layout on mobile. While I'm supportive of refactoring, I would encourage the refactoring to be complete - it feels like merging in these changes _introduces_ bugs.

Great point! It was definitely bad to re-break the mobile layout. Since there is a few other points to fix I will try to fix this as well, at least to the point where it is showing up on mobile.

PR #521, opened 2020-10-23
Feature/side main metabolites

mihai-sysbio:
I believe the intention behind this PR was to add currency metabolites based on the 2D maps.


mihai-sysbio:
Here are the reasons to close this PR and discard this branch:
- researchers have expressed the preference to define a hard-coded list of currency metabolites, instead of relying on a per-reaction definition
- we have stopped relying on 2D maps for input
- the backend has been changed from Django to Neo4j

PR #522, opened 2020-10-23
Feature/map viewer improvements

mihai-sysbio:
The intention behind this PR must have been to come up with a landing page for the Map Viewer.


mihai-sysbio:
The need for a nice landing page for the Map Viewer is still there. Instead of using a heatmap, the idea is to use a top-level metabolic view.

PR #525, opened 2020-10-26
fix: add ec (for reaction) to fulltext search index MA-876

e0:


api/src/neo4j/queries/search.js
```diff
@@ -277,10 +277,14 @@ const modelSearch = async ({ searchTerm, model, version, limit }) => {
 const _search = async ({ searchTerm, model, version, limit, includeCounts }) => {
   const v = version ? `:V${version}` : '';
 
+  // the EC field for reaction could contain ":", which is a special character
+  // in this case th search term is modified to be escape and perform an exact match
+  const term = searchTerm.includes("EC:") ? `\\"${searchTerm}~\\"` : `${searchTerm}~`;
```
mihai-sysbio, line 282: `includes` and not `startsWith`?

e0: Might be more accurate but I wouldn't bother changing it unless you think any other field can contain this substring.

mihai-sysbio: `EC` on its own could be present in other strings, but with `:` it's extremely unlikely.

PR #526, opened 2020-10-27
feat: webserver performance MA-732 MA-764

mihai-sysbio:


nginx/prod.Dockerfile
```diff
@@ -5,7 +5,7 @@ ENV VUE_APP_FTP_SERVER=${VUE_APP_FTP_SERVER}
 ENV VUE_APP_SVGMAPURL=${VUE_APP_SVGMAPURL}
 WORKDIR /project
 COPY frontend .
-RUN yarn & yarn build
+RUN yarn && yarn build
```
e0, line 8: Oh, nice catch here!

nginx/prod.nginx.conf
```diff
@@ -1,11 +1,24 @@
 limit_req_zone $binary_remote_addr zone=api_limit:10m rate=25r/s;
-limit_req_zone $binary_remote_addr zone=nginx_limit:10m rate=30r/s;
+limit_req_zone $binary_remote_addr zone=nginx_limit:10m rate=40r/s;
 
 upstream api {
   least_conn;
   server api:8081 weight=10 max_fails=3 fail_timeout=30s;
 }
 
+proxy_cache_path /nginx-cache levels=1:2 keys_zone=my_cache:30m inactive=60m use_temp_path=off max_size=3g;
```
e0, line 9: Since we don't have this enabled for local environment, have you tested the cache? I'm wondering specifically if it can distinguish different query params for the same path.

e0 (commented):
I think it looks good overall but wanted left a question about something I'm unsure about before approving it.

nginx/prod.nginx.conf
```diff
@@ -1,11 +1,24 @@
...
+
+map $sent_http_content_type $expires {
+    default                 off;
+    text/html               7d;
+    text/css                7d;
+    text/js                 7d;
+    text/javascript         7d;
+    application/javascript  7d;
+    application/json        7d;
+    ~image/                 max;
```
e0, line 19: How come we don't have max for everything?

nginx/prod.nginx.conf
mihai-sysbio: After deploying this, I can see that the `random-components` does not get cached since new components are received with every request. Also, the `/nginx-cache` directory is empty.

nginx/prod.nginx.conf
mihai-sysbio: If we deploy a new release, the browser should request new resources. I think `js` and `css` have a hash in the name, so we could have `max` for these. But for `json`, I wouldn't cache it forever.

nginx/prod.nginx.conf
e0: Can you please elaborate on that? I'm not sure how new components are being received.

mihai-sysbio: It looks like the `random-component` calls are not cached by NGINX, since new data is received every time, meaning that the request is passed via the API to Neo4j.

nginx/prod.nginx.conf
e0: When we deploy a new release doesn't everything get reset?

mihai-sysbio: Yes, but if the URL to the js or css would stay the same, the browser might think it's the same (cached) resource. I'm not sure if  `Etag`s help with this. Anyway I've increased all non-json resources to `max`.

nginx/prod.nginx.conf
mihai-sysbio: Alright, the caching problems have been fixed - all requests are cached except for `random-components`.

PR #528, opened 2020-11-02
feat: menu bar search MA-907

e0:


frontend/src/components/Explorer.vue
```diff
@@ -111,21 +111,7 @@ export default {
...
-      await this.$store.dispatch('models/getModels');
-      let modelKey = Object.keys(this.models)[0];
-      if (this.$route.params && this.$route.params.model in this.models) {
-        modelKey = this.$route.params.model;
-      } else if (this.model) {
-        modelKey = this.model.short_name;
-        this.$router.replace({ params: { model: modelKey } });
-      }
-      this.$store.dispatch('models/selectModel', modelKey);
-    },
```
mihai-sysbio, line 128: I remember having to add these lines so that this component would work when entering directly from the url. How come this is not needed anymore?

e0: Yes, the search is in the menubar which means it exists on every page now so this would have been a duplicate.

frontend/src/App.vue
```diff
@@ -166,13 +178,32 @@ export default {
...
+    });
+  },
+  methods: {
+    handleBodyClick(e) {
+      const navbar = document.querySelector('#navbar');
+
+      if (!navbar.contains(e.target) && this.showGemSearch) {
+        this.showGemSearch = false;
+      }
+    },
```
mihai-sysbio, line 205: This function is essentially running on every click on the website, which we should try to avoid if possible.
The previous `GemSearch` used a different implementation
```
@focus="showResults = true"
@blur="blur()">
```
Could a similar implementation be used here?

mihai-sysbio (changes requested):

frontend/src/App.vue
e0: That's a bit harder to do now, for example there is a select now as well. Clicking on the HTML body except something is a common pattern though. Is there any specific scenarios where this would cause problems?

mihai-sysbio: As a programming principle, I think it is better to apply changes as locally as possible. Overloading all clicks on the website feels unnecessary, and it may have side-effects (although maybe not at the moment). I've figured out an alternative approach, have a look [here](https://github.com/MetabolicAtlas/MetabolicAtlas/compare/feat/menu-bar-search...feat/blur-menu-bar-search).

frontend/src/components/explorer/gemBrowser/GemSearch.vue
```diff
@@ -102,25 +112,41 @@ export default {
...
+  async created() {
+    await this.getIntegratedModelList();
+  },
   mounted() {
-    if (this.searchTermString.length === 0) {
-      $('#search').focus();
-    } else {
-      this.blur();
-    }
+    $('#search').focus();
```
mihai-sysbio, line 126: This focus doesn't seem to work at the moment, might it be because of the transition?

frontend/src/App.vue
e0: Although I agree with the principle in general this is still a common exception. We are not overriding any behaviors, it's rather that all clicks which do not fall into this if condition simply behave as they usually would.

I tried your new branch and it doesn't work if you click outside the search area right after the search shows. Currently you need to click either the select or the input first, and then click outside.

mihai-sysbio: That behaviour is because of the broken `focus()` I mentioned below.

frontend/src/components/explorer/gemBrowser/GemSearch.vue
e0: I'm looking into this now. Not sure what happened.

e0: I added a fix for it now. It must have been something I overlooked before so thanks for spotting it.

frontend/src/App.vue
e0: With the auto-focus working I see it works to hide it directly now with both clicking outside and ESC so that's good. But I notice that I can't click on any search results now, looks like maybe the blur is happening before the click.

Also the previous behavior was clicking outside the navbar hides it the search but the new behavior is clicking outside the search component. So even if someone clicks inside the navbar but outside the search it still hides the search. This is maybe an ok change but I feel like the previous one was more straightforward.

mihai-sysbio: The blur was indeed happening too soon. In the previous implementation of the GemSearch it was 200ms, now it's 100 and it seems to work fine.

Regarding the second point, I think the new implementation is in line with the inspiration from [Apple's website](https://apple.com).

e0: The blur seems a bit hit-or-miss for me now. Result still not going through sometimes, is it working for you 100%?

> Regarding the second point, I think the new implementation is in line with the inspiration from Apple's website.

Ok I can buy that. :)

mihai-sysbio: I've now increased the delay until blurring occurs to 200ms, I hope it's okay. For me 100ms was enough, but you're right it's better to be on the safe side.

mihai-sysbio:
Upon further testing, I spotted that changing the model from the `<select>` immediately changes the current model. This is visible on the page, but the url stays the same. I think it the selection should not update the current model shown in the navbar, and should use that selected model only for the search.

e0:
> Upon further testing, I spotted that changing the model from the `<select>` immediately changes the current model. This is visible on the page, but the url stays the same. I think the selection should not update the current model shown in the navbar, and should use that selected model only for the search.

That's a good catch. I was thinking that this behavior would work well for most of the cases but for the case for when you are on a component page -> toggle the search -> change the model -> dismiss the search, it actually produces some buggy behavior where a map link leads to a wrong URL, so that is definitely too serious to overlook.

mihai-sysbio (approved):

mihai-sysbio:
This is far better than the GEM search we had before, many thanks @e0 for the idea and the implementation.

PR #530, opened 2020-11-26
feat: gem comparison MA-841

e0:


mihai-sysbio:
The cells in the table seem to take space proportional with the content, so the cells that have 5 digits take a bit more than the ones with 4. This can make the table look misaligned.

frontend/src/components/shared/ComparisonMatrix.vue
```diff
@@ -0,0 +1,201 @@
...
+      colors: ['#eef6fc', '#fffbeb'],
+    };
+  },
+  computed: {
+    ...mapState({
+      comparisons: state => state.compare.comparisons,
+      modelList: state => state.models.modelList,
+      selectedCell: state => state.compare.selectedCell,
+    }),
+    mergedComparisons() {
```
mihai-sysbio, line 53: Just a note: in theory it would make sense to compute this in the `api` so that it would be cached, instead of having each client do this computation.

mihai-sysbio (changes requested):
The cells in the table seem to take space proportional with the content, so the cells that have 5 digits take a bit more than the ones with 4. This can make the table look misaligned.

The 3 digit values would look better if they would be centered, so I guess this should apply to the entire table.

I think it would be more visually appealing to split the page in 2 equal columns, ie the table on the left to take equal space compared to the card (`is-half`), to maintain symmetry.

e0:
> The cells in the table seem to take space proportional with the content, so the cells that have 5 digits take a bit more than the ones with 4. This can make the table look misaligned.
> 
> The 3 digit values would look better if they would be centered, so I guess this should apply to the entire table.
> 

Thanks! I think it makes sense to have a 50% width distribution for each number in the table cell and center all of the text.

> I think it would be more visually appealing to split the page in 2 equal columns, ie the table on the left to take equal space compared to the card (`is-half`), to maintain symmetry.

I actually had two column layout like that initially but it felt a bit odd to stretch out the table cells like that. Also I think it makes sense for the comparison details card to take as much width as possible so it doesn't look too disproportionate (in contrast to the table) in height due to the lists of unique components.

mihai-sysbio:
Another thing I noticed is that the interface requires interaction the first time. I'm thinking it would be clearer for the user if a default selection is provided, eg Human-Gem vs Mouse-GEM. What do you think @e0 ?

e0:
> Another thing I noticed is that the interface requires interaction the first time. I'm thinking it would be clearer for the user if a default selection is provided, eg Human-Gem vs Mouse-GEM. What do you think @e0 ?

Sounds good!

mihai-sysbio (approved):

PR #531, opened 2020-12-11
feat: flag for force overwriting existing data file at build-stack

nanjiangshu:
It is a bit annoying to run `build-stack` and type yes-or-no for file overwriting hundred times when there are existing data files in the directory. I added a flag "-q" for `build-stack` function so that when this flag is enabled, existing data files will be overwritten by default.

### DIFF
--- a/proj.sh
+++ b/proj.sh
@@ -14,7 +14,7 @@ function generate-data {
 }
 
 function build-stack {
-  generate-data $@
+  generate-data
   docker-compose -f docker-compose.yml -f docker-compose-local.yml build $@
 }
 
--- a/proj.sh
+++ b/proj.sh
@@ -5,20 +5,12 @@ function generate-data {
   # enable flag "-q" to force overwritting existing data files
   echo 'Data generation started.'
   source .env && yarn --cwd $DATA_GENERATOR_PATH start $DATA_FILES_PATH "$@"
-  if [ "$(unalias cp 2> /dev/null)" != "" ];then 
-      unalias cp
-  fi
-
-  extra_opt=
-  if [ $1 ==  "-q" ]; then
-      extra_opt="--force"
-  fi
-  cp -r $extra_opt $DATA_GENERATOR_PATH/data/ neo4j/import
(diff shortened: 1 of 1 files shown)


e0:
@nanjiangshu The change looks fine I think. Could you provide some more details about your development environment? This is not a problem I had encountered before. Just wondering if there is an alternative solution that we should use as a standard instead, maybe it's a docker preference.

Since this is one of the first PRs, please use the same [format of commit messages](https://www.conventionalcommits.org/en/v1.0.0/) which we try to follow. You can also look at some previous commits for examples.

nanjiangshu:
Hi @e0 , 

I have tested `buid-stack` with Bash terminal on both Linux and MacOS.
On my Linux box : 
```
$ bash --version
GNU bash, version 4.4.20(1)-release (x86_64-pc-linux-gnu)
```

and on my Laptop with MacOS Catalina 
```
$ bash --version
GNU bash, version 5.0.2(1)-release (x86_64-apple-darwin18.2.0)
```
/bin/cp by default will ask if the target file to be copied exists.

Regarding format of the commits and pull requests, I've added a tag `feat:` for the message of this pull request. For commit messages, I will also add a tag in the future. Is there anything more that I have missed for the format guideline?

e0:
Thanks for the info. Can you also check what you get if you run `type cp`? The reason is that I just spotted that you have the `unalias cp` step and I'm wondering if your `cp` command is currently aliased to `cp -I`, which would ask you before overwriting.

nanjiangshu:
I think an alternative way (or maybe better) is to use `/bin/cp` instead of `cp`, in this case, we bypass the potential cp alias problem and do not need to mention it.

e0:
I think that would be better as well. Can you also check if the `--force` is necessary in that case? It might be enough with just the change to `/bin/cp`.

nanjiangshu:
changing to `/bin/cp` works on both Linux and Mac. `--force` is not needed. I've cleaned the changes and pushed the commit.

e0:
Thank you! I think we can omit the '$@' as well.

e0 (approved):

PR #533, opened 2020-12-16
feat: configurable port numbers MA-915

nanjiangshu:
enable configurable port numbers for containers in the local docker-compose setup

Related issue: MA-915


e0 (approved):
Looks great!

PR #535, opened 2021-01-08
feat: metabolite pics

mihai-sysbio:
This PR addresses the fetching of incorrect external ids, and adds a picture of the metabolite from ChEBI.

### DIFF
--- a/frontend/src/components/explorer/gemBrowser/Metabolite.vue
+++ b/frontend/src/components/explorer/gemBrowser/Metabolite.vue
@@ -23,7 +23,7 @@
         <loader v-if="showLoaderMessage" :message="showLoaderMessage" class="columns" />
         <template v-else>
           <div class="columns is-multiline metabolite-table is-variable is-8">
-            <div class="column is-10-widescreen is-9-desktop is-full-tablet">
+            <div class="column is-7-widescreen is-7-desktop is-full-tablet">
               <div class="table-container">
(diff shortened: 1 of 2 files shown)


frontend/src/components/explorer/gemBrowser/Metabolite.vue
```diff
@@ -70,6 +70,12 @@
               </div>
               <ExtIdTable :type="type" :external-dbs="metabolite.externalDbs"></ExtIdTable>
             </div>
+            <div v-if="metabolite && metabolite.externalDbs && metabolite.externalDbs.ChEBI"
```
e0, line 73: This is a nice addition but I think we could find a better placement for it. Suggestions are:

1. Inline with the table, perhaps between Formula and Charge.
2. In the sidebar at the top.

mihai-sysbio: In both cases, the image looked to small to be meaningful - it seemed more like a logo than an image that conveys useful information.

e0: Sounds reasonable. If we are keeping the image where it is I think we should consider making two changes.

1. Add a faint border/frame and a caption for the image. There is a lot of vertical whitespace right now.
2. If there is no chebi image, then use the original layout (table width). Since the table is always the same size now, the sidebar is not aligned to the right of the page.

PR #538, opened 2021-02-17
feat: api styling MA-866

nanjiangshu:
Completed features:

1. Click the logo re-direct to the frontend
2. Topbar has the same height as the frontend
3. /api point to the version 2 of API

Non-completed feature:
* Footbar not added. Swagger UI does not provide an easy way to change the footbar. One needs to modify the Swagger UI template in order to add a customized footbar. Maybe we can ignore this feature or postpone it.


mihai-sysbio:
A quick note: for this new API we had decided to introduce a version number so that whoever relies on the API can easily figure out there has been an update. Therefore, commit 938f79844e8c63891e2601d88cdbf1fba89c48f8 isn't necessary - we want to continue having `/api/v2` in the code.

nanjiangshu:
> /frontend/src/router.js



> > re-route /api to /api/v2
> 
> When navigating to `/api` nothing is shown. There should be a redirect/re-route from `/api` to `/api/v2`.
> 
> > when navigating to /api the frontend shouldn't be loaded only to redirect to /api/v2 -> do the redirect in nginx
> 
> When users go to `/api/v2` the redirect should not happen in the frontend, but it should rather happen in Nginx. This is partly achieved by:
> 
> https://github.com/MetabolicAtlas/MetabolicAtlas/blob/7471a0f52d4c22176c11104d7781eaaf96f37f0d/nginx/prod.nginx.conf#L68-L74
> 
> To finish the implementation, the redirect should be removed from the frontend:
> 
> https://github.com/MetabolicAtlas/MetabolicAtlas/blob/7471a0f52d4c22176c11104d7781eaaf96f37f0d/frontend/src/router.js#L42

I'm more confused than before after this explanation.
What should be the expected behavior when the users visit `<host>/api`? Should they expect `404` or redirected to the version 2 of API `<host>/api/v2` ?

In addition, suppose current users have used the API in their pipelines and after the change of API, they will get an error when accessing e.g. 

```$ curl -X GET "http://localhost/api/repository/integrated_models" -H  "accept: application/json"
```
How should they be informed that they actual need to use 

```
$ curl -X GET "http://localhost/api/v2/repository/integrated_models" -H  "accept: application/json"
```

instead

mihai-sysbio:
> > > re-route /api to /api/v2
> > 
> > When navigating to `/api` nothing is shown. There should be a redirect/re-route from `/api` to `/api/v2`.
>
> I'm more confused than before after this explanation.
> What should be the expected behavior when the users visit `<host>/api`? Should they expect `404` or redirected to the version 2 of API `<host>/api/v2` ?

Redirect to `/api/v2`.

---

> In addition, suppose current users have used the API in their pipelines and after the change of API, they will get an error when accessing e.g.

``` bash
$> curl -X GET "https://metatlas.csbi.chalmers.se/api/repository/integrated_models" -H  "accept: application/json"  
<!DOCTYPE html>
<html lang="en">
<head>
<meta charset="utf-8">
<title>Error</title>
</head>
<body>
<pre>Cannot GET /api/repository/integrated_models</pre>
</body>
</html>

```

This is the desired behavior. It breaks backwards compatibility, which is what the JSON response is doing anyway.

nanjiangshu:
Hi @mihai-sysbio, re-routing of `/api` -> `/api/v2` is now carried out in in nginx. Not 100 percent sure about the configuration in `prod.nginx.conf` since I can not test it.

mihai-sysbio:
The redirect seems to work good locally, and it should work just fine in production. In one of the files it was closely placed to the `/api/v2` route, which I thought was good, so I did the same for the other one (and fixed the indentation/spacing at the same time).

Next I gave the footer a shot by using the Bulma CDN and copy-pasting the footer text from the frontend. In order for the images (logos) be rendered, they had to not include the hash at the end (eg `logo.s0m3h4ash.png`), so I moved them to `frontend/public`. This trick was then applied to the main `logo`, so that it's not any more duplicated between `frontend` and `api`. With this, I think MA-866 is complete.

mihai-sysbio (approved):

nanjiangshu (commented):
It is great that the same logo at different places all link to the same file

nanjiangshu:
Looks great. I think we can merge it now.

PR #540, opened 2021-02-22
fix: on-click gene expression info icon to documentation MA-704

nanjiangshu:
- [x] direct to documentation by clicking the gene expression info icon

### DIFF
--- a/frontend/src/components/explorer/mapViewer/DataOverlay.vue
+++ b/frontend/src/components/explorer/mapViewer/DataOverlay.vue
@@ -6,8 +6,13 @@
     <div class="has-text-centered"
          title="Load a TSV file with gene IDs and TPM values.
          More information can be found in the documentation.">
-      Load custom gene expression data<router-link :to="{ name: 'documentation', hash: '#Data-overlay'}">
-        <span class="icon"><i class="fa fa-info-circle"></i></span> </router-link>
+      Load custom gene expression
+      <span style="white-space: nowrap;">
+        data
+        <router-link :to="{ name: 'documentation', hash: '#Data-overlay'}">
+          <span class="icon"><i class="fa fa-info-circle"></i></span>
+        </router-link>
+      </span>
     </div>
     <div class="file is-centered mb-2">
       <label class="file-label">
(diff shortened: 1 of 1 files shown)


mihai-sysbio (changes requested):
Good addition! I wouldn't have guessed that the blue color fits so well.

There is an indentation warning shown in `logs frontend`. We don't have automatic fixing of such cases, but it's on the roadmap MA-825.

When having a look at this on a half-screen window, the icon ended up alone on a next line. I think a <nobr> would "glue" it to the word before.

nanjiangshu:
> Good addition! I wouldn't have guessed that the blue color fits so well.
> 
> There is an indentation warning shown in `logs frontend`. We don't have automatic fixing of such cases, but it's on the roadmap [MA-825](https://chalmersysbiocsb.atlassian.net/projects/MA/issues/MA-825).
> 
> When having a look at this on a half-screen window, the icon ended up alone on a next line. I think a would "glue" it to the word before.

Hi, the indentation warning is fixed now. For the positioning of icon `i`, it moves as the layout of the webpage changes. Do you mean there is a way to bind the word `data` and the icon?

mihai-sysbio:
That's funny, the `<nobr>` tag  in my message was recognized by the markdown parser and it was hidden from the rendered html. Looking into this, [`<nobr>` is the best way to do this - `nowrap` should be used](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nobr). I tried this out in 15eef27630786203abe3ed4047d54e6638848836.

frontend/src/components/explorer/mapViewer/DataOverlay.vue
```diff
@@ -6,8 +6,13 @@
     <div class="has-text-centered"
          title="Load a TSV file with gene IDs and TPM values.
          More information can be found in the documentation.">
-      Load custom gene expression data<router-link :to="{ name: 'documentation', hash: '#Data-overlay'}">
-        <span class="icon"><i class="fa fa-info-circle"></i></span> </router-link>
+      Load custom gene expression
+      <span style="white-space: nowrap;">
```
nanjiangshu, line 10: nice solution!

nanjiangshu: I think it is OK to merge now.

PR #542, opened 2021-03-11
feat: resync doc MA-750

nanjiangshu:
All points in the check list of the card MA-750 are addressed.

* The search section has been added as a top layer section
* The sections GEM browser, Map viewer and Interaction partners have been re-written substantially.


mihai-sysbio:
We use the Font Awesome icon pack in this project like below. It would be good to continue using the icons in the same way.

https://github.com/MetabolicAtlas/MetabolicAtlas/blob/c83a49bf28666bb5831c1ed278c47fe486e93572/frontend/src/components/Home.vue#L34-L36

mihai-sysbio:
Do you have some good examples of websites embedding snapshots of buttons in the documentation?

nanjiangshu (commented):
Good changes! Better view with the more compact side table of contents.

e0 (approved):
Looks like there is some good discussion and progress here. I have taken a quick look at the branch and it looks like all of the checklist items from the Jira task are completed. I personally like the addition of images.

Unless there is any major problems, I suggest merging this PR. Any additional changes we would like could be collected and specified in a new Jira task.

nanjiangshu:
> @nanjiangshu do you have some good examples of websites embedding snapshots of buttons in the documentation?

Do you mean snapshots like this https://github.com/MetabolicAtlas/MetabolicAtlas/blob/feat/resync-doc-MA-750/frontend/public/img/button-export-tsv.png? They are all in this folder https://github.com/MetabolicAtlas/MetabolicAtlas/tree/feat/resync-doc-MA-750/frontend/public/img.

mihai-sysbio:
@nanjiangshu I was asking if you know of a reputable website (think of the scale Apple or UniProt) that do an excellent documentation. How do they approach the inclusion of images, and specifically screenshots of buttons, in the documentation?

e0:
That type of documentation might be difficult to search for based using keywords but here is an example from Apple documenting their Buttons design. https://developer.apple.com/design/human-interface-guidelines/sign-in-with-apple/overview/buttons/

nanjiangshu:
> @nanjiangshu I was asking if you know of a reputable website or two (think of the scale Apple or UniProt) that has an excellent documentation. How do they approach the inclusion of images, and specifically screenshots of buttons, in the documentation?

The official PDB website embeds images in a similar way. There are plenty of them, e.g. 
https://www.rcsb.org/docs/search/advanced-search/attribute-search

mihai-sysbio:
What I see from both examples is that neither adds images inline with text. If we are to follow the PDB example, we should revert back to the text-only description of buttons and include different screenshots:

image

The downside is that these more comprehensive screenshots are more prone to falling behind the website.

@nanjiangshu it's up to you: if you feel strongly about adding the bigger screenshots, please go ahead and add them. Make sure to use Google's Guetzli algorithm to compress the images before uploading them. Either way, please remove the screenshots of the buttons.

nanjiangshu:
I don't think it is that important with or without these images (although I personally think it increases the readability). The important thing is that I re-wrote some of the sentences so that the descriptions are easier to understand as a relatively new user, not someone who are very familiar with it. 
I think the fastest way to change this pull request so that it balances our opinions is to remove these images and describe these buttons with text, as used before.

mihai-sysbio (changes requested):
@nanjiangshu very nice changes overall, the documentation does indeed read better. I've made a few tweaks here and there so you didn't have to. I really liked the adding of figure captions, very professional.
I'm waiting for you to replace the pngs with text before I have a final read.

nanjiangshu:
Hi @mihai-sysbio, embedded images are removed now.

mihai-sysbio (approved):
I've had a more thorough read and made some further refinements. All looks good to me now - I think this is ready to be merged.

frontend/src/components/Documentation.vue
```diff
@@ -131,8 +131,8 @@
           <p>Similar to the 2D SVG maps, the 3D graphs are interactive. Clicking a node on the 3D grap, be it a metabolite, reaction, or gene, the information panel on the left sidebar will be updated for the clicked node. Additonal information of the selected node can be accessed by clicking the button "GEM Browser" under information panel on the left sidebar.</p>
 
           <h5 id="Data-overlay" class="is-size-5">Data overlay</h5>
-          <div class="columns m-0">
-            <div class="column p-0 is-8 content">
+          <div class="columns is-8">
```
nanjiangshu, line 128: Is the indentation of the paragraph "Data overlay" made by purpose or a side affect of increasing the gap between columns, see https://www.dropbox.com/s/mdk1pv78lisahz5/Screenshot%202021-03-16%20at%2022.10.29.png?dl=0

mihai-sysbio: Great that you spotted this @nanjiangshu. That extra indentation is a mistake, definitely not on purpose, but it might have to do will the changes that have increased gap between the columns.

mihai-sysbio: How does it look now after https://github.com/MetabolicAtlas/MetabolicAtlas/pull/542/commits/9dd8bd277f77d923e0237ad955bafbeb48db5d6d?

nanjiangshu: Looks great now! I will merge the branch.

PR #544, opened 2021-03-15
feat: error handler for all requests MA-578

e0:
Non-existent ids in gem browser, map viewer, and interaction partners now display an error message using the `NotFound` component.

An `ErrorPanel` component is added and used in:
- specific scenarios such as failing to apply RNA levels or exporting TSV file
- general, non 404, scenarios where a web request fails

### DIFF
--- a/api/src/neo4j/queryHandlers/single.js
+++ b/api/src/neo4j/queryHandlers/single.js
@@ -8,6 +8,11 @@ const querySingleResult = async (statement) => {
 
   try {
     const response = await session.run(statement);
+
+    if (response.records.length === 0) {
+      throw new Error('404');
+    }
+
     result = response.records[0].get(0);
 
     if (Object.values(result).flat().length === 0) {
--- a/api/src/neo4j/queryHandlers/single.js
+++ b/api/src/neo4j/queryHandlers/single.js
@@ -9,6 +9,10 @@ const querySingleResult = async (statement) => {
   try {
     const response = await session.run(statement);
     result = response.records[0].get(0);
+
+    if (Object.values(result).flat().length === 0) {
+      throw new Error('404');
+    }
   } catch (e) {
     error = e;
   } finally {
--- a/api/src/neo4j/queryHandlers/single.js
+++ b/api/src/neo4j/queryHandlers/single.js
@@ -15,7 +15,8 @@ const querySingleResult = async (statement) => {
 
     result = response.records[0].get(0);
 
-    if (Object.values(result).flat().length === 0) {
+    if (Object.values(result).flat().filter(x => x !== 0).length === 0) {
+      // the result contains only empty lists of 0 values
       throw new Error('404');
     }
   } catch (e) {
--- a/frontend/src/components/NotFound.vue
+++ b/frontend/src/components/NotFound.vue
@@ -13,7 +13,7 @@
             Probably there is a typo in the {{ type }} identifier in the URL
             <br>
             Use the
-            <span v-if="type === 'map'">{{ type }} listing</span>
+            <span v-if="type === 'map'">list of {{ type }}s</span>
             <span v-else>search bar above </span>
             to find other {{ type }}s
           </p>
--- a/frontend/src/components/NotFound.vue
+++ b/frontend/src/components/NotFound.vue
@@ -1,7 +1,6 @@
 <template>
   <div class="column has-text-centered is-three-fifths-desktop is-three-quarters-tablet is-fullwidth-mobile">
     <div class="box has-background-light content">
-      <template v-if="componentId">
         <template v-if="type">
           <p class="title is-size-5">
             <span class="is-capitalized">{{ type }}&nbsp;</span> "&nbsp;{{ componentId }}&nbsp;" &nbsp;not found
@@ -12,19 +11,14 @@
           <template v-else>
             <p>
               Probably there is a typo in the {{ type }} identifier in the URL
-              <br>Use the search bar above to find other {{ type }}s
+              <br>
+              Use the
+              <span v-if="type === 'map'">{{ type }} listing</span>
+              <span v-else>search bar above </span>
+              to find other {{ type }}s
             </p>
           </template>
         </template>
-        <template v-else>
-          <p class="title is-size-5">
-            <span class="is-capitalized">{{ componentId }}</span>&nbsp;" &nbsp;not found
-          </p>
-          <p>
-            Probably there is a typo in the identifier in the URL
-          </p>
-        </template>
-      </template>
       <template v-else>
         <h1 class="is-size-1 has-text-weight-bold">¯\_(ツ)_/¯<br>404</h1><br>
         <p class="is-size-5">
--- a/frontend/src/components/NotFound.vue
+++ b/frontend/src/components/NotFound.vue
@@ -1,17 +1,27 @@
 <template>
   <div class="column has-text-centered is-three-fifths-desktop is-three-quarters-tablet is-fullwidth-mobile">
     <div class="box has-background-light content">
-      <template v-if="type">
-        <p class="title is-size-5">
-          <span class="is-capitalized">{{ type }}&nbsp;</span> "&nbsp;{{ componentId }}&nbsp;" &nbsp;not found
-        </p>
-        <template v-if="type === 'model'">
-          <p>Visit the Explore page to select one of the integrated models</p>
+      <template v-if="componentId">
+        <template v-if="type">
+          <p class="title is-size-5">
+            <span class="is-capitalized">{{ type }}&nbsp;</span> "&nbsp;{{ componentId }}&nbsp;" &nbsp;not found
+          </p>
+          <template v-if="type === 'model'">
+            <p>Visit the Explore page to select one of the integrated models</p>
+          </template>
+          <template v-else>
+            <p>
+              Probably there is a typo in the {{ type }} identifier in the URL
+              <br>Use the search bar above to find other {{ type }}s
+            </p>
+          </template>
         </template>
         <template v-else>
+          <p class="title is-size-5">
+            <span class="is-capitalized">{{ componentId }}</span>&nbsp;" &nbsp;not found
+          </p>
           <p>
-            Probably there is a typo in the {{ type }} identifier in the URL
-            <br>Use the search bar above to find other {{ type }}s
+            Probably there is a typo in the identifier in the URL
           </p>
         </template>
       </template>
--- a/frontend/src/components/NotFound.vue
+++ b/frontend/src/components/NotFound.vue
@@ -5,19 +5,15 @@
         <p class="title is-size-5">
           <span class="is-capitalized">{{ type }}</span><code>{{ componentId }}</code>not found
         </p>
-        <template v-if="type === 'model'">
-          <p>Visit the Explore page to select one of the integrated models</p>
-        </template>
-        <template v-else>
-          <p>
-            Probably there is a typo in the {{ type }} identifier in the URL
-            <br>
-            Use the
-            <span v-if="type === 'map'">list of {{ type }}s</span>
-            <span v-else>search bar above </span>
-            to find other {{ type }}s
-          </p>
-        </template>
+        <p v-if="type === 'model'">{{ messages.modelNotFound }}</p>
+        <p v-else>
+          Probably there is a typo in the {{ type }} identifier in the URL
+          <br>
+          Use the
+          <span v-if="type === 'map'">list of {{ type }}s</span>
+          <span v-else>search bar above </span>
+          to find other {{ type }}s
+        </p>
       </template>
       <template v-else>
         <h1 class="is-size-1 has-text-weight-bold">¯\_(ツ)_/¯<br>404</h1><br>
@@ -36,11 +32,18 @@
 </template>
 <script>
 
+import { default as messages } from '@/helpers/messages';
+
 export default {
   name: 'NotFound',
   props: {
     type: String,
     componentId: String,
+  },
+  data() {
+    return {
+      messages,
+    };
   },
 };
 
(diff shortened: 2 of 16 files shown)


frontend/src/components/NotFound.vue
```diff
@@ -1,17 +1,27 @@
...
+          <template v-else>
+            <p>
+              Probably there is a typo in the {{ type }} identifier in the URL
+              <br>Use the search bar above to find other {{ type }}s
+            </p>
+          </template>
         </template>
         <template v-else>
+          <p class="title is-size-5">
+            <span class="is-capitalized">{{ componentId }}</span>&nbsp;" &nbsp;not found
```
nanjiangshu: Is the space between quotation mark and the componentId added by purpose? When you look at the rendered "not found" page, it seems there are trailing white spaces around the ID.

frontend/src/components/explorer/MapViewer.vue
```diff
@@ -89,6 +90,7 @@ import { debounce } from 'vue-debounce';
 import DataOverlay from '@/components/explorer/mapViewer/DataOverlay.vue';
 import ErrorPanel from '@/components/shared/ErrorPanel';
```
nanjiangshu, line 91: Not related to the pull request, but I'm curious what is the difference between importing with `.vue` and without.
e.g. 
import DataOverlay from '@/components/explorer/mapViewer/DataOverlay.vue'; 
  vs 
import ErrorPanel from '@/components/shared/ErrorPanel';

frontend/src/components/NotFound.vue
```diff
@@ -1,20 +1,24 @@
...
           </p>
+          <template v-if="type === 'model'">
+            <p>Visit the Explore page to select one of the integrated models</p>
+          </template>
+          <template v-else>
+            <p>
+              Probably there is a typo in the {{ type }} identifier in the URL
+              <br>
+              Use the
+              <span v-if="type === 'map'">{{ type }} listing</span>
```
nanjiangshu, line 16: For `map listing`, it will be even better to add a helper about how to use map listing.

nanjiangshu (changes requested):
Hi @shan-chalmers , the changes look great! I added a few comments for the display of "not found" page and asked for your feedback. However, I don't know in which condition a `404` error or a redirect to the `404` page will be activated. Could you please send me an example?

frontend/src/components/NotFound.vue
e0: Thanks for this comment. Since this is from an outdated commit, I will refer to [this line](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/544/files#diff-3a5b53f2754c9efb675a20e09d7f9cc439c90d32a80736950ba1ea75e23c6126R6) in the diff: the whitespaces are unchanged in this PR. I imagine that the whitespaces were initially added to avoid the confusion if the quotation marks could be a part of the ID. I did notice that the lines in this file are not indented properly though so I will try to push a fix for that soon.

frontend/src/components/explorer/MapViewer.vue
e0: Ah interesting. I'm not sure about it but apparently it seems to work with both, I imagine it also depends on the JavaScript syntax configuration one sets for the project. We should probably decide on a consensus about this.

frontend/src/components/NotFound.vue
e0: Since this error message is to be displayed right next to the map listing, I didn't think it would be necessary to add a helper. Do you have anything in mind on what would be a good helper message?

e0:
@nanjiangshu Thanks for the feedback. For `404`, the `NotFound` component should be presented for whenever a non-existent identifier is provided in the gem browser, map viewer, or interaction partners page. Here is an example of a non-existent reaction id: http://localhost/explore/Human-GEM/gem-browser/reaction/HMR_30091

frontend/src/helpers/messages.js
```diff
@@ -9,4 +9,5 @@ export default {
   mapViewerName: 'Map Viewer',
   gemBrowserName: 'GEM Browser',
   interPartName: 'Interaction Partners',
+  fileSaveError: 'There was a problem saving the file. Please try using a different browser or contact us.',
```
mihai-sysbio, line 12: Having seen this message and wanting to contact us, a user should not need to spend any time searching for the contact information. Some suggestions are to: include a link to the contact section in About, or include the email address.

frontend/src/components/shared/ErrorPanel.vue
```diff
@@ -0,0 +1,69 @@
...
+</script>
+
+<style>
+
+#errorPanel {
+  z-index: 11;
+  position: fixed;
+  width: 350px;
+  right: 20px;
+  bottom: 80px;
```
mihai-sysbio, line 54: This error only occurs in critical cases, so it should be more visible to the user. Instead of the corner of the page, could we have in the middle at the bottom, or in the center of the page?

frontend/src/helpers/messages.js
e0: This is based on the `unknownError` in `messages.js`. I agree that it is better to make it a email link, do you want this to be changed for all "contact us" messages in this PR?

frontend/src/components/shared/ErrorPanel.vue
e0: I think at the bottom is more suitable if we want to keep this animation. I don't feel too strongly about it bottom-center vs bottom-right so I can move it to the center.

mihai-sysbio: Bottom-center would make it more visible, thanks.

frontend/src/components/NotFound.vue
mihai-sysbio: Yes that's the explanation for the extra space. Any other ideas to do this? Code markup?

frontend/src/helpers/messages.js
mihai-sysbio: I'd prefer it this would be done for all messages, but it's up to you.

frontend/src/components/NotFound.vue
e0: I support using the `code` tag for this.

frontend/src/components/shared/ErrorPanel.vue
e0: A commit has been added to address this. It also includes the change to always display the error panel even for 404 responses.

frontend/src/components/NotFound.vue
nanjiangshu: I agree. using `code` tag might be better than adding extra white space.

mihai-sysbio: Let's go for `that`!

frontend/src/components/NotFound.vue
nanjiangshu: For the "map listing", I though it means the maplisting in the API. Not sure if other people will also be confused. 
Maybe change the message to "use the list on the left sidebar to find other maps.

e0:
@nanjiangshu @mihai-sysbio Thanks again for your review and feedback. I have added a few commits now, which should cover the requested changes.

mihai-sysbio:
I need a bit more time to look at this again, but overall yes all feedback has been incorporated.

nanjiangshu:
Hi @shan-chalmers, my previous points have all been solved nicely. However, I found a few new issues. 

1. The `not found` is activated for all other components but not `compartment`, e.g. 
http://localhost/explore/Human-GEM/gem-browser/compartment/golgi_apparatusd

2. When `not found` is activated, e.g. in the screenshot below, it pops up another warning message in the bottom. This feels a bit redundant for me (but maybe a nice feature for others)
https://www.dropbox.com/s/aqo1fvtvgwvlmvg/Screenshot%202021-03-19%20at%2010.15.41.png?dl=0 

3. When both the model and the component are not found, three warning message cards will be shown, e.g. when visiting the url http://localhost:9080/explore/Human1-GEM/gem-browser/subsystem/ami2no_sugar_and_nucleotide_sugar_metabolism. as shown in the screenshot (https://www.dropbox.com/s/k2m8hqe4uveqo6m/Screenshot%202021-03-19%20at%2010.46.56.png?dl=0) The top two boxes look a bit awkward since they are not aligned. If there is an easy fix, I think it is better to make them aligned. Otherwise, just leave as it is.  

4.  I'm not able to activate the error message coded in https://github.com/MetabolicAtlas/MetabolicAtlas/blob/feat/error-handler_MA-578/frontend/src/components/NotFound.vue#L8-L10. In which condition is `type == 'model'`?

e0:
@nanjiangshu Thanks for the detailed testing and feedback. I took the opportunity to refactor the `NotFound` component to further reduce reused boilerplate markup.

1. Fixed in https://github.com/MetabolicAtlas/MetabolicAtlas/pull/544/commits/e9ac7ea4249407e7b795a9c7bc28f40f4179af99.
2. This was intended. It is good that you brought it up and if we would like to have further discussion on this I suggest we open up a new Jira issue for it.
3. This has been addressed in the following way: the error message on the page (not the error panel) now only displays either the model error message or the component error message but **not both**.
4. Following the previous point, the model error message in `NotFound` component is now used instead of using the ones directly from the components.

nanjiangshu:
Hi @shan-chalmers, the error messages look much better now! The feature to show both the `not found` card and the `Opps` card is OK with me. Although it seems a bit redundant, but the webpage looks less empty with the addition of the `Opps` card in the bottom.

e0:
@nanjiangshu I'm inclined to agree with you. I think the argument for always showing the `ErrorPanel` is for consistency, so that it applies for all network errors. It is also a more general 404 message compared to the one shown in the `NotFound` component. But I'm not sure if this is enough of an argument to offset the redundancy of having both. @mihai-sysbio do you want to have another look at the [screenshot](https://www.dropbox.com/s/aqo1fvtvgwvlmvg/Screenshot%202021-03-19%20at%2010.15.41.png?dl=0) @nanjiangshu provided and add your opinion? After that I can try to make the change if necessary, or perhaps merge this PR if there is no other issues.

frontend/src/components/NotFound.vue
mihai-sysbio: It looks really neat, it's so much better than the trailing spaces. Thanks for bringing this up @nanjiangshu and for the fix @e0.

frontend/src/components/explorer/MapViewer.vue
mihai-sysbio: Good idea, let's take this separately.

frontend/src/components/NotFound.vue
mihai-sysbio: I though the map listing here was the API endpoint used for the Map Viewer. If so, I'm not able to trigger this alternative message:
```
<span v-if="type === 'map'">{{ type }} listing</span>
```
Could you share a broken url for this case?

api/src/neo4j/queryHandlers/single.js
```diff
@@ -8,7 +8,17 @@ const querySingleResult = async (statement) => {
...
   try {
     const response = await session.run(statement);
+
+    if (response.records.length === 0) {
+      throw new Error('404');
+    }
+
     result = response.records[0].get(0);
+
+    if (Object.values(result).flat().filter(x => x !== 0).length === 0) {
```
mihai-sysbio, line 18: Should we instead fail this request at source instead of allowing a broken response?

frontend/src/components/shared/ErrorPanel.vue
```diff
@@ -0,0 +1,69 @@
+<template>
+  <transition name="slide-fade">
+    <article v-if="message" id="errorPanel" class="message is-danger">
+      <div class="message-header">
+        <b>{{ title }}</b>
+        <button v-if="hasHideListener" class="delete" aria-label="delete" @click="$emit('hideErrorPanel')" />
+      </div>
+      <div class="message-body has-text-centered">
+        <h5 class="title is-6" v-html="message" />
```
mihai-sysbio, line 9: There's an asymmetric space above the error message, so I'm suggesting
```suggestion
        <p class="title is-6" v-html="message" />
```

frontend/src/components/shared/ErrorPanel.vue
mihai-sysbio: Nice!

mihai-sysbio (approved):
> @mihai-sysbio do you want to have another look at the screenshot @nanjiangshu provided and add your opinion? After that I can try to make the change if necessary, or perhaps merge this PR if there is no other issues.

I understand the feeling of redundancy this gives. However, I don't have great way to improve this at the moment, and I think the current solution is both acceptable and achieves the purpose, so I'm inclined to not ask for any changes on this.

Overall, this is a great addition to the website that has been missing from the very beginning. I'm going to approve this already, but I have provided some feedback.

api/src/neo4j/queryHandlers/single.js
e0: What do you mean by "at source"? In the cypher query?

frontend/src/components/shared/ErrorPanel.vue
e0: Great change and thanks for making it convenient with the "Commit suggestion" feature!

api/src/neo4j/queryHandlers/single.js
mihai-sysbio: Yes. Maybe we can discuss about this today, let's not drag out this PR any longer.

mihai-sysbio: For reference, there is a new card MA-932.

frontend/src/components/NotFound.vue
mihai-sysbio: Thanks for the demo @e0.

nanjiangshu (approved):

PR #546, opened 2021-03-30
feat: reimplement HPA endpoints MA-852

e0:


api/src/neo4j/queries/gene.js
```diff
@@ -46,5 +47,94 @@ RETURN apoc.map.mergeList(COLLECT(value.data)) as gene
   return { ...gene, externalDbs: reformatExternalDbs(gene.externalDbs) };
 };
 
+const getGenesForHPA = async () => {
+  const statement = `
+MATCH (g:Gene:HumanGem)-[:V1_3_0]-(r:Reaction)-[:V1_3_0]-(s:Subsystem)-[:V1_3_0]-(ss:SubsystemState)
```
mihai-sysbio, line 52: Even though the model is static (which is as expected), it would be highly desirable for the version to automatically take the latest. We should not need to remember to bump up the version here every time we update the model in `data-files`.
Note: this version is used in multiple places in this file.

api/src/neo4j/queries/gene.js
```diff
@@ -46,5 +47,94 @@ RETURN apoc.map.mergeList(COLLECT(value.data)) as gene
...
+MATCH (:Gene:HumanGem {id:'${id}'})-[:V1_3_0]-(r:Reaction)-[:V1_3_0]-(s:Subsystem)-[:V1_3_0]-(ss:SubsystemState)
+WITH DISTINCT s.id as sids, COLLECT(r.id) as rids, ss
+UNWIND sids as sid
+CALL apoc.cypher.run("
+  MATCH (:Subsystem {id: $sid})-[:V1_3_0]-(r:Reaction)-[:V1_3_0]-(cm:CompartmentalizedMetabolite)
+  USING JOIN ON r
+  WITH DISTINCT cm
+  MATCH (m:Metabolite)-[:V1_3_0]-(cm)-[:V1_3_0]-(c:Compartment)-[:V1_3_0]-(cs:CompartmentState)
+  USING JOIN ON c
+  RETURN DISTINCT { id: $sid, compartments: COLLECT(DISTINCT(cs.name)), model_metabolite_count: COUNT(DISTINCT(m.id)), compartment_metabolite_count: COUNT(DISTINCT(cm.id)) } as data
```
mihai-sysbio, line 73: It looks to me like there are a bit too many `DISTINCT` clauses here. For example, the `cm` should already be distinct on line 70, so maybe there is no need to run `DISTINCT` again before the count. Also, in terms of counts, I think graph nodes can be counted just fine, and might more efficient than counting the ids.

```suggestion
  RETURN DISTINCT { id: $sid, compartments: COLLECT(DISTINCT(cs.name)), model_metabolite_count: COUNT(DISTINCT(m)), compartment_metabolite_count: COUNT(DISTINCT(cm) } as data
```

api/src/neo4j/queries/gene.js
```diff
@@ -46,5 +47,94 @@ RETURN apoc.map.mergeList(COLLECT(value.data)) as gene
...
+  UNION
+
+  MATCH (:Subsystem {id: $sid})-[:V1_3_0]-(r:Reaction)
+  WHERE r.id IN $rids
+  RETURN DISTINCT { id: $sid, reactions_catalysed: COUNT(DISTINCT(r)) } as data
+
+  UNION
+
+  MATCH (:Subsystem {id: $sid})-[:V1_3_0]-(ssvg:SvgMap)
+  RETURN { id: $sid,  svgs: COLLECT(DISTINCT(ssvg.filename)) } as data
```
mihai-sysbio, line 95: There is, and there will always be, a single SVG map per subsystem. Is it really needed to have `DISTINCT` here?

api/src/neo4j/queries/gene.js
```diff
@@ -46,5 +47,94 @@ RETURN apoc.map.mergeList(COLLECT(value.data)) as gene
...
+  UNION
+
+  MATCH (:Subsystem {id: $sid})-[:V1_3_0]-(r:Reaction)-[:V1_3_0]-(g:Gene)
+  USING JOIN ON r
+  RETURN DISTINCT { id: $sid, genes: COLLECT(DISTINCT(g.id)) } as data
+
+  UNION
+
+  MATCH (:Subsystem {id: $sid})-[:V1_3_0]-(r:Reaction)
+  RETURN DISTINCT { id: $sid, reaction_count: COUNT(DISTINCT(r.id)) } as data
```
mihai-sysbio, line 84:
```suggestion
  RETURN DISTINCT { id: $sid, reaction_count: COUNT(DISTINCT(r)) } as data
```

mihai-sysbio (changes requested):

mihai-sysbio:
Nice! Overall the queries seem to do the job. Since HPA queries a gene at a time, and there are thousands of them, it might make sense to consider further optimization. It's not of high importance though to perfect the `/hpa/gene/:id` endpoint.

api/src/neo4j/queries/gene.js
e0: Absolutely, this was sloppy by me.

api/src/neo4j/queries/gene.js
e0: Thanks! Addressed in: https://github.com/MetabolicAtlas/MetabolicAtlas/pull/546/commits/e3f6ac9e8e5d84decd379fdd3fbe4829cbe1b61a

e0:
> Nice! Overall the queries seem to do the job. Since HPA queries a gene at a time, and there are thousands of them, it might make sense to consider further optimization. It's not of high importance though to perfect the `/hpa/gene/:id` endpoint.

Would love to explore this more but this is one of the things that is highly unpredictable in terms of time vs reward. Currently all the requests take around 200-300ms for me locally. Do you think that's acceptable or do we want to target a certain lower time?

mihai-sysbio (approved):

nanjiangshu (commented):
Hi @e0, the changes look great to me. However, when I compared the running time of this API compared to the v1, it seems this API is slower. The number of genes retrieved is about twice as many for the v2 compared to v1 (6495 vs 3625) but takes almost 4x time.
Has the speed improvement been implemented in this pull request? 
```
$ time -p curl https://metabolicatlas.org/api/Human-GEM/genes > /data3/tmp/dir2/t1.txt
real 4.85
user 0.02
sys 0.02

$ time -p curl localhost:9080/api/v2/hpa/genes?model=Human-GEM > /data3/tmp/dir2/t2.txt  
real 17.14
user 0.01
sys 0.01      
```

e0:
Hi @nanjiangshu Thanks for the review! I think the current link for HPA genes is: https://metabolicatlas.org/api/hpa/genes and it returns 6604 genes.
As for the speed, I think almost all of the requests are slower compared to the current production site, which runs on Python PostgreSQL. Our new backend that uses neo4j provides a lot of new capabilities in terms of what we can retrieve but is not as fast. The speed improvements we mentioned previously was in terms of breakthroughs we made in neo4j cypher queries.

mihai-sysbio:
We can change the nginx caching rule for these requests so that they are cached after the first time, and to keep them in cache forever. This would also be a good time to re-check the size of the cache.

nanjiangshu:
> Hi @nanjiangshu Thanks for the review! I think the current link for HPA genes is: https://metabolicatlas.org/api/hpa/genes and it returns 6604 genes.
> As for the speed, I think almost all of the requests are slower compared to the current production site, which runs on Python PostgreSQL. Our new backend that uses neo4j provides a lot of new capabilities in terms of what we can retrieve but is not as fast. The speed improvements we mentioned previously was in terms of breakthroughs we made in neo4j cypher queries.

Hi Shan @e0, you are right, the url for the current sever should be https://metabolicatlas.org/api/hpa/genes and it returns 6604 genes. The speed difference is quite big, 0.22 second vs 17 seconds. Despite richer querying capabilities of neo4j, it seems the efficiency is not satisfactory for simple querying like this one. 

I guess a proper caching mechanism as mentioned by @mihai-sysbio might be a solution. Is caching enabled for the API querying in the local deployment? When I repeated the same querying 10 times there is no sudden speed improvement.

mihai-sysbio:
@nanjiangshu 17s would be indeed a bit much. However, my system shows values comparable with the previous implementation. The first time running the query it took a bit longer though, but I think that's because the db had not warmed up.             

```
$  time -p curl localhost/api/v2/hpa/genes > test
real 0.33
user 0.00
sys 0.00
```

e0:
Thanks for helping to confirm @mihai-sysbio. I'm not sure what the problem is with the setup for @nanjiangshu but it sounds like it's repeatedly taking much longer. If you can't find the culprit maybe we can have a look at it together.

mihai-sysbio:
@nanjiangshu did you get a chance to look at the latest version of this branch?

nanjiangshu (approved):
I got the same speed as yours with the latest changes. 
```
$ time -p curl localhost:9080/api/v2/hpa/genes > test
real 0.23
user 0.00
sys 0.01
```

PR #547, opened 2021-03-30
fix: improve search performance MA-912

e0:
This PR uses `JOIN` where appropriate to improve the response time for search requests. For example, before the fix it took > 20s to perform global search on `ATP` and with the fix in place it takes about 5s.

### DIFF
--- a/api/src/neo4j/queries/search.js
+++ b/api/src/neo4j/queries/search.js
@@ -91,6 +91,7 @@ CALL apoc.cypher.run("
   MATCH (:Gene:${model} {id: $gid})-[${version}]-(:Reaction)-[${version}]-(cm:CompartmentalizedMetabolite)
   WITH DISTINCT cm
   MATCH (cm)-[${version}]-(c:Compartment)-[${version}]-(cs:CompartmentState)
+  USING JOIN on c
   RETURN { id: $gid, compartment: COLLECT(DISTINCT({ id: c.id, name: cs.name })) } as data
 ", {gid:gid}) yield value
 RETURN apoc.map.mergeList(apoc.coll.flatten(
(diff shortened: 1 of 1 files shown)


mihai-sysbio (approved):
I'm really loving the performance gain! And it was such a small change.
Having had a look at the `DISTINCT` clauses, I have a hunch that there might be some that are redundant, in the cases where this is used both in an `WITH` and a subsequent `RETURN`. Would it make sense to have a look at these?

Now, I'm thinking if there are other places in the interface where the queries appear to be lagging, even if not as bad as the global search. I can think of two parts - the random tiles and the reaction table. If those really are slow, @e0 would you prefer to have another card for these?

e0:
> Having had a look at the `DISTINCT` clauses, I have a hunch that there might be some that are redundant, in the cases where this is used both in an `WITH` and a subsequent `RETURN`. Would it make sense to have a look at these?

I removed a couple of instances where they are obvious. There may be some other places where the `WITH` and `RETURN` do not involve the same component but could be removed too but I haven't touched those for now. I think in general I was thinking that it is safer to always have it at almost 0 performance cost. But having redundant `DISTINCT`s could certainly confusing to other developers.

___

> Now, I'm thinking if there are other places in the interface where the queries appear to be lagging, even if not as bad as the global search. I can think of two parts - the random tiles and the reaction table. If those really are slow, @e0 would you prefer to have another card for these?

I tried it a bit on reaction table yesterday and "only" got about 40% improvement, which is still worth adding I think. Random tiles is definitely another good candidate. I do think having another card is more appropriate since these do not trigger the timeout issues in this associated Jira task.

mihai-sysbio:
> having another card

A new card MA-931 points to this thread.

nanjiangshu (approved):

PR #549, opened 2021-04-05
feat: serve repository models using http MA-916

e0:


frontend/src/components/Repository.vue
```diff
@@ -128,7 +128,10 @@
                 <template v-for="file in selectedModel.files">
                   <a :key="file.path" class="button" :href="`${filesURL}${file.path}`">{{ file.format }}</a>&nbsp;
                 </template>
-                <br><br>
+                <div class="notification">
```
mihai-sysbio, line 131: There is a [Bulma spacing helper class](https://bulma.io/documentation/helpers/spacing-helpers/) that could be used instead of the custom CSS:

```suggestion
                <div class="notification mt-4">
```

frontend/src/components/Repository.vue
```diff
@@ -343,4 +346,10 @@ export default {
...
 </script>
 
-<style lang="scss"></style>
+<style lang="scss" scoped>
+  .modal {
+    .notification {
+      margin-top: 1rem;
+    }
+  }
+</style>
```
mihai-sysbio, line 355: With the Bulma class this becomes redundant.

```suggestion
```

api/src/endpoints/repository.js
```diff
@@ -41,4 +41,6 @@ routes.get('/models', async (req, res) => {
   }
 });
 
+routes.use('/models', express.static('repository'));
+
```
mihai-sysbio, line 45: How about adding
```
app.use('/models', express.static('repository'));
```
to [index.js](https://github.com/MetabolicAtlas/MetabolicAtlas/blob/develop/api/src/index.js)?

mihai-sysbio (approved):
Nice! I'm impressed by how little changes were needed to make this work. There is some feedback, but only cosmetics - feel free to resolve and merge.

api/src/endpoints/repository.js
e0: I see that argument and I had it there first actually. But I wanted to follow the old URL pattern as well as the route nesting architecture we have set up in the `api` so it made a bit more sense for me to have in the `repository.js` file.

e0:
Thanks for the feedback @mihai-sysbio. I applied the styling changes you suggested and left the routing unchanged for now (with the reasoning in comment above).

PR #550, opened 2021-04-06
feat: random interaction partners MA-741

e0:


mihai-sysbio:
You found a really good way to reuse the request @e0!

Some initial thoughts:
- [ ] on my machine, the random tiles for the GEM Browser are not loading
- [ ] to maintain consistency with the GEM Browser, it would make sense to move the demo video of the Interaction Partners to the Documentation
- [ ] with the change of the `random-components` request, the Swagger config/description of this endpoints should be updated as well
- [ ] as a user, I would like to go back to the main Interaction Partner page, so that I can try another random component. At the moment, this can only be done by going back in the browser. I'm thinking that the `Interaction Partners` text could be a link.

e0:
Thanks for the review and feedback!
___
> * [ ]  on my machine, the random tiles for the GEM Browser are not loading

Looks like it is only working for Human and Yeast at the moment. Looking more into it.

___
> * [ ]  to maintain consistency with the GEM Browser, it would make sense to move the demo video of the Interaction Partners to the Documentation

It's a good suggestion but it also makes the page looks relatively empty since there is not as many components shown. I think we could add a new Jira task to consider this though.

___
> * [ ]  with the change of the `random-components` request, the Swagger config/description of this endpoints should be updated as well

Good point, will try to fix this soon.

___
> * [ ]  as a user, I would like to go back to the main Interaction Partner page, so that I can try another random component. At the moment, this can only be done by going back in the browser. I'm thinking that the `Interaction Partners` text could be a link.

Makes sense and I think in general the Interaction Partners component needs some refactoring. It's a quite large file and I don't see a good reason for keeping the main page and component page in the same file. I think this is also something that could use a new Jira task.

mihai-sysbio:
Thinking more about moving the demo video to the documentation, I think I see your point @e0 - the page would appear too empty for a long time until we come up with a good way to fill up the space (not that empty space is a problem in moderate quantities).
Regarding the navigation on the website, I guess we can sweep this under the rug until we get professional UX advice.

The current side-by-side layout is not as good as it should be, though. 

I know we already talked about an alternative layout, and you mentioned trying the one below, which resulted in elements shifting down as random components are loaded.
image
Perhaps @inghylt or @nanjiangshu have some other ideas to improve the layout?

mihai-sysbio (changes requested):
From a technical perspective, all looks good, but I find the side-by-side layout needs improvement.

e0:
Good input! The current layout is what I personally thought was good enough. If there is any specific suggestions on what you don't like or what you would prefer, I would be happy to consider it.

nanjiangshu (commented):
Basically looks good. Maybe we can adjust the height of the cards on the left a bit so that the total height of cards region (on the right) are similar to the height of the demo video region (on the left)

e0:
> Basically looks good. Maybe we can adjust the height of the cards on the left a bit so that the total height of cards region (on the right) are similar to the height of the demo video region (on the left)

That's a good suggestion but it might not be needed if we are planning to move the video to the documentation page. To clarify I have no problem with moving the video to the documentation page as soon as possible. I only ask that we create a new Jira card for it so we try to keep the habit of not letting Jira card scopes grow.

nanjiangshu:
> > Basically looks good. Maybe we can adjust the height of the cards on the left a bit so that the total height of cards region (on the right) are similar to the height of the demo video region (on the left)
> 
> That's a good suggestion but it might not be needed if we are planning to move the video to the documentation page. To clarify I have no problem with moving the video to the documentation page as soon as possible. I only ask that we create a new Jira card for it so we try to keep the habit of not letting Jira card scopes grow.

Fair enough. Then it looks good to me now.

nanjiangshu (approved):

mihai-sysbio (approved):

mihai-sysbio:
For the planned layout changes see MA-936.

PR #551, opened 2021-04-13
fix: max gems comparison MA-928

e0:


mihai-sysbio (approved):
Looks good to me!

inghylt (approved):
Haven't tried it live but the changes looks good to me!

nanjiangshu (changes requested):
Works like a charm. Just have one comment, I think the first character of the text below the comparison table should be capitalized. 
I mean ```click a cell to see the comparison details
legend: reactions metabolites```
should be 
```Click a cell to see the comparison details
Legend: reactions metabolites```

mihai-sysbio:
Just to be clear, if `legend` is to be capitalized, so should `click`.

nanjiangshu (approved):
Hi @shan-chalmers, do you want to capitalize the leading words `click` and `legend` in the table legend? It's all up to you and fine to keep as it is. Either yes or no I can approve this pull request.

e0:
> Hi @shan-chalmers, do you want to capitalize the leading words `click` and `legend` in the table legend? It's all up to you and fine to keep as it is. Either yes or no I can approve this pull request.

I personally prefer it the way it is and I think @mihai-sysbio does too so I think we can leave it like that for now. Thanks again for looking at this!

PR #552, opened 2021-04-13
fix:  missing random reaction MA-929

e0:


mihai-sysbio (approved):
Interesting, so the missing tiles where for cases when reactions did not have reactants or products.

Looks good to me, including performance-wise on the frontend.

nanjiangshu (approved):
Sorry I don't quite understand the neo4j part. Seems the new code try to select only reactions with both reactants and products. Moreover, I never encounter any missing reaction card after refreshing the page for 50+ times.

e0:
> Sorry I don't quite understand the neo4j part. Seems the new code try to select only reactions with both reactants and products. Moreover, I never encounter any missing reaction card after refreshing the page for 50+ times.

Sorry, I should try to remember to explain the PRs in more details. I think the problem before (please let me know if I'm wrong @mihai-sysbio) was that we would try to get 2 random reactions, and for each reaction we are looking for at least 2 reactants and 2 products. The problem is that some reactions do not have at least 2 reactants or/and 2 products and then the reaction does not get included in the response. For example if it's an exchange reaction.

The change adds the `OPTIONAL` keyword in front of the `MATCH` for reactants and products so that the reaction still gets returned regardless if there the reactants and products are there or not.

mihai-sysbio:
In a first step, some reactions would be chosen. After this, their details were queries, with a `match` on reactants _and_ products. If, by chance, a reaction like [HMR_10024](https://metabolicatlas.org/explore/Human-GEM/gem-browser/reaction/HMR_10024) would be chosen, its details could not be fetched so no result would be returned.

PR #553, opened 2021-04-13
feat: db performance enhancement MA-931

e0:
This PR covers performance enhancements for the following queries:

- random components: using the `cytosol` compartment as an example, response time improvement: `~40s` -> `~1s`
- related reactions: using the `transport_reactions` subsystem as an example, response time improvement: from `~6300ms` -> `~1200ms`

### DIFF
--- a/api/src/neo4j/queries/randomComponents.js
+++ b/api/src/neo4j/queries/randomComponents.js
@@ -57,7 +57,8 @@ CALL apoc.cypher.run("
   MATCH (:Compartment${m} {id: $cid})-[${v}]-(:CompartmentalizedMetabolite)-[${v}]-(:Reaction)-[${v}]-(g:Gene)
   RETURN { geneCount: count(distinct g) } as data
   UNION
-  MATCH (:Compartment${m} {id: $cid})-[${v}]-(:CompartmentalizedMetabolite)-[${v}]-(:Reaction)-[${v}]-(:Subsystem)-[${v}]-(sss:SubsystemState)
+  MATCH (:Compartment${m} {id: $cid})-[${v}]-(:CompartmentalizedMetabolite)-[${v}]-(:Reaction)-[${v}]-(s:Subsystem)-[${v}]-(sss:SubsystemState)
+  USING JOIN on s
   RETURN { majorSubsystems: COLLECT(DISTINCT(sss.name))[..15] } as data
 ", {cid:cid}) yield value
 RETURN { compartment: apoc.map.mergeList(COLLECT(value.data)) } as xs
--- a/api/src/neo4j/queries/relatedReactions.js
+++ b/api/src/neo4j/queries/relatedReactions.js
@@ -67,6 +67,7 @@ CALL apoc.cypher.run("
   MATCH (:Reaction${m} {id: $rid})-[${v}]-(cm:CompartmentalizedMetabolite)
   WITH DISTINCT cm
   MATCH (cm)-[${v}]-(c:Compartment)-[${v}]-(cs:CompartmentState)
+  USING JOIN on c
   RETURN { id: $rid, compartments: COLLECT(DISTINCT(cs {id: c.id, .*})) } as data
  
   UNION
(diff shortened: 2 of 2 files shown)


api/src/neo4j/queries/relatedReactions.js
```diff
@@ -87,8 +88,9 @@ CALL apoc.cypher.run("
  
   MATCH (:Reaction${m} {id: $rid})-[cmE${v}]-(cm:CompartmentalizedMetabolite)
   WITH DISTINCT cm, cmE
-  MATCH (cm)-[${v}]-(:Metabolite)-[${v}]-(ms:MetaboliteState)
   MATCH (cm)-[${v}]-(c:Compartment)-[${v}]-(cs:CompartmentState)
+  USING JOIN on c
+  OPTIONAL MATCH (cm)-[${v}]-(:Metabolite)-[${v}]-(ms:MetaboliteState)
```
mihai-sysbio, line 93: I wasn't expected to see the `OPTIONAL` here, is there special reason for it?

mihai-sysbio (approved):
awesome!

api/src/neo4j/queries/relatedReactions.js
e0: Without it the request takes about twice as long.

mihai-sysbio: That's really weird, it must mean that it chooses to traverse the network differently.
How would the results be affected if there is no match for that `optional` clause?

e0: I would guess that it returns the last part of the query as `{id: $rid, metabolites: []}` since we are performing a list aggregate operation, so it should be safe. In theory there shouldn't be any `CompartmentalizedMetabolite` that is not connected to any `Metabolite`.

I actually ran a `PROFILE` to compare the query with and without `OPTIONAL` now and the output plans are identical. I even saved them as PNGs and used a tool to compare the images to make sure.

We have previously used similar patterns with `OPTIONAL` in certain queries and they made a positive impact, which is why I tried it here as well.

mihai-sysbio: Thanks for sharing these details, `OPTIONAL` might come in handy. Personally, I'm still skeptical about (over)using it, but in this case it seems to do a good job.

api/src/neo4j/queries/randomComponents.js
```diff
@@ -57,7 +57,8 @@ CALL apoc.cypher.run("
   MATCH (:Compartment${m} {id: $cid})-[${v}]-(:CompartmentalizedMetabolite)-[${v}]-(:Reaction)-[${v}]-(g:Gene)
   RETURN { geneCount: count(distinct g) } as data
   UNION
-  MATCH (:Compartment${m} {id: $cid})-[${v}]-(:CompartmentalizedMetabolite)-[${v}]-(:Reaction)-[${v}]-(:Subsystem)-[${v}]-(sss:SubsystemState)
+  MATCH (:Compartment${m} {id: $cid})-[${v}]-(:CompartmentalizedMetabolite)-[${v}]-(:Reaction)-[${v}]-(s:Subsystem)-[${v}]-(sss:SubsystemState)
+  USING JOIN on s
```
nanjiangshu, line 61: It's a bit strange that I don't get any performance improvement for the compartment `cytosol`
Without `JOIN on`
```
time -p curl http://localhost:9080/api/v2/compartments/cytosol?model=HumanGem 
real 1.41
``` 
and with `JOIN on`
```
$ time -p curl http://localhost:9080/api/v2/compartments/cytosol?model=HumanGem > t1.txt
real 1.50
```

e0: This only applies for the `api/v2/random-components` endpoint.

api/src/neo4j/queries/relatedReactions.js
```diff
@@ -67,6 +67,7 @@ CALL apoc.cypher.run("
   MATCH (:Reaction${m} {id: $rid})-[${v}]-(cm:CompartmentalizedMetabolite)
   WITH DISTINCT cm
   MATCH (cm)-[${v}]-(c:Compartment)-[${v}]-(cs:CompartmentState)
+  USING JOIN on c
```
nanjiangshu, line 70: For the related-reactions, I do see a performance improvement. 
With `JOIN on`
```
time -p curl http://localhost:9080/api/v2/subsystems/transport_reactions/related-reactions?model=HumanGem > t1.txt
real: 45.01
```
without `JOIN on`

timeout after 60 seconds

Is that because neo4j is much slower on my Linux machine?

nanjiangshu (commented):
even for the random-component endpoint, I see only marginal improvement.
I repeated the request 20 times and calculate the average time. 
```
$ for i in $(seq 20) ; do time -p curl http://localhost:9080/api/v2/random-components  > /dev/null; done 2> t1.txt 
```
This branch : 0.6945 second
develop branch:  0.835 second

Maybe I should try it also on my laptop as well.

api/src/neo4j/queries/relatedReactions.js
e0: For random components, could you try running it more times? The bottleneck is on compartments and it only gets noticeably slower if the randomized compartment is something like `Cytosol`.

e0: For related reactions, it was using a limit which is not hard-coded in the backend, so the request should look like `/api/v2/subsystems/transport_reactions/related-reactions?model=HumanGem&limit=200`.

e0: @nanjiangshu sorry, actually I notice that on the server the random components endpoint is not slow even if we are getting `Cytosol`, it is probably just my machine that was acting up the other day which was causing the long load times. I think the change still has a positive effect though, albeit much less significant.

nanjiangshu: An improvement is an improvement :)

nanjiangshu (approved):

PR #554, opened 2021-04-14
fix: deploy cmd for prod env

mihai-sysbio:


e0 (approved):
We should also consider updating the other commands and documentation on Docker version requirement etc but of course we can merge in this urgent change first.

proj.sh
```diff
@@ -41,7 +41,7 @@ function logs {
 
 function deploy-stack {
   generate-data
-  docker-compose -f docker-compose.yml -f docker-compose-prod.yml --context $1 up -d --build
+  docker --context $1 compose -f docker-compose.yml -f docker-compose-local.yml logs -f nginx
```
nanjiangshu, line 44: The change of `docker-compose-prod.yml` to `docker-compose-local.yml`, is that just a temporary fix? How is the SSL certificate set up in this case?

e0: Oh I didn't realize that haha. It wasn't intentional and we should probably not merge this in yet..

mihai-sysbio: also we  might not need the `logs -f nginx` at the end either

e0, line 44:
```suggestion
  docker --context $1 compose -f docker-compose.yml -f docker-compose-prod.yml up -d --build --force-recreate
```
Fixes the script and also adds the `--force-recreate` flag. This ensures that we do not use cached images when deploying.

mihai-sysbio: This is great.

nanjiangshu (approved):

PR #555, opened 2021-04-15
fix: make sure compartment is requested after route change MA-926

e0:
Jira task: MA-926

The compartment component would not load in certain situations, this fixes that.

Steps to reproduce former buggy behavior (written by @nanjiangshu):
1. On the home page, click "Map viewer"
2. Then click "GEM Browser"
3. Then click one of the 2D or 3D Map Viewer
4. Then click "GEM Browser" again, and the server will be in the spinning mode here.

### DIFF
--- a/frontend/src/components/explorer/gemBrowser/Compartment.vue
+++ b/frontend/src/components/explorer/gemBrowser/Compartment.vue
@@ -130,16 +130,14 @@ export default {
     async setup() {
       this.showLoaderMessage = 'Loading compartment data';
       this.cName = this.$route.params.id;
-      if (this.cName !== this.compartment.id) {
-        try {
-          const payload = { model: this.model, id: this.cName };
-          await this.$store.dispatch('compartments/getCompartmentSummary', payload);
-          this.componentNotFound = false;
-          this.showLoaderMessage = '';
-        } catch {
-          this.componentNotFound = true;
-          document.getElementById('search').focus();
-        }
+      try {
+        const payload = { model: this.model, id: this.cName };
+        await this.$store.dispatch('compartments/getCompartmentSummary', payload);
+        this.componentNotFound = false;
+        this.showLoaderMessage = '';
+      } catch {
+        this.componentNotFound = true;
+        document.getElementById('search').focus();
       }
     },
   },


frontend/src/components/explorer/gemBrowser/Compartment.vue
```diff
@@ -130,16 +130,14 @@ export default {
...
-          document.getElementById('search').focus();
-        }
+      try {
+        const payload = { model: this.model, id: this.cName };
+        await this.$store.dispatch('compartments/getCompartmentSummary', payload);
+        this.componentNotFound = false;
+        this.showLoaderMessage = '';
+      } catch {
+        this.componentNotFound = true;
+        document.getElementById('search').focus();
```
mihai-sysbio, line 140: In principle, I think it makes sense to dispatch a request only if needed, in which the loader message should also be conditional:

```suggestion
      if (this.cName !== this.compartment.id) {
        this.showLoaderMessage = 'Loading compartment data';
        try {
          const payload = { model: this.model, id: this.cName };
          await this.$store.dispatch('compartments/getCompartmentSummary', payload);
          this.componentNotFound = false;
          this.showLoaderMessage = '';
        } catch {
          this.componentNotFound = true;
        }
      }
```

Also, this line is not needed anymore now that the GEM Browser doesn't have its dedicated search bar:
```
          document.getElementById('search').focus();
```

frontend/src/components/explorer/gemBrowser/Compartment.vue
```diff
@@ -130,16 +130,14 @@ export default {
     async setup() {
       this.showLoaderMessage = 'Loading compartment data';
```
mihai-sysbio, line 131: To me, it doesn't look like this is the right place for this.

```suggestion
```

mihai-sysbio (commented):
However we decide to set this up, it should be the same for other components too.

frontend/src/components/explorer/gemBrowser/Compartment.vue
e0: Do you remember why that check was added in the first place? I'm wondering if it was necessary due to the `keep-alive` stuff that we used to have. Anyhow, I didn't notice any problems for the other components and none of them had this `if` clause so I simply matched this one to the others.

frontend/src/components/explorer/gemBrowser/Compartment.vue
e0: Where would be right for it?

e0:
> However we decide to set this up, it should be the same for other components too.

Thanks for the review. I think there is some points worth considering for a refactor work. This PR specifically focuses on solving the bug as described in MA-926. Let's try to create a new task for other considerations unless they impact solving this bug.

frontend/src/components/explorer/gemBrowser/Compartment.vue
nanjiangshu: Shouldn't the value of `this.showLoaderMessage` to initialized if `this.cName === this.compartment.id`?

nanjiangshu (approved):
Can be merged since the bug of this issue is solved.

frontend/src/components/explorer/gemBrowser/Compartment.vue
mihai-sysbio: The point of the `if` was to avoid making the same request again if the compartment id that is in the component is the same as the id in the url. However, I think Vuex requests are cached so making this request is only a few ms of JS. As @e0 mentioned, this old `if` is likely a relic from `keep-alive`.

@e0 what about the line below, marked for removal in my suggestion?
```
document.getElementById('search').focus();
```

frontend/src/components/explorer/gemBrowser/Compartment.vue
mihai-sysbio: It was moved to the previous suggestion, but now irrelevant.

frontend/src/components/explorer/gemBrowser/Compartment.vue
e0: @mihai-sysbio Although it's not related to this bug fix I think that absolutely should be removed. The only reason why I don't want to merge it along with this PR is we have the same line in all the other components.

mihai-sysbio: The suggested refactor is carried out in #559, so the current PR is self-contained and can be approved.

mihai-sysbio (approved):

PR #556, opened 2021-04-15
fix: create folder api/repository if it does not exist

nanjiangshu:
Create the folder `api/repository` if it does not exist (which is the case after a fresh clone), otherwise, the command ```/bin/cp -r $DATA_FILES_PATH/ftp-models/* api/repository/``` will end up in an error.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -86,9 +86,9 @@ pids
 *.pid.lock
 
 # GEM repository in FTP
-ftp-models/
+ftp/repository/
 
-# GEM repository in FTP
+# GEM repository in API
 api/repository/
 
 # SVG folder in API
--- a/ftp/Dockerfile
+++ b/ftp/Dockerfile
@@ -1,6 +1,6 @@
 FROM stilliard/pure-ftpd:hardened
 
-COPY ftp-models /var/ftp/models
+COPY repository /var/ftp/models
 RUN useradd -d /var/ftp -s /sbin/nologin ftp
 
 COPY ./prep-cert.sh /project/prep-cert.sh
--- a/proj.sh
+++ b/proj.sh
@@ -11,7 +11,7 @@ function generate-data {
   /bin/cp -r $DATA_FILES_PATH/gemsRepository.json api/src/data/
   /bin/cp -r $DATA_FILES_PATH/svg api/
   /bin/cp -r $DATA_FILES_PATH/ftp-models ftp/
-  /bin/cp -r $DATA_FILES_PATH/ftp-models/* api/repository/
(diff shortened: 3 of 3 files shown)


e0 (approved):
Nice catch!

mihai-sysbio (changes requested):
This case seems conceptually identical with that on line `12` above. Wouldn't it be simpler to rename the `ftp-models` folder in `data-files` to `repository`?
Also, `ftp-models` is a name we don't use so much - imho, it add confusion, so I support renaming it regardless of the current issue.

inghylt (approved):

nanjiangshu:
> This case seems conceptually identical with that on line `12` above. Wouldn't it be simpler to rename the `ftp-models` folder in `data-files` to `repository`?
> Also, `ftp-models` is a name we don't use so much - imho, it adds confusion, so I support renaming it regardless of the current issue.

@mihai-sysbio , Good point! I think it is less confusion if name `ftp-models` to `repository`, then we have the same name for the folder containing downloadable models. Actually, we may copy these models only at one place and the other one just create a symlink. 

However, if we rename `ftp-models` to `repository`, it means the folder `ftp/ftp-models/` will not be existing anymore. The path to this folder needs to be changed at other places as well. Except for the file `ftp/Dockerfile`, do you know other places to be changed?

mihai-sysbio:
`proj.sh` comes to mind, but since it's a unique name a find+replace should do.

e0 (changes requested):
Please add `ftp/repository/` to the `.gitignore` file.

e0 (approved):

nanjiangshu:
> Please add `ftp/repository/` to the `.gitignore` file.

Done! 

P.S. I'm sorry that I have pushed this commit to your branch `feat/interaction-partners-layout_MA-936` but not reverted.

e0:
> > Please add `ftp/repository/` to the `.gitignore` file.
> 
> Done!
> 
> P.S. I'm sorry that I have pushed this commit to your branch `feat/interaction-partners-layout_MA-936` but not reverted.

No worries but thanks for mentioning it. I can do a force push to remove it.

nanjiangshu:
> > > Please add `ftp/repository/` to the `.gitignore` file.
> > 
> > 
> > Done!
> > P.S. I'm sorry that I have pushed this commit to your branch `feat/interaction-partners-layout_MA-936` but not reverted.
> 
> No worries but thanks for mentioning it. I can do a force push to remove it.

Typo! I mean I have reverted the commit :)

mihai-sysbio (approved):
Neat!

PR #557, opened 2021-04-15
fix: readonly api MA-930

e0:
Jira issue MA-930

To avoid dangerous cypher injection into neo4j database, this PR ensures that requests made from API to neo4j are in readonly mode.


mihai-sysbio (approved):
lgtm

inghylt (approved):

PR #558, opened 2021-04-20
fix: make sure the linked integrated model is selected MA-940

e0:
Jira issue MA-940

### Steps to reproduce buggy behavior:

1. Visit the `/about` page.
2. Click `Human-GEM` link in the second paragraph of the page.
3. The user is now taken to the explore page with the correct URL `/explore/Human-GEM` but the wrong model is selected. (At the time of writing, the selected one is `Fruitfly-GEM`, which is the first one listed.)

### Acceptance test scenario:

1. Visit the `/about` page.
2. Click `Human-GEM` link in the second paragraph of the page.
3. The user should now be taken to the explore page with the correct URL `/explore/Human-GEM` and the `Human-GEM` model selected.
4. Repeat steps 1-3 for `Yeast-GEM`.

### DIFF
--- a/frontend/src/components/Explorer.vue
+++ b/frontend/src/components/Explorer.vue
@@ -110,10 +110,18 @@ export default {
       model: state => state.models.model,
     }),
   },
+  mounted() {
+    const modelShortName = this.$route.params.model;
+    if (modelShortName && this.model && modelShortName !== this.model.short_name) {
+      this.selectModel(modelShortName);
+    }
+  },
   methods: {
     async selectModel(modelShortName) {
-      if (modelShortName !== this.model.short_name) {
+      if (modelShortName !== this.$route.params.model) {
         this.$router.replace({ params: { model: modelShortName } });
+      }
+      if (modelShortName !== this.model.short_name) {
         this.$store.dispatch('models/selectModel', modelShortName);
       }
     },
--- a/frontend/src/components/Explorer.vue
+++ b/frontend/src/components/Explorer.vue
@@ -117,7 +117,7 @@ export default {
     }
   },
   methods: {
-    async selectModel(modelShortName) {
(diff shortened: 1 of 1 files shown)


nanjiangshu (approved):
Well done!

inghylt (approved):
Nice! Neat PR description as well

frontend/src/components/Explorer.vue
```diff
@@ -110,10 +110,18 @@ export default {
...
+    const modelShortName = this.$route.params.model;
+    if (modelShortName && this.model && modelShortName !== this.model.short_name) {
+      this.selectModel(modelShortName);
+    }
+  },
   methods: {
-    async selectModel(modelShortName) {
-      if (modelShortName !== this.model.short_name) {
+    selectModel(modelShortName) {
+      if (modelShortName !== this.$route.params.model) {
```
mihai-sysbio, line 121: This check is a bit weird here in the context looking how on line 114
```
const modelShortName = this.$route.params.model;
```

I'm really not sure how to improve this by reducing all the checks, especially that this setup seems to work.

e0: I'm not sure either since these two `if` conditions are needed depending on the conditions and where `selectModel` is being called from. We could perhaps write two different methods instead, such as `selectModel` and `restoredModelFromURL` but since this method is not overcomplicated at the moment I think we can hold off on that.

mihai-sysbio: I think some refactoring would be due, since the last check on line 115 is actually duplicated in 124.

e0: Ah you are right! I was looking at line 114 first but we could indeed remove the second condition in line 115. Since this branch is already merged I will include this in a different PR.

PR #559, opened 2021-04-20
refactor: there is no need to focus on search in GEM Browser

mihai-sysbio:
Through this PR the pages for the model components are refactored, removing the browser focus on the search, since that functionality is requires a click now. This focus was relevant before, when the search bar was embedded in the page above the model component, always visible.


e0 (changes requested):
Looks fine but there is a merge conflict that needs to be fixed.

e0 (approved):

inghylt (approved):

PR #560, opened 2021-04-21
feat: add faint border, caption and link to chebi image and let table fill available space if no image available MA-921

inghylt:


frontend/src/components/explorer/gemBrowser/Metabolite.vue
```diff
@@ -184,4 +186,10 @@ export default {
 };
 </script>
 
-<style lang="scss"></style>
+<style lang="scss">
+#chebi-img {
+   max-width: 400px;
+   width:100%;
```
mihai-sysbio, line 192: I couldn't find a scenario where these 2 rules would be required.

```suggestion
```

frontend/src/components/explorer/gemBrowser/Metabolite.vue
```diff
@@ -184,4 +186,10 @@ export default {
...
+   border: 1px solid #dbdbdb;
```
mihai-sysbio, line 193: For easier maintenance, we should try to rely on Bulma colors or [vars.scss](https://github.com/MetabolicAtlas/MetabolicAtlas/blob/develop/frontend/src/style/vars.scss):

```suggestion
   border: 1px solid $grey-lighter;
```

frontend/src/components/explorer/gemBrowser/Metabolite.vue
```diff
@@ -70,7 +70,9 @@
             <div v-if="metabolite && metabolite.externalDbs && metabolite.externalDbs.ChEBI"
                  class="column is-3-widescreen is-2-desktop is-full-tablet has-text-centered px-2">
               <a :href="metabolite.externalDbs.ChEBI[0].url" target="_blank">
-                <img :src="`https://www.ebi.ac.uk/chebi/displayImage.do?defaultImage=true&imageIndex=0&chebiId=${metabolite.externalDbs.ChEBI[0].id.slice(6)}&dimensions=400`" />
+                <img id="chebi-img" :src="`https://www.ebi.ac.uk/chebi/displayImage.do?defaultImage=true&imageIndex=0&chebiId=${metabolite.externalDbs.ChEBI[0].id.slice(6)}&dimensions=400`" class="hoverable"/>
+                <a :href="metabolite.externalDbs.ChEBI[0].url" target="_blank"  style="display: block;">
+                  ChEBI {{metabolite.name}} </a>
```
mihai-sysbio, line 75: For people reading ltr the focus appears to be on ChEBI. I tried `ChEBI: ` and it wasn't better. How about:
```suggestion
                  {{metabolite.name}} via ChEBI</a>
```

mihai-sysbio (changes requested):

mihai-sysbio:
This is a definite improvement! I can mentioned that I tried using the `box` Bulma class, but the shadow seemed to be competing for the user's attention with the right sidebar (Interaction Partners and Map Viewer).

e0 (changes requested):
Looks good @inghylt! I like the suggestions mentioned by @mihai-sysbio as well. There is also some code formatting linter warnings that we should fix. You can view them with `logs frontend`, which is provided by `proj.sh`.

frontend/src/components/explorer/gemBrowser/Metabolite.vue
inghylt: Nice, me neither tbh. Think it was added to avoid the table taking up too much space but as you say it seems to work fine without it

frontend/src/components/explorer/gemBrowser/Metabolite.vue
inghylt: Tried it out and I think it looks good!

mihai-sysbio (approved):

e0 (approved):

nanjiangshu:
Looks really good @inghylt! 
I just have one comment for the display when ChEBI does not have a valid image. e.g. for 
http://localhost/explore/Human-GEM/gem-browser/metabolite/m02576c (a screenshot can be found here https://www.dropbox.com/s/xlgyv5t7xr4oxtw/Screenshot%202021-04-22%20at%2015.19.40.png?dl=0).
Since the ChEBI image does not exists, it shows an icon with non-existing image on the webpage, looks at the website is broken. Maybe there is a better solution for handling such cases?

nanjiangshu (changes requested):
Looks really good @inghylt!
I just have one comment for the display when ChEBI does not have a valid image. e.g. for [m02576c](http://localhost/explore/Human-GEM/gem-browser/metabolite/m02576c) (a screenshot can be found [here](https://www.dropbox.com/s/xlgyv5t7xr4oxtw/Screenshot%202021-04-22%20at%2015.19.40.png?dl=0)).
Since the ChEBI image does not exists, it shows an icon with non-existing image on the webpage, looks at the website is broken. Maybe there is a better solution for handling such cases?

e0:
> Looks really good @inghylt!
> I just have one comment for the display when ChEBI does not have a valid image. e.g. for
> http://localhost/explore/Human-GEM/gem-browser/metabolite/m02576c (a screenshot can be found here https://www.dropbox.com/s/xlgyv5t7xr4oxtw/Screenshot%202021-04-22%20at%2015.19.40.png?dl=0).
> Since the ChEBI image does not exists, it shows an icon with non-existing image on the webpage, looks at the website is broken. Maybe there is a better solution for handling such cases?

That's a good catch @nanjiangshu. Can you please add a new Jira issue for this? I don't think it should block this PR.

nanjiangshu:
You are right @shan-chalmers , this should be done in a new issue.

nanjiangshu (approved):

PR #562, opened 2021-04-21
fix: show reaction details in sidebar

mihai-sysbio:
When loading a 2D map, one is expect to click on all nodes on the map: metabolites, genes and reactions, with the details of each component being shown in the sidebar.
The reaction information was not displayed, just showing as loading forever. Apparently the wrong function was called, throwing some errors in the JS console.
The fix was tested for both 2D and 3D, on several models.

### DIFF
--- a/frontend/src/components/explorer/MapViewer.vue
+++ b/frontend/src/components/explorer/MapViewer.vue
@@ -238,6 +238,8 @@ export default {
 #mapViewerContainer {
 
   #mapSidebar {
+    word-wrap: break-word;
+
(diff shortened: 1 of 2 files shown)


mihai-sysbio:
Great idea @e0, it's now applied to the entire left sidebar through 84caf72.

inghylt (approved):

mihai-sysbio:
Sorry for the force-push, I wanted to fix a typo in the commit message, going to merge this in.

PR #563, opened 2021-04-22
fix: 2D search bar MA-927

nanjiangshu:
Fixing the bug that the search bar on the 2D map can not type in text.

### DIFF
--- a/frontend/src/components/explorer/mapViewer/DataOverlay.vue
+++ b/frontend/src/components/explorer/mapViewer/DataOverlay.vue
@@ -213,9 +213,6 @@ export default {
       this.showFileLoader = true;
     });
   },
-  updated() {
-    setTimeout(() => document.getElementById('searchInput').focus());
-  },
   methods: {
     reloadGeneExpressionData() {
       if (this.mapLoaded && this.HPATissues.length > 0) {
--- a/frontend/src/components/explorer/mapViewer/DataOverlay.vue
+++ b/frontend/src/components/explorer/mapViewer/DataOverlay.vue
@@ -213,6 +213,9 @@ export default {
       this.showFileLoader = true;
     });
   },
+  updated() {
+    setTimeout(() => document.getElementById('searchInput').focus());
+  },
   methods: {
     reloadGeneExpressionData() {
       if (this.mapLoaded && this.HPATissues.length > 0) {
--- a/frontend/src/components/explorer/mapViewer/MapSearch.vue
+++ b/frontend/src/components/explorer/mapViewer/MapSearch.vue
@@ -74,6 +74,11 @@ export default {
       }
       this.currentSearchMatch = 0;
     },
+    loading(now, before) {
+      if (before === true && now === false) {
+        setTimeout(() => document.getElementById('searchInput').focus());
+      }
+    },
   },
   created() {
     this.search = debounce(this.search, 300);
--- a/frontend/src/components/explorer/mapViewer/MapSearch.vue
+++ b/frontend/src/components/explorer/mapViewer/MapSearch.vue
@@ -75,9 +75,7 @@ export default {
       this.currentSearchMatch = 0;
     },
     loading(now, before) {
-      if (before === true && now === false) {
-        setTimeout(() => document.getElementById('searchInput').focus());
-      }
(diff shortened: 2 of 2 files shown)


frontend/src/components/explorer/mapViewer/DataOverlay.vue
```diff
@@ -213,6 +213,9 @@ export default {
       this.showFileLoader = true;
     });
   },
+  updated() {
+    setTimeout(() => document.getElementById('searchInput').focus());
+  },
```
e0, line 218: Good catch on this interaction too. The `RNAexpression` component, via the `DataOverlay` component, currently uses `EventBus` to broadcast changes with RNA level application, using the line:

```javascript
EventBus.$emit(this.showing2D ? 'apply2DHPARNAlevels' : 'apply3DHPARNAlevels', {});
```

You can see how this is handled in `Svgmap`.

```javascript
  created() {
    EventBus.$off('apply2DHPARNAlevels');
    EventBus.$on('apply2DHPARNAlevels', (levels) => {
      this.applyHPARNAlevelsOnMap(levels);
    });
```

Using a similar pattern, I think it would be better to handle the `EventBus` message inside the `MapSearch` component so that it can focus on the `searchInput` by itself.

Two things I want to mention about this suggestion:

1. It seems like quite a bit more work to achieve the same goal but this keeps components more modular. The `DataOverlay` should not directly manipulate other components.
2. The `EventBus` is something we have been gradually trying to remove and refactor by using Vuex because we believe that in general Vuex provides a better and more explicit (less guesswork) design pattern. Having said that, there is still currently a lot of `EventBus` usage especially for dealing with RNA levels. So I think for now it is still recommendable to use `EventBus` for this case, until we plan a more thorough refactor work.

e0 (changes requested):
@nanjiangshu Good job and this seem to solve the bug. I would appreciate if we can do a bit of code refactoring before merging as posted in the inline comments. @mihai-sysbio one of the suggestion involve additional usage of `EventBus`, please take a look to see if it's something you agree with.

frontend/src/components/explorer/mapViewer/MapSearch.vue
```diff
@@ -137,6 +140,11 @@ export default {
       }
       this.$emit('centerViewOn', this.matches[this.currentSearchMatch]);
     },
+    focusOnInputSearch(now, before) {
+      if (before === true && now === false) {
+        setTimeout(() => document.getElementById('searchInput').focus());
```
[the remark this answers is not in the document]
nanjiangshu: Hi @shan-chalmers , thanks for the positive comments. It is actually your input for this point. I have moved the variables `before` and `now` away from the function `focusOnInputSearch()` and added a comment for the code.

frontend/src/components/explorer/mapViewer/DataOverlay.vue
nanjiangshu: Hi @shan-chalmers, thanks for the explanation and I understand slightly better with the relationships of these components and hooks after the debugging process. For this part of code, I have tried to replace the code `setTimeout(() => document.getElementById('searchInput').focus());` by reusing the function `focusOnInputSearch()` from `MapSearch`. However, it just does not work. Even I define the same function inside `DataOverlay.vue` but still not working. Feels a bit weird.

e0: Sorry, the `EventBus` stuff definitely is outside of our usual work with Vue, I should have been a bit more detailed in the explanation.

I think it would make sense to add the following to the `MapSearch` component:
```javascript
created() {
    EventBus.$off('apply2DHPARNAlevels');
    EventBus.$on('apply2DHPARNAlevels', () => {
      this. focusOnInputSearch();
    });
}
```

Is that what you tried? I haven't tried this myself so not 100% sure that it works but I think it should do the job.

e0 (approved):

inghylt (approved):
Nice! I would enjoy being able to double click and mark all the text inside the search bar, but I guess that should be a separate issue?

mihai-sysbio:
It's nice that you avoided the use of `jQuery`.

PR #564, opened 2021-04-26
fix: source url for matomo

mihai-sysbio:
The url needs to contain the full path up until the `/piwik.js` suffix.

### DIFF
--- a/frontend/src/main.js
+++ b/frontend/src/main.js
@@ -19,7 +19,7 @@ Vue.use(vueDebounce);
 
 if (navigator.doNotTrack !== '1') {
   Vue.use(VueMatomo, {
-    host: 'https://csbi.chalmers.se/',
+    host: 'https://csbi.chalmers.se/matomo',
     siteId: process.env.VUE_APP_MATOMOID,
     router,
   });


e0 (approved):
@mihai-sysbio I haven't really worked with this, don't think I have access to Matomo either but I guess you are sure of this so here is one of the mandatory approvals. 😅

mihai-sysbio:
@e0 thanks for pointing this out. After disabling the `Do Not Track` header in your browser, by inspecting the requests on metabolicatlas.org, you can see that there is a failed request for `https://csbi.chalmers.se//piwik.js`. To double check this, you can try to request this url in a new tab - and it fails indeed. However `https://csbi.chalmers.se/matomo/piwik.js` works, which is why I just added the subpath `matomo` in url.

nanjiangshu (approved):
I don't know how to test this neither but feels https://csbi.chalmers.se/matomo/ should be the correct URL. Not sure how it worked with https://csbi.chalmers.se/piwik.js previously.

mihai-sysbio:
> I don't know how to test this neither but feels https://csbi.chalmers.se/matomo/ should be the correct URL. Not sure how it worked with https://csbi.chalmers.se/piwik.js previously.

@nanjiangshu I think it was just my mistake not using the correct/full url, since `https://csbi.chalmers.se/piwik.js` has no chance of working.

PR #565, opened 2021-04-26
feat: gene and metabolite names MA-910

e0:
Jira issue MA-910

This PR includes the following changes:

- For the reactions page, for example `/explore/Human-GEM/gem-browser/reaction/HMR_0001`, display gene names instead of gene ids for "Gene rule".
- For the 3d map viewer, display names instead of ids on the nodes, related work: https://github.com/MetabolicAtlas/3d-network-viewer/commit/fc2a4d5b683aede007ac5bf914d3bfbec7e23798.
- For the map viewer's sidebar, when a component (gene or metabolite) node has been selected on the map, swap positions of name and id for genes and metabolites so that the names (more familiar for users) are more prominent.

This issue is originally intended to only focus on genes. For the map viewer however, both in the 3D viewer and the sidebar, the changes from id to name are applied for genes AND metabolites. This made the most sense to me for both users and from an implementation perspective. Please let me know if this doesn’t make sense and we want to have these changes only for genes and not metabolites.

### DIFF
--- a/api/src/neo4j/queries/reaction.js
+++ b/api/src/neo4j/queries/reaction.js
@@ -2,6 +2,15 @@ import querySingleResult from 'neo4j/queryHandlers/single';
 import reformatExternalDbs from 'neo4j/shared/formatter';
 import parseParams from 'neo4j/shared/helper';
 
+const formatGeneRule = ({ geneRule, genes }) => geneRule.split(/\s+/).map(w => {
+  if (w.match(/and|or/)) {
+    return w;
+  }
+
+  const gene = genes.find(g => g.id === w);
+  return gene.name;
+}).join(' ');
+
 const getReaction = async ({ id, model, version }) => {
   const [m, v] = parseParams(model, version);
 
@@ -59,6 +68,10 @@ RETURN apoc.map.mergeList(COLLECT(value.data)) as reaction
 `;
 
   const reaction = await querySingleResult(statement);
+  if (reaction.geneRule) {
+    reaction.geneRule = formatGeneRule(reaction);
+  }
+
   return { ...reaction, externalDbs: reformatExternalDbs(reaction.externalDbs) };
 };
 
--- a/api/src/neo4j/queries/reaction.js
+++ b/api/src/neo4j/queries/reaction.js
@@ -2,15 +2,6 @@ import querySingleResult from 'neo4j/queryHandlers/single';
 import reformatExternalDbs from 'neo4j/shared/formatter';
 import parseParams from 'neo4j/shared/helper';
 
-const formatGeneRule = ({ geneRule, genes }) => geneRule.split(/\s+/).map(w => {
-  if (w.match(/and|or/)) {
-    return w;
-  }
-
-  const gene = genes.find(g => g.id === w);
-  return gene.name;
-}).join(' ');
-
 const getReaction = async ({ id, model, version }) => {
   const [m, v] = parseParams(model, version);
 
@@ -68,10 +59,6 @@ RETURN apoc.map.mergeList(COLLECT(value.data)) as reaction
 `;
 
   const reaction = await querySingleResult(statement);
-  if (reaction.geneRule) {
-    reaction.geneRule = formatGeneRule(reaction);
-  }
-
   return { ...reaction, externalDbs: reformatExternalDbs(reaction.externalDbs) };
 };
 
--- a/frontend/src/components/explorer/mapViewer/SidebarDataPanels.vue
+++ b/frontend/src/components/explorer/mapViewer/SidebarDataPanels.vue
@@ -47,7 +47,7 @@
         <header class="card-header is-clickable"
                 @click.prevent="showSelectionCardContent = !showSelectionCardContent">
           <p class="card-header-title is-inline is-capitalized is-unselectable">
-            {{ selectionData.type }}: <i>{{ selectionData.data.id }}</i>
+            {{ selectionData.type }}: <i>{{ selectionData.data.name || selectionData.data.id }}</i>
           </p>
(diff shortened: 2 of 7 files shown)


inghylt (commented):
I think it looks good and that it is legit with changes applied for genes and metabolites. 

One thought, some of the names in the 3d map viewer are very long, .e.g. 1-Docosahexenoyl..etc... Do researchers want all of that info with Deltas, Lpe etc?
mv_3d_27APR21

frontend/src/components/explorer/mapViewer/SidebarDataPanels.vue
```diff
@@ -143,12 +143,12 @@ export default {
...
       selectedElementDataKeys: {
         metabolite: [
-          { name: 'name' },
+          { name: 'id' },
           { name: 'formula' },
           { name: 'compartment' },
         ],
         gene: [
-          { name: 'name', display: 'Gene&nbsp;name' },
+          { name: 'id', display: 'Gene&nbsp;id' },
```
nanjiangshu, line 151: I think the swapping of the `id` and `name` here looks great. 
However, maybe it is better to change `Gene id` to `Id` so that the format is in consistence with metabolite. See line 146 above.

nanjiangshu (commented):
The changes look great. I think both the use of gene name on the `reaction page` and the swapping of `id` and `name` on the sidebar are good changes. For the gene names displayed on the 3D map viewer, some gene names are very long because the alternate names are used. I think the name (the short version) should be used exclusively to avoid the messy display on the 3D map. It is a bit weird though why short names are displayed at some places but long names are used at other places.

frontend/src/components/explorer/mapViewer/SidebarDataPanels.vue
e0: I agree this would be better. @mihai-sysbio was there a reason why it was 'Gene name:' vs 'Name:' for metabolite before?

mihai-sysbio: I can't recall any rationale behind this. It might because how this field is called on Ensembl ("Name"), but on Uniprot it's just "Gene". The way I see it, short and clear is good.

api/src/neo4j/queries/reaction.js
```diff
@@ -2,6 +2,15 @@ import querySingleResult from 'neo4j/queryHandlers/single';
...
 
+const formatGeneRule = ({ geneRule, genes }) => geneRule.split(/\s+/).map(w => {
+  if (w.match(/and|or/)) {
+    return w;
+  }
+
+  const gene = genes.find(g => g.id === w);
+  return gene.name;
+}).join(' ');
+
```
mihai-sysbio, line 13: Wouldn't we want to offload this client-side?

api/src/neo4j/queries/reaction.js
```diff
@@ -59,6 +68,10 @@ RETURN apoc.map.mergeList(COLLECT(value.data)) as reaction
 `;
 
   const reaction = await querySingleResult(statement);
+  if (reaction.geneRule) {
+    reaction.geneRule = formatGeneRule(reaction);
+  }
+
```
mihai-sysbio, line 74: Conceptually, I would move all necessary checks in the gene rule formatting.

```suggestion
  reaction.geneRule = formatGeneRule(reaction);
```

mihai-sysbio:
> I think it looks good and that it is legit with changes applied for genes and metabolites.
> 
> One thought, some of the names in the 3d map viewer are very long, .e.g. 1-Docosahexenoyl..etc... Do researchers want all of that info with Deltas, Lpe etc?
> mv_3d_27APR21

MA-910 does not mention anything about changing the metabolite IDs to metabolite names. As @inghylt notes, having the metabolite names shown might be too annoying as-is, and cropping the name wouldn't be an improvement over showing the IDs. Moreover, the IDs are compartment specific, and this information is lost if only the name is shown.

api/src/neo4j/queries/reaction.js
e0: Good suggestion! I had some problems with this when I tried it before but it seems fine now actually.

api/src/neo4j/queries/reaction.js
e0: We could but since our API is usable outside of the frontend don't we want the response to have the formatting directly? Either way is fine, just let me know.

frontend/src/components/explorer/mapViewer/SidebarDataPanels.vue
e0: Reverting back to displaying gene id in the 3d-viewer as discussed. Addressed in https://github.com/MetabolicAtlas/MetabolicAtlas/pull/565/commits/67a5fe518b9fe553b1538b145061e67876583b8b.

api/src/neo4j/queries/reaction.js
mihai-sysbio: I guess in terms of efficiency one could argue both ways.
For the external API users though I think it's best to also provide an untouched version of the GPR, since that may be used in different workflows that expect the GPR to contain gene IDs.

e0: Ok sounds good. I will completely revert this in the backend then.

e0: Addressed in https://chalmersysbiocsb.atlassian.net/browse/MA-910.

mihai-sysbio (approved):
I haven't tested this (issues with Docker), but the changes make sense and look good.

inghylt (approved):
I think it looks good!

nanjiangshu (commented):
Looks good to me now

PR #567, opened 2021-04-27
feat: improve footer MA-812

inghylt:
- increase size of logos
- change layout with logos equally spaced and centered on the first row, and the copyright notice on a row under
- update WCPR logo link

Also decided to make the copyright text short for mobile view as well as I thought it looked cleaner

### DIFF
--- a/frontend/src/App.vue
+++ b/frontend/src/App.vue
@@ -86,7 +86,7 @@
               <a href="https://kaw.wallenberg.org/" title="Knut and Alice Wallenberg Foundation">
                 <img class="my-0 mx-2" src="/img/wallenberg.gif" />
               </a>
-              <a href="https://www.kth.se/en/bio/centres/wcpr" title="CBH | KTH Royal Institute of Technology">
+              <a href="https://www.kth.se/wcpr" title="CBH | KTH Royal Institute of Technology">
                 <img class="my-0 mx-2" src="/img/wpcr.jpg" />
               </a>
               <a href="https://nbis.se/">
--- a/frontend/src/App.vue
+++ b/frontend/src/App.vue
@@ -73,16 +73,9 @@
     <router-view></router-view>
     <ErrorPanel :message="errorMessage" @hideErrorPanel="errorMessage=''" />
     <footer id="footer" class="footer has-background-primary-lighter is-size-6 py-4">
-      <div class="columns is-gapless">
-        <div class="column is-7-desktop is-5-tablet has-text-centered-mobile">
-          <p>2021 ©
-            <span class="is-hidden-tablet-only">
-              &nbsp;Department of Biology and Biological Engineering |
-            </span>
-            &nbsp;Chalmers University of Technology</p>
-        </div>
-        <div class="column">
-          <div class="content has-text-right has-text-centered-mobile is-size-2-mobile">
+      <div class="columns columns is-gapless is-multiline">
+        <div class="column is-full">
+          <div class="content has-text-centered">
             <p>
               <a href="https://www.sysbio.se" title="SysBio">
                 <img class="my-0 mx-2" src="/img/sysbio-logo.png" />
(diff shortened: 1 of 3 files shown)


frontend/src/App.vue
```diff
@@ -106,6 +99,13 @@
             </p>
           </div>
         </div>
+        <div class="column has-text-centered">
+          <p>2021 ©
+            <span class="is-hidden-touch">
```
e0, line 104:
```suggestion
            <span class="is-hidden-mobile">
```

Do you think this is more appropriate? Here is a list of the `is-hidden-X` helpers: https://bulma.io/documentation/helpers/visibility-helpers/#hide. If you intended to hide it for both mobile and tablet please ignore this.

e0 (commented):
I tested in a few different resolutions and I think it looks good!

Referring to the Jira issue, there is one point I'm unclear about. @mihai-sysbio mentioned to "follow the Elixir look and feel for the footer". Is the footer from https://elixir-europe.org/ meant to be used as an inspiration?

mihai-sysbio (changes requested):
Since the overall height of the `<footer>` increases, the Map Viewer page becomes scrollable now, whereas before it was meant to be full-height. How can this be addressed?

nanjiangshu (commented):
I also feel the height of the footer is a bit too much when we place the logo and text in two rows. As @mihai-sysbio pointed out, on the `MapViewer` page, the text about Chalmers is hidden when loaded. Maybe we can move the text to the same row as the logo as before. Does it still fit in one row with the larger logos?

frontend/src/App.vue
```diff
@@ -73,16 +73,9 @@
...
-        <div class="column is-7-desktop is-5-tablet has-text-centered-mobile">
-          <p>2021 ©
-            <span class="is-hidden-tablet-only">
-              &nbsp;Department of Biology and Biological Engineering |
-            </span>
-            &nbsp;Chalmers University of Technology</p>
-        </div>
-        <div class="column">
-          <div class="content has-text-right has-text-centered-mobile is-size-2-mobile">
+      <div class="columns columns is-gapless is-multiline">
```
mihai-sysbio, line 76: This is a bit unexpected, not only because of the duplicated `columns`, but also of the expected Bulma layout where a `columns` parent has one or more `column` children.

frontend/src/components/explorer/MapViewer.vue
```diff
@@ -141,6 +141,12 @@ export default {
       this.handleQueryParamsWatch(newQuery, oldQuery);
     },
   },
+  async beforeCreate() {
+    const copyright = document.getElementById('footer-copyright');
+    copyright.className = 'column';
+    const logos = document.getElementById('footer-logos');
+    logos.style.display = 'none';
+  },
```
mihai-sysbio, line 149: If two layouts are needed, it is preferred to do this purely in CSS, even if it would imply two alternative designs.

frontend/src/App.vue
inghylt: Woops, nicely spotted. This is a legit fix or were you going for something else as well?
```suggestion
      <div class="columns is-gapless is-multiline">
```

mihai-sysbio: I think that would not be enough - I don't see the html children nodes with the class `column` ([more info on Bulma](https://bulma.io/documentation/columns/basics/)).

frontend/src/components/explorer/MapViewer.vue
```diff
@@ -163,6 +163,16 @@ export default {
     await this.$store.dispatch('maps/getMapsListing', this.model);
     this.$store.dispatch('maps/initFromQueryParams', this.$route.query);
     this.loadMapFromParams();
+    const footer = document.getElementById('footer');
+    footer.querySelector('#footer-logos').style.display = 'none';
+    footer.querySelector('#copyright-standard').style.display = 'none';
+    footer.querySelector('#copyright-mapviewer').style.display = 'block';
```
mihai-sysbio, line 169: These lines should go as high up as possible, since at the moment the footer on the Map Viewer flickers - the users can see the layout change.

frontend/src/App.vue
mihai-sysbio: Sorry, I don't know where I was looking - the children `column` exist.

e0 (approved):

nanjiangshu (approved):
Looks great!

mihai-sysbio (approved):

PR #568, opened 2021-04-29
feat: improve nginx cache conditions

mihai-sysbio:
This PR refactors the use of Docker, according to recent updates. It also revisits the caching rules in Nginx.


nginx/prod.nginx.conf
```diff
@@ -6,7 +6,7 @@ upstream api {
   server api:8081 weight=10 max_fails=3 fail_timeout=30s;
 }
 
-proxy_cache_path /nginx-cache levels=1:2 keys_zone=my_cache:30m inactive=60m use_temp_path=off max_size=4g;
+proxy_cache_path /nginx-cache levels=2:2 keys_zone=my_cache:300m inactive=1y use_temp_path=off max_size=4g;
```
e0, line 9: Why is this changed from `1:2` to `2:2`?

e0: For `16^5` it would need to be `1:2:2` or some permutation of that since only 1 or 2 are allowed. I'm not clear on the optimal number of files per folder so it's difficult for me to say what is enough. Anyways, if there are no performance downsides of having more folders than I don't see why we wouldn't go for the max of `2:2:2`.

mihai-sysbio: Ah right:
> levels of a cache: from 1 to 3, each level accepts values 1 or 2

There performance penalties for traversing more directories, but I'm hoping the machines are now using fast drives. Since `2:2` is not enough, `2:2:2` it is then.

nginx/prod.nginx.conf
```diff
@@ -6,7 +6,7 @@ upstream api {
   server api:8081 weight=10 max_fails=3 fail_timeout=30s;
 }
 
-proxy_cache_path /nginx-cache levels=1:2 keys_zone=my_cache:30m inactive=60m use_temp_path=off max_size=4g;
+proxy_cache_path /nginx-cache levels=2:2:2 keys_zone=my_cache:300m inactive=1y use_temp_path=off max_size=4g;
```
nanjiangshu, line 9: If `inactivte=1y` and `max_size=4g`, does it mean when the size of cache exceeds 4 GB, no data will be cached? If so, it would be better to cache recent visits than to keep very old visits in the cache.

mihai-sysbio: How about 10g? It's not enough if all the cache paths are used, but it should still be plenty:
```suggestion
proxy_cache_path /nginx-cache levels=2:2:2 keys_zone=my_cache:300m inactive=1y use_temp_path=off max_size=10g;
```
I'd rather keep the `1y` unless the storage limit is hit.

mihai-sysbio: @nanjiangshu what do you think?

e0 (approved):
I was waiting for the discussion to resolve but don't think the change would make a breaking difference one way or another so approving this for now.

nginx/prod.nginx.conf
nanjiangshu: OK. I think it is fair enough to set max_size to 10g since the root volume of VM is usually not very big. 
I will approve this PR

nanjiangshu (approved):

PR #569, opened 2021-04-30
fix: force copy in generate-data MA-947

nanjiangshu:
This pull request solves the problem that `generate-data` does not overwrite existing data files in the folder `neo4j/import` on the Linux system.

### DIFF
--- a/proj.sh
+++ b/proj.sh
@@ -7,8 +7,8 @@ function generate-data {
   source .env && yarn --cwd $DATA_GENERATOR_PATH start $DATA_FILES_PATH "$@"
   /bin/cp -rf $DATA_GENERATOR_PATH/data/* neo4j/import
   /bin/cp  -f $DATA_GENERATOR_PATH/data/hpaRna.json api/src/data/
-  /bin/cp -rf $DATA_FILES_PATH/integrated-models/integratedModels.json api/src/data/
-  /bin/cp -rf $DATA_FILES_PATH/gemsRepository.json api/src/data/
+  /bin/cp  -f $DATA_FILES_PATH/integrated-models/integratedModels.json api/src/data/
+  /bin/cp  -f $DATA_FILES_PATH/gemsRepository.json api/src/data/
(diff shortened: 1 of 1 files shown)


mihai-sysbio (approved):
Looks good to me.
What's the advantage of using `/bin/cp` instead of `cp`?

e0:
> Looks good to me.
> What's the advantage of using `/bin/cp` instead of `cp`?

Previous discussion here: https://github.com/MetabolicAtlas/MetabolicAtlas/pull/531#issuecomment-744488040

proj.sh
```diff
@@ -5,13 +5,13 @@ function generate-data {
...
-  /bin/cp    $DATA_GENERATOR_PATH/data/hpaRna.json api/src/data/
-  /bin/cp -r $DATA_FILES_PATH/integrated-models/integratedModels.json api/src/data/
-  /bin/cp -r $DATA_FILES_PATH/gemsRepository.json api/src/data/
-  /bin/cp -r $DATA_FILES_PATH/svg api/
-  /bin/cp -r $DATA_FILES_PATH/repository ftp/
-  /bin/cp -r $DATA_FILES_PATH/repository api/
+  /bin/cp -rf $DATA_GENERATOR_PATH/data/* neo4j/import
+  /bin/cp  -f $DATA_GENERATOR_PATH/data/hpaRna.json api/src/data/
+  /bin/cp -rf $DATA_FILES_PATH/integrated-models/integratedModels.json api/src/data/
+  /bin/cp -rf $DATA_FILES_PATH/gemsRepository.json api/src/data/
```
e0, line 11: Quite minor but should it be `-f` for lines 10 and 11 as well?

e0 (approved):
Looks good @nanjiangshu. Left a comment, I'm pretty sure either way is fine in practice but just wanted to point it out since it was different for the same type of use case.

proj.sh
nanjiangshu: Good catch @shan-chalmers :) Fixed now.

inghylt (approved):

PR #570, opened 2021-05-04
feat: enable panning using arrow keys for 2D maps MA-598

e0:
Jira issue MA-598

This PR introduces the ability to pan 2D maps by using the arrow keys. During testing I found that 10 is a good unit of movement for different maps and zoom levels, but please let me if you feel like it does not work well enough. Note that you can also hold down an arrow key for continuous panning of the map.

### DIFF
--- a/frontend/src/components/explorer/mapViewer/Svgmap.vue
+++ b/frontend/src/components/explorer/mapViewer/Svgmap.vue
@@ -153,6 +153,37 @@ export default {
       this.$store.dispatch('maps/setLoading', true);
       const payload = { model: this.model.short_name, svgName: this.mapData.svgs[0].filename };
       await this.$store.dispatch('maps/getSvgMap', payload);
+      this.bindKeyboardShortcuts();
+    },
+    bindKeyboardShortcuts() {
+      document.addEventListener('keydown', (event) => {
+        const key = event.key || event.keyCode;
+
+        switch (key) {
+          case 'ArrowLeft':
+          case 37:
+            this.relativePan(-10, 0);
+            break;
+          case 'ArrowUp':
+          case 38:
+            this.relativePan(0, -10);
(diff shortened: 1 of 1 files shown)


frontend/src/components/explorer/mapViewer/Svgmap.vue
```diff
@@ -153,6 +153,37 @@ export default {
...
+          case 38:
+            this.relativePan(0, -10);
+            break;
+          case 'ArrowRight':
+          case 39:
+            this.relativePan(10, 0);
+            break;
+          case 'ArrowDown':
+          case 40:
+            this.relativePan(0, 10);
```
mihai-sysbio, line 177: How about defining a constant for the pan distance?

```suggestion
        const key = event.key || event.keyCode;
        const panDistance = 10;
        switch (key) {
          case 'ArrowLeft':
          case 37:
            this.relativePan(-panDistance, 0);
            break;
          case 'ArrowUp':
          case 38:
            this.relativePan(0, -panDistance);
            break;
          case 'ArrowRight':
          case 39:
            this.relativePan(panDistance, 0);
            break;
          case 'ArrowDown':
          case 40:
            this.relativePan(0, panDistance);
```

mihai-sysbio (approved):
Looks good to me.

nanjiangshu (approved):
Looks great.

frontend/src/components/explorer/mapViewer/Svgmap.vue
e0: Great suggestion!

inghylt:
Is the intention to have the map moving right when pressing left arrow, moving down when pressing up arrow etc? I don't expect an "inverted behavior" when pressing the arrow keys, but maybe that's just me

PR #571, opened 2021-05-05
feat: data overlay reaction and met

mihai-sysbio:
I'd be strategic if we can integrate as much as possible from this work done by @pecholleyc before we start refactoring the _Data overlay_.


e0:
Looks like there was some interesting work here but a lot of conflicting changes has happened since the last commit. I think the best approach is to manually go through each conflicting file and make the necessary updates. This is definitely still work in progress so converting the PR to draft for now.

mihai-sysbio:
As suggest by @e0, @nanjiangshu and @inghylt - let's close this and keep track of the issue for the time being.

PR #572, opened 2021-05-06
fix: api url for files in gem repository MA-958

mihai-sysbio:
When trying to download files from the GEM repository, the urls that are computed in the frontend appear to be wrong, and thus not working, potentially due to previous refactoring.


nanjiangshu (approved):
Looks good to me :)

inghylt (approved):
Nice work!

PR #574, opened 2021-05-11
feat: page for external ids MA-918

### LINKED ISSUE #575: page for external ids
As a user, I would like to see all the entities on Met Atlas that correspond to an external identifier, so that I can see if it is used in any of the integrated models and link to it directly.

For more context, a guiding example is that `identifiers.org/metabolicatlas.externalid:dbname+id` would link to `metabolicatlas.org/externalid/dbname/id` . The setting up of this on identifiers.org  is a different card.

inghylt:
Originally a draft to get feedback, this PR now closes #575

This PR was created by @e0 and @inghylt

### DIFF
--- a/frontend/src/components/ExternalDb.vue
+++ b/frontend/src/components/ExternalDb.vue
@@ -1,13 +1,13 @@
 <template>
   <section class="section extended-section">
     <div v-if="externalDb" class="container is-fullhd">
-      <h3 class="title is-3">
+      <h3 class="title is-3 mb-2">
         {{ externalDb.dbName }} {{ components[0].componentType }} {{ externalDb.externalId }}
       </h3>
-      <h4 class="subtitle">
+      <p class="my-3">
         This database identifier is associated with the following Metabolic Atlas
         {{ components.length === 1 ? 'component' : 'components' }}:
-      </h4>
+      </p>
       <div class="tags are-medium is-flex-direction-column is-align-items-flex-start">
         <span v-for="c in components" :key="c.id + c.model + c.version">
           {{ c.componentType }}
--- a/frontend/src/components/ExternalDb.vue
+++ b/frontend/src/components/ExternalDb.vue
@@ -5,9 +5,8 @@
         {{ externalDb.dbName }} {{ components[0].componentType }} {{ externalDb.externalId }}
       </h3>
       <h4 class="subtitle">
-        The following Metabolic Atlas
-        {{ components.length === 1 ? 'component is' : 'components are' }}
-        associated with this external DB reference.
+        This database identifier is associated with the following Metabolic Atlas
+        {{ components.length === 1 ? 'component' : 'components' }}:
       </h4>
       <div class="tags are-medium is-flex-direction-column is-align-items-flex-start">
         <span v-for="c in components" :key="c.id + c.model + c.version" class="tag is-light">
--- a/frontend/src/components/ExternalDb.vue
+++ b/frontend/src/components/ExternalDb.vue
@@ -1,22 +1,27 @@
 <template>
-  <div class="section extended-section">
-    <template v-if="externalDb">
-      <h3 class="title is-3">
-        References to external id {{ externalDb.externalId }}
+  <section class="section extended-section">
+    <div v-if="externalDb" class="container is-fullhd">
+      <h3 class="title">
+        {{ externalDb.dbName }} {{ components[0].componentType }} {{ externalDb.externalId }}
       </h3>
-      <h6 v-if="externalDb.url" class="subtitle is-6">
-        Link: <a :href="externalDb.url" target="_blank"> {{ externalDb.url }} </a>
-      </h6>
-      <ul id="example-1">
-        <li v-for="c in components" :key="c.id">
+      <h4 class="subtitle">
+        The following Metabolic Atlas
+        {{ components.length === 1 ? 'component is' : 'components are' }}
+        associated with this external DB reference.
+      </h4>
+      <div class="tags are-medium is-flex-direction-column is-align-items-flex-start">
+        <span v-for="c in components" :key="c.id + c.model + c.version" class="tag is-light">
           <router-link :to="{ name: c.componentType.toLowerCase(), params: { model:
             c.model, id: c.id } }">
             {{ c.componentType }} {{ c.id }} from {{ c.model }} {{ c.version }}
           </router-link>
-        </li>
-      </ul>
-    </template>
-  </div>
+        </span>
+      </div>
+      <p v-if="externalDb.url">
+        For more details, visit <a :href="externalDb.url" target="_blank"> {{ externalDb.url }} </a>.
+      </p>
+    </div>
+  </section>
 </template>
 
 <script>
--- a/frontend/src/components/ExternalDb.vue
+++ b/frontend/src/components/ExternalDb.vue
@@ -11,10 +11,12 @@
       <div class="tags are-medium is-flex-direction-column is-align-items-flex-start">
         <span v-for="c in components" :key="c.id + c.model + c.version">
           {{ c.componentType }}
-          <router-link :to="{ name: c.componentType.toLowerCase(), params: { model:
-            c.model, id: c.id } }">
-            <span class="tag is-light">{{ c.id }}</span>
-          </router-link>
+          <span class="tag is-light">
+            <router-link :to="{ name: c.componentType.toLowerCase(), params: { model:
+              c.model, id: c.id } }">
+              {{ c.id }}
+            </router-link>
+          </span>
           from {{ c.model }} {{ c.version }}
         </span>
       </div>
--- a/frontend/src/components/ExternalDb.vue
+++ b/frontend/src/components/ExternalDb.vue
@@ -8,10 +8,10 @@
         Link: <a :href="externalDb.url" target="_blank"> {{ externalDb.url }} </a>
       </h6>
       <ul id="example-1">
-        <li v-for="component in components" :key="component.id">
-          <router-link :to="{ name: component.componentType.toLowerCase(), params: { model:
-            component.model.replace('Gem', '-GEM'), id: component.id } }">
-            {{ component.componentType }} {{ component.id }} from {{ component.model }}
+        <li v-for="c in components" :key="c.id">
+          <router-link :to="{ name: c.componentType.toLowerCase(), params: { model:
+            c.model, id: c.id } }">
+            {{ c.componentType }} {{ c.id }} from {{ c.model }} {{ c.version }}
           </router-link>
         </li>
       </ul>
--- a/frontend/src/components/ExternalDb.vue
+++ b/frontend/src/components/ExternalDb.vue
@@ -1,19 +1,21 @@
 <template>
   <div class="section extended-section">
-    <h3 class="title is-3">
-      References to external id {{ externalDb.externalId }}
-    </h3>
-    <h6 v-if="externalDb.url" class="subtitle is-6">
-      Link: <a :href="externalDb.url" target="_blank"> {{ externalDb.url }} </a>
-    </h6>
-    <ul id="example-1">
-      <li v-for="component in components" :key="component.id">
-        <router-link :to="{ name: component.componentType.toLowerCase(), params: { model:
-          component.model.replace('Gem', '-GEM'), id: component.id } }">
-          {{ component.componentType }} {{ component.id }} from {{ component.model }}
-        </router-link>
-      </li>
-    </ul>
+    <template v-if="externalDb">
+      <h3 class="title is-3">
+        References to external id {{ externalDb.externalId }}
+      </h3>
+      <h6 v-if="externalDb.url" class="subtitle is-6">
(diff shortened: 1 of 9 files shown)


e0:
Great stuff and very actionable points @mihai-sysbio !

---

> * the word "Link" should be avoided, perhaps "go to {{ dbname }}" with a link icon
> * there is no value in showing the full link to the user, so it should be a redirect on-click

We debated this a bit during the implementation and thought it could be useful to display the full link. If we don't show the full link at all maybe we can turn the title into a link instead.

---

> * the models should be listed with their "original" name instead of the raw dbname, ie `Rat-GEM` instead of `RatGem` (since this is already implemented in the codebase, perhaps this can be a shared helper function)

This could be done either in the frontend or backend. Any preferences?

---

> * the model version should also be listed

Sounds good. Want to add that for implementation, we currently don't include the version in the URLs for gem browser components. So for now it would be enough to list the version but for the future we should keep in mind to update this page when we start supporting multiple versions of the same model.

mihai-sysbio:
> We debated this a bit during the implementation and thought it could be useful to display the full link. If we don't show the full link at all maybe we can turn the title into a link instead.

There is value in providing users with the link, but it doesn't have to be directly under the title. On the desktop, users can  hover over to see the link. Otherwise they can right click or long press to copy the link. If you still think they should see it, could it be placed somewhere less "prime"?

> This could be done either in the frontend or backend. Any preferences?

Could it be done entirely in the backend?

> it would be enough to list the version

👍🏻

e0:
Thanks again for the feedback @mihai-sysbio . @inghylt and I have made some updates. Please take another look when you have time. Here is a screenshot for convenience:

image

frontend/src/components/ExternalDb.vue
```diff
@@ -0,0 +1,51 @@
+<template>
+  <section class="section extended-section">
+    <div v-if="externalDb" class="container is-fullhd">
+      <h3 class="title">
+        {{ externalDb.dbName }} {{ components[0].componentType }} {{ externalDb.externalId }}
+      </h3>
+      <h4 class="subtitle">
+        The following Metabolic Atlas
+        {{ components.length === 1 ? 'component is' : 'components are' }}
+        associated with this external DB reference.
```
mihai-sysbio, line 10:
```suggestion
        This external database reference is associated with the following Metabolic Atlas component(s):
```

frontend/src/router.js
```diff
@@ -39,6 +40,7 @@ const routes = [
   { path: '/gems/comparison', name: 'comparemodels', component: CompareModels },
   { path: '/resources', name: 'resources', component: Resources },
   { path: '/documentation', name: 'documentation', component: Documentation },
+  { path: '/external-db/:dbName/:externalId', name: 'externalDb', component: ExternalDb },
```
mihai-sysbio, line 43: Thinking more about this, it would make sense to rephrase "external-db" to "identifier", at least in all places that the user sees. At least I'm not seeing the value in specifying "external", as it can lead to confusion (eg "external to whom?").

frontend/src/components/ExternalDb.vue
```diff
@@ -0,0 +1,51 @@
...
+      </h4>
+      <div class="tags are-medium is-flex-direction-column is-align-items-flex-start">
+        <span v-for="c in components" :key="c.id + c.model + c.version" class="tag is-light">
+          <router-link :to="{ name: c.componentType.toLowerCase(), params: { model:
+            c.model, id: c.id } }">
+            {{ c.componentType }} {{ c.id }} from {{ c.model }} {{ c.version }}
+          </router-link>
+        </span>
```
mihai-sysbio, line 18: How about restricting the tag highlight just to the id?

```suggestion
        <span v-for="c in components" :key="c.id + c.model + c.version">
          <router-link :to="{ name: c.componentType.toLowerCase(), params: { model:
            c.model, id: c.id } }">
            {{ c.componentType }}
            <span  class="tag is-light">{{ c.id }}</span>
            from {{ c.model }} {{ c.version }}
          </router-link>
        </span>
```

mihai-sysbio:
Overall this is really cool 👍🏻  I've left some minor feedback, not sure if I should already approve the PR.

frontend/src/components/ExternalDb.vue
```diff
@@ -0,0 +1,51 @@
+<template>
+  <section class="section extended-section">
+    <div v-if="externalDb" class="container is-fullhd">
+      <h3 class="title">
```
mihai-sysbio, line 4: Last time I checked, this is what the Bulma guideline, also matching the rest of our codebase:

```suggestion
      <h3 class="title is-3">
```

frontend/src/components/ExternalDb.vue
```diff
@@ -0,0 +1,51 @@
...
+        {{ externalDb.dbName }} {{ components[0].componentType }} {{ externalDb.externalId }}
+      </h3>
+      <h4 class="subtitle">
```
mihai-sysbio, line 7: To avoid having this text extra large, this change would match the style of the rest of our pages:
```suggestion
      <p>
```

frontend/src/components/ExternalDb.vue
```diff
@@ -0,0 +1,51 @@
...
+        The following Metabolic Atlas
+        {{ components.length === 1 ? 'component is' : 'components are' }}
+        associated with this external DB reference.
+      </h4>
```
mihai-sysbio, line 11: Matching end tag for the paragraph, and some extra space that is now gone after removing the `subtitle` class.
```suggestion
      </p>
      <br>
```

mihai-sysbio:
Overall this is really cool 👍🏻  I've left some minor feedback, not sure if I should already approve the PR.

frontend/src/components/ExternalDb.vue
e0: I like this suggestion! I would prefer to do the work for the user and handle the pluralizing for them though, so we show the extra `s` conditionally.

frontend/src/router.js
e0: Good suggestion although I'm not sure about having `identifier` as the route name. I personally might prefer something like `db-reference`. Would be good to get some more opinion on this as well. @inghylt @nanjiangshu

frontend/src/components/ExternalDb.vue
e0: Can you point me to the guideline please? I did this since `is-3` is the default for `is-title` (see https://bulma.io/documentation/elements/title/#sizes). If it's a part of the guideline then I'm all for it but if not I would actually want to open up a new issue to remove it from the rest of the codebase.

frontend/src/components/ExternalDb.vue
e0: Sounds good! @inghylt also suggested something along these lines with an additional change.

frontend/src/components/ExternalDb.vue
e0: Since the rest of the link is blue text, this contrast actually makes the id look like it's the only part of the link that is not a link. I personally prefer having the whole thing as a tag, or if we only tag the id I suggest making the rest of the line not a part of the link.

frontend/src/components/ExternalDb.vue
mihai-sysbio: It was right there at the link you shared:
image
Since it's the default value I guess we could skip it, but for clarity and consistency I vote for having it.
Also, I meant that at the moment this is how things are in the rest of the codebase - a heading tag with class `title` also has the size specified. I might be wrong, but I have the memory of having done this check at some point, so there should be no need of opening an issue.

e0: Ah I thought you meant specifically that there was documentation that mentions you should always include the default modifiers. I feel like it's good practice to rely on the framework and avoid default modifiers when possible but it's not a strong preference. So we can merge this change and not open a new issue.

@inghylt might be working on this branch atm so I will leave it for her to merge this suggestion to avoid potential branch conflicts.

frontend/src/components/ExternalDb.vue
mihai-sysbio: We have some tags in the GEM Browser, but they are used for the `id` only. so yeah it would make sense to make the rest of the text black.

frontend/src/router.js
inghylt: I'm up for changing it! I think there is a value to somehow emphasize that the id:s are from separate databases so users understands what's metatlas specific and what's not. So would say I'm more in favour of `db-reference` or similar

mihai-sysbio: I see all models as independent of Metabolic Atlas, so I don't see much of a difference between what is "internal" and what is "external" - all of the source files are (at least supposed to) come straight from the model repository.

mihai-sysbio: How about `database-reference` or `database-identifier`?

inghylt: Ah, I see. I feel like the models are very close in relation to metatlas compared to the other databases, but I get what you mean. I would go for `database-identifier`, feels more in line with the usage of ID etc. on the web page

frontend/src/components/ExternalDb.vue
inghylt: image
Looks good like this?

mihai-sysbio: It does to me 👍🏻

frontend/src/components/ExternalDb.vue
inghylt: image
Just replacing it with `<p>` made the spacing between the elements quite weird IMO so I played around with the margins a bit. Does this look good to you?

inghylt: @e0 haha yeah I was referring to the spacing between the elements. Thanks for input :+1:

frontend/src/components/ExternalDb.vue
e0: Just realized that in the original implementation the text inside the tag give a stronger indication that they are links. Was this style intentionally removed? I mentioned a thought about spacing in [another comment](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/574#discussion_r640728837) but even if we ignore that and only have the tag on the id/name, I think it would still look better if the tags look like links. Thoughts?

image

inghylt: Pushed a commit fixing so that the text inside the tag look like a link :+1:

frontend/src/router.js
inghylt: Suggestion: how about only `identifiers`? Then we kind of mimic identifiers.org which me and Nanjiang think makes sense

mihai-sysbio: It would make sense to me too, it's along the lines of what I suggested initially.

mihai-sysbio:
Alright, to me there are the outstanding items to discuss before being ready to merge this in:

- the term `CompartmentalizedMetabolite` is unknown to the user, so it should be replaced with the more familiar `Metabolite`
- can we do the switch from `external-db` to `identifier`?
- the list is visually hard to dissociate from the rest of the content (first introductory line, last line with link to identifiers.org); one way to solve this would be to bring back the list, or use some Bulma styling for lists
- do you like the idea to make the tag `hoverable` as below (green border)?

In the meantime, I'll go ahead and approve this.

image

mihai-sysbio (approved):

e0:
> Alright, to me there are the outstanding items to discuss before being ready to merge this in:
> 
> * the term `CompartmentalizedMetabolite` is unknown to the user, so it should be replaced with the more familiar `Metabolite`
> * can we do the switch from `external-db` to `identifier`?
> * the list is visually hard to dissociate from the rest of the content (first introductory line, last line with link to identifiers.org); one way to solve this would be to bring back the list, or use some Bulma styling for lists
> * do you like the idea to make the tag `hoverable` as below (green border)?
> 
> In the meantime, I'll go ahead and approve this.
> 
> image

Thanks for the new feedback @mihai-sysbio ! @inghylt and I have gone through them and made some changes. We thought the tags with the link styling already make it clear that they are clickable, so we decided to not implement the green hover effect.

nanjiangshu (commented):
This PR looks great to me. I also think the text of the reaction with hyperlink looks better without the green hover. I have only one comment regarding the order of GEMs 
e.g. for http://localhost/identifier/BiGG/PPNCL3, it shows
```
Reaction  PPNCL3 from Rat-GEM 1.1.0
Reaction  PPNCL3 from Mouse-GEM 1.1.0
Reaction  PPNCL3 from Human-GEM 1.6.0
```
I don't know in which order are the reactions listed for different GEMs. Maybe we can order it alphabetically as done in the Explore (http://localhost/explore) page?

mihai-sysbio:
>order it alphabetically as done in the Explore page?

@nanjiangshu very good observation.

Since now there are more and more places where results could/should be sorted alphabetically, how about we create an issue to revise the API and implement these were adequate? It feels like this could be a self-contained issue, without affecting the frontend at all.

frontend/src/router.js
inghylt: Created issue #608, resolving conversation

e0 (approved):

PR #576, opened 2021-05-18
feat: disable data sources for non Human GEMs MA-956

inghylt:
Removed functionality to use data overlay from protein atlas for non human models, but kept the functionality to use custom gene expression data.

Note, I haven't tried using custom gene expression data for neither human or non human models, so if you have the possibility I think it would be nice to verify that everything works as intended


e0 (approved):
Looks great @inghylt! It's nice that you made the labels conditional as well.

nanjiangshu (approved):
Fixed very nicely with the function `dataSourcesAvailable`

mihai-sysbio (approved):
Very nice fix indeed!

PR #585, opened 2021-05-18
fix: correct variable name for the external db MA-948

nanjiangshu:
This pull request fixes the bug that only one record is shown for the external database table. All records in one external database entry should have separate links as well. The related Jira issue is [MA-948](https://chalmersysbiocsb.atlassian.net/browse/MA-948). 

Examples to check

1. http://localhost/explore/Human-GEM/gem-browser/metabolite/m00016m
The external database `MetaNetX` should have two records 

2. http://localhost/explore/Human-GEM/gem-browser/gene/ENSG00000000419
The external database `Ensembl protein` should have four records (with the new TSV database)


e0 (approved):
Looks good!

inghylt (approved):
Nice work and PR description!

mihai-sysbio (approved):
Excellent!

Issue #580, opened 2021-05-18
Improve UX for first time users of Interaction Partners


e0:
When at https://metabolicatlas.org/ and pressing Interaction Partners the model and id are “hardcoded“ to be `Human-GEM` and `m02007p`. That leads to https://metabolicatlas.org/explore/Human-GEM/interaction-partners/m02007p, i.e. a page with no info on how to choose a different metabolite or information about the tool. For first time users, it would be beneficial with a “choose random metabolite” or short info regarding using the search bar + interaction icon and link to the documentation. 

Preferably, the “don’t make me think“ mentality should be kept in mind when designing this.

e0:
<img width="563" alt="image" src="https://user-images.githubusercontent.com/423498/118652281-bb706b00-b7e6-11eb-9bb7-596fd8a7fc12.png">

inghylt:
Fixed with pull request #587

PR #587, opened 2021-05-19
feat: improve first time users experience of Interaction Partners by removing hardcoded component id MA-941 

inghylt:
When at https://metabolicatlas.org/ and pressing Interaction Partners the component id is hardcoded to 'm02007p'. For first time users it makes more sense to use the landing page for the Interaction Partners that is similar to the GEM Browser where they can choose a random component

closes issue #580


e0 (approved):

nanjiangshu (approved):
Looks great to me.

mihai-sysbio (approved):

PR #588, opened 2021-05-19
feat: code analysis

mihai-sysbio:
This PR integrates CodeQL through a GH Action, which then automatically highlights issues in the `Security` tab of the repository.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,71 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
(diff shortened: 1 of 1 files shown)


nanjiangshu:
Hi @mihai-sysbio, I think this is a valuable addition to the professionality of our codebase. However, it will be great to see an example when an issue is highlighted as mentioned by @shan-chalmers.

mihai-sysbio:
It looks like the details are not publicly accessible, so here are some screenshots:

image
image
image

e0:
> It looks like the details are not publicly accessible, so here are some screenshots:

Thanks for providing the screenshots. Would the details be available for contributors of this repo?

e0 (approved):

mihai-sysbio:
> Thanks for providing the screenshots. Would the details be available for contributors of this repo?

Yes, I think that's how we are supposed to make use of it.

Issue #583, opened 2021-05-18
GEM Browser shows broken image when ChEBI image does not exist


e0:
When the ChEBI entry exists but ChEBI image does not, the GEM-Browser still shows the ChEBI frame but with broken image. Maybe we can display it in a better way. 

An example is for the metabolite `m02576c`, url: http://localhost/explore/Human-GEM/gem-browser/metabolite/m02576c.  

This will be solved by pretending there is no ChEBI id.
![image](https://user-images.githubusercontent.com/423498/118655769-27080780-b7ea-11eb-9c91-86ffb97f1b53.png)

e0:
<img width="542" alt="image" src="https://user-images.githubusercontent.com/423498/118655919-4c951100-b7ea-11eb-9b0e-a2021bbdfc42.png">

inghylt:
Fixed with pull request #589

PR #589, opened 2021-05-19
fix: only show ChEBI image div if ChEBI image exist MA-943

inghylt:
Issue #583

Fix so no broken ChEBI images are shown

Examples:
With image
http://localhost/explore/Human-GEM/gem-browser/metabolite/CE5122_p

With broken image
http://localhost/explore/Human-GEM/gem-browser/metabolite/m02576c

With no ChEBI entry at all
http://localhost/explore/Human-GEM/gem-browser/metabolite/m02865c

Please see if the ChEBI image check is placed suitably in method setup() (line165), and that there are no undesired side effects such as slow response time

### DIFF
--- a/frontend/src/components/explorer/gemBrowser/Metabolite.vue
+++ b/frontend/src/components/explorer/gemBrowser/Metabolite.vue
@@ -67,7 +67,7 @@
               </div>
               <ExtIdTable :type="type" :external-dbs="metabolite.externalDbs"></ExtIdTable>
             </div>
-            <div v-if="metabolite && metabolite.externalDbs && metabolite.externalDbs.ChEBI"
+            <div v-if="chebiImagePresent"
                  class="column is-3-widescreen is-2-desktop is-full-tablet has-text-centered px-2">
               <a :href="metabolite.externalDbs.ChEBI[0].url" target="_blank">
                 <img id="chebi-img" :src="`https://www.ebi.ac.uk/chebi/displayImage.do?defaultImage=true&imageIndex=0&chebiId=${metabolite.externalDbs.ChEBI[0].id.slice(6)}&dimensions=400`" class="hoverable" />
@@ -97,6 +97,7 @@
 </template>
 
 <script>
+import axios from 'axios';
 import { mapState } from 'vuex';
 import MapsAvailable from '@/components/explorer/gemBrowser/MapsAvailable';
 import ReactionTable from '@/components/explorer/gemBrowser/ReactionTable';
@@ -138,6 +139,7 @@ export default {
       showLoaderMessage: '',
       modelNotFound: false,
       messages,
+      chebiImagePresent: false,
     };
   },
   computed: {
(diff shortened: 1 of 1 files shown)


frontend/src/components/explorer/gemBrowser/Metabolite.vue
```diff
@@ -67,7 +67,7 @@
               </div>
               <ExtIdTable :type="type" :external-dbs="metabolite.externalDbs"></ExtIdTable>
             </div>
-            <div v-if="metabolite && metabolite.externalDbs && metabolite.externalDbs.ChEBI"
+            <div v-if="chebiImagePresent"
```
e0, line 70: What do you think about using something called `chebiImageLink` instead? It could be set as null by default and then in the `setup` method you could set it equal to the link if we get data back. Something like:

```javascript
const link = `https://www.ebi.ac.uk/chebi/displayImage.do?defaultImage=true&imageIndex=0&chebiId=${this.metabolite.externalDbs.ChEBI[0].id.slice(6)}`;
const { data } = await axios.get(link);
if (data !== '') {
  this.chebiImageLink = link;
}
```

This can then of coursed be reused in the img's `src` as well. I was originally going to suggest including `dimensions=400` as well in `setup` but it looks like adding that param triggers a CORS error.

e0 (approved):
Looks good! Thanks for including links for the different cases, makes it really easy to test.

I left a refactoring suggestion in a comment but it's not needed since I think the solution is good enough. So I'm approving this PR and you can decide if you want to make the change or not.

frontend/src/components/explorer/gemBrowser/Metabolite.vue
```diff
@@ -166,6 +168,10 @@ export default {
         const payload = { model: this.model, id: this.metaboliteId };
         await this.$store.dispatch('metabolites/getMetaboliteData', payload);
         this.componentNotFound = false;
+        if (this.metabolite.externalDbs.ChEBI) {
+          const { data } = await axios.get(`https://www.ebi.ac.uk/chebi/displayImage.do?defaultImage=true&imageIndex=0&chebiId=${this.metabolite.externalDbs.ChEBI[0].id.slice(6)}`);
```
nanjiangshu, line 172: I think it will be better to define the url for ChEBI as a variable so that we do not need to change it at multiple places when this url is changed.

nanjiangshu (commented):
Hi @inghylt , It looks great and very well written PR description. 
I just left one comment regarding the dynamic use of ChEBI url.

frontend/src/components/explorer/gemBrowser/Metabolite.vue
inghylt: Ah ofc, great feedback!

frontend/src/components/explorer/gemBrowser/Metabolite.vue
inghylt: Really nice idea! Made a commit incorporating this, was able to add dimension parameter as well

e0: Nicely done, much better to have the link concatenation in the method!

nanjiangshu (approved):
Looks great now!

mihai-sysbio:
Is there a way to avoid making 2 requests to ChEBI, especially that they are not cached?

PR #596, opened 2021-05-26
fix: ftp server access

### LINKED ISSUE #595: broken FTP server
As a user, I want to download all model files easily, so that I can do my research. At the moment, the FTP server is unavailable - logging in just hangs.

mihai-sysbio:
This PR closes #595 by using a dedicated subdomain for the ftp server, in order to set it as unproxied in the DNS settings. Moreover, since this changes the FTP host name, it was replaced by the IP of the VM. The fix was deployed and has been verified.
Potentially one could use a variable for this, but then this variable would have to be used both in Vue and in the Dockerfile.

### DIFF
--- a/frontend/src/components/Documentation.vue
+++ b/frontend/src/components/Documentation.vue
@@ -255,7 +255,7 @@ export default {
   data() {
     return {
       multipleColors,
-      ftpUrl: 'metabolicatlas.org',
+      ftpUrl: 'ftp.metabolicatlas.org',
     };
   },
 };
--- a/ftp/Dockerfile
+++ b/ftp/Dockerfile
@@ -6,4 +6,4 @@ RUN useradd -d /var/ftp -s /sbin/nologin ftp
 COPY ./prep-cert.sh /project/prep-cert.sh
 RUN chmod +x /project/prep-cert.sh
 
-CMD /project/prep-cert.sh && /run.sh pure-ftpd -4 -A -c 50 -C 5 -e -G -H -i -p $FTP_MIN_PORT:$FTP_MAX_PORT -P $PUBLICHOST -r -R -s -X -x -Z $FLAGS
+CMD /project/prep-cert.sh && /run.sh pure-ftpd -4 -A -c 50 -C 5 -e -G -H -i -p $FTP_MIN_PORT:$FTP_MAX_PORT -P 129.16.69.158 -r -R -s -X -x -Z $FLAGS
--- a/ftp/Dockerfile
+++ b/ftp/Dockerfile
@@ -6,4 +6,4 @@ RUN useradd -d /var/ftp -s /sbin/nologin ftp
 COPY ./prep-cert.sh /project/prep-cert.sh
 RUN chmod +x /project/prep-cert.sh
 
-CMD /project/prep-cert.sh && /run.sh pure-ftpd -4 -A -c 50 -C 5 -e -G -H -i -p $FTP_MIN_PORT:$FTP_MAX_PORT $FTP_IP -P 129.16.69.158 -r -R -s -X -x -Z $FLAGS
+CMD /project/prep-cert.sh && /run.sh pure-ftpd -4 -A -c 50 -C 5 -e -G -H -i -p $FTP_MIN_PORT:$FTP_MAX_PORT -P $FTP_IP -r -R -s -X -x -Z $FLAGS
--- a/ftp/Dockerfile
+++ b/ftp/Dockerfile
@@ -6,4 +6,4 @@ RUN useradd -d /var/ftp -s /sbin/nologin ftp
 COPY ./prep-cert.sh /project/prep-cert.sh
 RUN chmod +x /project/prep-cert.sh
 
-CMD /project/prep-cert.sh && /run.sh pure-ftpd -4 -A -c 50 -C 5 -e -G -H -i -p $FTP_MIN_PORT:$FTP_MAX_PORT -P 129.16.69.158 -r -R -s -X -x -Z $FLAGS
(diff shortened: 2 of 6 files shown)


frontend/src/components/Documentation.vue
```diff
@@ -255,7 +255,7 @@ export default {
   data() {
     return {
       multipleColors,
-      ftpUrl: 'metabolicatlas.org',
+      ftpUrl: 'ftp.metabolicatlas.org',
```
nanjiangshu, line 258: I received a `Your connection is not private` error when trying to access https://ftp.metabolicatlas.org/. After accepting the risk, it just directed me to the metatlas homepage. Maybe the DNS settings has not taken the effect yet?

ftp/Dockerfile
```diff
@@ -6,4 +6,4 @@ RUN useradd -d /var/ftp -s /sbin/nologin ftp
 COPY ./prep-cert.sh /project/prep-cert.sh
 RUN chmod +x /project/prep-cert.sh
 
-CMD /project/prep-cert.sh && /run.sh pure-ftpd -4 -A -c 50 -C 5 -e -G -H -i -p $FTP_MIN_PORT:$FTP_MAX_PORT -P $PUBLICHOST -r -R -s -X -x -Z $FLAGS
+CMD /project/prep-cert.sh && /run.sh pure-ftpd -4 -A -c 50 -C 5 -e -G -H -i -p $FTP_MIN_PORT:$FTP_MAX_PORT -P 129.16.69.158 -r -R -s -X -x -Z $FLAGS
```
nanjiangshu, line 9: Do you mean if we change `129.16.69.158` to something like `$PUBLICFTPHOST`, it should be changed in the Vue as well? I feel like this `PUBLICFTPHOST` only needs to be configured in the `.env`.

frontend/src/components/Documentation.vue
mihai-sysbio: > @mihai-sysbio When I used the default `macOS` method (Finder -> Connect to Server -> Connect As Guest), it doesn't seem to be able to make the connection. Does that work for you?

It worked for me, but I think I forgot to purge the cache.

mihai-sysbio: > I received a `Your connection is not private` error when trying to access https://ftp.metabolicatlas.org/. After accepting the risk, it just directed me to the metatlas homepage. Maybe the DNS settings has not taken the effect yet?

I tested this and it worked for me, the TTL was 2 min.

nanjiangshu: I see. It works with either FTP client and in Mac Finder when connected as Guest user.

e0: @nanjiangshu Good that you can confirm it works for you. It still doesn't connect for me but maybe I have some local settings that's preventing it. I consider this solved now since both of you can verify it.

e0 (changes requested):
I was going to approve this since it's quite urgent but since it's already deployed now and the problem is fixed, I think we could address the issue @nanjiangshu mentioned about the hardcoded IP before merging.

ftp/Dockerfile
mihai-sysbio: The previous value for `$PUBLICHOST` was `metabolicatlas.org`, which was incompatible with the FTP server defined at `ftp.metabolicatlas.org`, even though it is the same IP. So, I suspect this variable would need to be the exact same in both the _Documentation_ page (Vue) and in the `pure-ftpd` command. Sharing variables like this is definitely possible but not immediately straightforward - I would suggest to capture it in an independent issue. Or perhaps somebody else would want to implement it in this branch?

e0: If we don't want to override `$PUBLICHOST` we could (and probably should) just introduce a new variable in `.env` to represent this IP address. This is just so we can handle potential IP changes and support multiple environments. I don't think this would introduce any incompatibility.

nanjiangshu: I agree and this is exactly what I mentioned in https://github.com/MetabolicAtlas/MetabolicAtlas/pull/596#discussion_r639446693 to use a new variable for instance `$PUBLICFTPHOST`.

nanjiangshu (approved):
Looks good to me now!

ftp/Dockerfile
```diff
@@ -6,4 +6,4 @@ RUN useradd -d /var/ftp -s /sbin/nologin ftp
 COPY ./prep-cert.sh /project/prep-cert.sh
 RUN chmod +x /project/prep-cert.sh
 
-CMD /project/prep-cert.sh && /run.sh pure-ftpd -4 -A -c 50 -C 5 -e -G -H -i -p $FTP_MIN_PORT:$FTP_MAX_PORT -P $PUBLICHOST -r -R -s -X -x -Z $FLAGS
+CMD /project/prep-cert.sh && /run.sh pure-ftpd -4 -A -c 50 -C 5 -e -G -H -i -p $FTP_MIN_PORT:$FTP_MAX_PORT $FTP_IP -P 129.16.69.158 -r -R -s -X -x -Z $FLAGS
```
e0, line 9: Is this supposed to be the following?

```suggestion
CMD /project/prep-cert.sh && /run.sh pure-ftpd -4 -A -c 50 -C 5 -e -G -H -i -p $FTP_MIN_PORT:$FTP_MAX_PORT -P $FTP_IP -r -R -s -X -x -Z $FLAGSWrite<br class="Apple-interchange-newline">
```

mihai-sysbio: Of course! I'm not sure how this slipped through, great that you spotted it!

e0 (approved):

Issue #597, opened 2021-05-26
Mismatch of reactions in model and SVG map


inghylt:
As a developer, I would like all reaction IDs in the SVG map to be found in the model, so that the IDs are coherent.

Verified issue is for the new TSV format Golgi apparatus there were 41 reaction IDs in the SVG map not found in the  data-files/integrated-models/Human-GEM/Human-GEM.yml

Verification has not been performed for old TSV format

nanjiangshu:
I can confirm the mismatch also for the old TSV format as in the branch `feat/more-models` for the repo `data-files`

mihai-sysbio:
As the models are being revised, they not only gain reactions but also lose reactions. Without updating the SVG maps, the reactions would still appear to be part of the model even though they're not.

mihai-sysbio:
I'm in favour of closing this issue, as the proper fix is to have a new way of obtaining the 2D maps.

PR #598, opened 2021-05-26
feat: 2D map coverage MA-784

### LINKED ISSUE #581: 2D map coverage
For the 2D maps, in the Map Viewer sidebar, show the coverage, i.e. how many of that compartment/subsystem reactions are on that map. Normally this would not be needed - the Human-GEM maps have very good coverage. This seems not be the case for the Yeast-GEM maps. Therefore, users need to be informed when a map has low coverage, and the best way to do that is to show the coverage of each map.The coverage should be computed by comparing the contents of the map vs the contents of the model - since the map are hand-drawn, there is a difference between those and the “actual” model content in the database.

It will also be good to inform the user which reactions are missing in the 2D map compared to the database. 

Note: the coverage should be limited to reactions and only for compartment and subsystem maps (not custom maps).

- [ ] use a selector to extract the reactions from the svg
- [ ] use (a new?) api endpoint to retrieve all the reaction ids for a compartment or a subsystem
- [ ] compare the two sets of the reactions
- [ ] show the results of the comparison in the frontend
- [ ] have a `Show more`/`Expand` button for detailed diff
- [ ] make sure the `Documentation` explains the cause of difference (the maps have been manually add using a previous instance of the model)
- [ ] use an `info` icon similar to the Data overlay to link to the relevant section in the Documentation

image

inghylt:
This PR closes  #581

This feature informs if there are missing reactions on the map and provides a detailed list of the missing and total reactions in a popup window.

- [x] use a selector to extract the reactions from the svg
- [x] use (a new?) api endpoint to retrieve all the reaction ids for a compartment or a subsystem
- [x] compare the two sets of the reactions
- [x] show the results of the comparison in the frontend
- [x] have a Show more/Expand button for detailed diff
- [x] make sure the Documentation explains the cause of difference (the maps have been manually add using a previous instance of the model) (Info in popup instead of Documentation)
- [x] use an info icon similar to the Data overlay to link to the relevant section in the Documentation (Info in popup instead of Documentation)

This PR was created by @nanjiangshu  and @inghylt

### DIFF
--- a/api/src/neo4j/queries/map.js
+++ b/api/src/neo4j/queries/map.js
@@ -17,6 +17,11 @@ CALL apoc.cypher.run("
   RETURN { id: $cid, reactionCount: COUNT(DISTINCT(r)) } as data
   
   UNION
+
+  MATCH (:Compartment${m} {id: $cid})-[${v}]-(:CompartmentalizedMetabolite)-[${v}]-(r:Reaction)
+  RETURN { id: $cid, reactionList: COLLECT(DISTINCT(r.id)) } as data
+  
+  UNION
   
   MATCH (csvg:SvgMap)-[${v}]-(:Compartment${m} {id: $cid})
   RETURN { id: $cid, svgs: COLLECT(DISTINCT(csvg {.*})) } as data
--- a/api/src/neo4j/queries/map.js
+++ b/api/src/neo4j/queries/map.js
@@ -41,6 +41,11 @@ CALL apoc.cypher.run("
   
   MATCH (:Subsystem${m} {id: $sid})-[${v}]-(r:Reaction)
   RETURN { id: $sid, reactionCount: COUNT(DISTINCT(r)) } as data
+
+  UNION
+  
+  MATCH (:Subsystem${m} {id: $sid})-[${v}]-(r:Reaction)
+  RETURN { id: $sid, reactionList: COLLECT(DISTINCT(r.id)) } as data
   
   UNION
   
--- a/frontend/src/components/explorer/mapViewer/SidebarDataPanels.vue
+++ b/frontend/src/components/explorer/mapViewer/SidebarDataPanels.vue
@@ -8,12 +8,59 @@
         </p>
       </header>
       <div class="card-content p-4">
-        <div class="content">
-          <p v-if="modelNumberOfReactions">
-          <span>Reaction coverage:</span> {{ mapNumberOfReactions }}
+        <div class="content" @click="showModal = true">
+          <modal v-show="showModal" @close="showModal = false"></modal>
+          <p v-if="currentMap.reactionList">
+          <span>Reaction coverage:</span> {{ mapReactionList.length }}
           out of {{ modelNumberOfReactions }} reactions are on the map </p>
         <p v-else class="has-text-centered"> Reaction coverage not available </p>
       </div>
+
+      <div>
+        <div v-if="showModal" class="modal is-active">
+          <div class="modal-background"></div>
+          <div class="modal-content">
+            <div class="box">
+              <table class="table main-table is-fullwidth m-0">
+                <tr>
+                  <td class="td-key has-background-primary has-text-white-bis" >Missing reactionIDs on the map</td>
+                  <td>
+                    <div v-html="missingReactionIdListHtml"></div>
+                    <div v-if="!showFullReactionList &&  modelNumberOfReactions > displayedReaction">
+                      <br>
+                      <button class="is-small button" @click="showFullReactionList=true">
+                        ... and {{ missingReactionList.length - displayedReaction }} more
+                      </button>
+                      <span v-show="missingReactionList.length >= limitReaction"
+                            class="tag is-medium is-warning is-pulled-right">
+                        The number of reactions displayed is limited to {{ limitReaction }}
+                      </span>
+                    </div>
+                  </td>
+                </tr>
+                <tr>
+                  <td class="td-key has-background-primary has-text-white-bis">All reactionIDs in the model</td>
+                    <div v-html="totalReactionIdListHtml"></div>
+                    <div v-if="!showFullReactionList &&  modelNumberOfReactions > displayedReaction">
+                      <br>
+                      <button class="is-small button" @click="showFullReactionList=true">
+                        ... and {{ modelNumberOfReactions - displayedReaction }} more
+                      </button>
+                      <span v-show="modelNumberOfReactions >= limitReaction"
+                            class="tag is-medium is-warning is-pulled-right">
+                        The number of reactions displayed is limited to {{ limitReaction }}
+                      </span>
+                    </div>
+                  <td></td>
+                </tr>
+              </table>
+            </div>
+          </div>
+          <button class="modal-close" @click="showModal = false"></button>
+        </div>
+      </div>
+
+
     </div>
       <footer v-if="currentMap.type !== 'custom'" class="card-footer">
         <router-link class="p-0 is-info is-outlined card-footer-item has-text-centered"
@@ -134,7 +181,7 @@
 
 <script>
 import { mapState } from 'vuex';
-import { capitalize, reformatStringToLink, reformatChemicalReactionHTML } from '@/helpers/utils';
+import { capitalize, reformatStringToLink, reformatChemicalReactionHTML, buildCustomLink } from '@/helpers/utils';
 import { chemicalFormula } from '@/helpers/chemical-formatters';
 import { default as messages } from '@/helpers/messages';
 
@@ -168,16 +215,60 @@ export default {
       showMapCardContent: true,
       showSelectionCardContent: true,
       messages,
+      showModal: false,
+      showFullReactionList: false,
+      displayedReaction: 40,
+      limitReaction: 999999999,
+      missingNumberOfReactions: null,
     };
   },
   computed: {
     ...mapState({
       model: state => state.models.model,
       loading: state => state.maps.loadingElement,
-      mapNumberOfReactions: state => state.maps.mapNumberOfReactions,
+      mapReactionList: state => state.maps.svgReactionsIdList,
     }),
     modelNumberOfReactions() {
       return this.currentMap.reactionList ? this.currentMap.reactionList.length : null;
+    },
+    totalReactionIdListHtml() {
+      const l = ['<span class="tags">'];
+      for (let i = 0; i < this.modelNumberOfReactions; i += 1) {
+        const r = this.currentMap.reactionList[i];
+        if ((!this.showFullReactionList && i === this.displayedReaction)
+          || i === this.limitReaction) {
+          break;
+        }
+        const customLink = buildCustomLink({ model: this.model.short_name, type: 'reaction', id: r, title: r, cssClass: 'target="_blank"' });
+        l.push(
+          `<span id="${r}" class="tag">${customLink}</span>`
+        );
+      }
+      l.push('</span>');
+      return l.join('');
+    },
+    missingReactionList() {
+      const modelReactionIdSet = new Set(this.currentMap.reactionList);
+      const mapReactionIdSet = new Set(this.mapReactionList);
+      const missingReactionIdSet = new Set([...modelReactionIdSet].filter(x => !mapReactionIdSet.has(x)));
+      const missingReactionList = Array.from(missingReactionIdSet);
+      return missingReactionList;
+    },
+    missingReactionIdListHtml() {
+      const l = ['<span class="tags">'];
+      for (let i = 0; i < this.missingReactionList.length; i += 1) {
+        const r = this.missingReactionList[i];
+        if ((!this.showFullReactionList && i === this.displayedReaction)
+          || i === this.limitReaction) {
+          break;
+        }
+        const customLink = buildCustomLink({ model: this.model.short_name, type: 'reaction', id: r, title: r, cssClass: 'target="_blank"' });
+        l.push(
+          `<span id="${r}" class="tag">${customLink}</span>`
+        );
+      }
+      l.push('</span>');
+      return l.join('');
     },
   },
   methods: {
--- a/frontend/src/components/explorer/mapViewer/SidebarDataPanels.vue
+++ b/frontend/src/components/explorer/mapViewer/SidebarDataPanels.vue
@@ -11,7 +11,7 @@
            class="card-content p-4 sidebarCardHover">
         <div class="content mb-0" @click="showModal = true">
           Please note that {{ missingReactionList.length }}
-          of the reactions in the model are not shown on the map
+          of the reactions in the {{ currentMap.type }} are not shown on the map
           <span class="icon"><i class="fa fa-info-circle"></i></span>
         </div>
         <div v-if="showModal" class="modal is-active">
--- a/frontend/src/components/explorer/mapViewer/SidebarDataPanels.vue
+++ b/frontend/src/components/explorer/mapViewer/SidebarDataPanels.vue
@@ -8,11 +8,11 @@
         </p>
       </header>
       <div v-if="currentMap.reactionList && missingReactionList.length > 0"
-           class="card-content p-4 sidebarCardHover">
-        <div class="content mb-0" @click="showModal = true">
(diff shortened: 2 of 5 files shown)


e0 (changes requested):
Really good job on this @inghylt and @nanjiangshu! This was not an easy issue to code from scratch and it's great to see how far you have come along.

I left some inline comments. For `SidebarDataPanels.vue`, I haven't looked through it line by line yet but I have some overall comments based on testing it.

Take a look at the attached image for example:

image

This is from `/explore/Human-GEM/map-viewer/acyl_coa_hydrolysis?dim=2d`.

- It seems that there are more reactions displayed than exists in the subsystem. Not sure if there is a good solution for this, but wanted to point it out.
- The "Missing reactions on the map" shows "40 more" but the list is empty. Clicking it doesn't show any either.
- It would be nice with some padding to wrap the list of ids. It looks like there is no top and left spacing for the tags in "All reactions in the model".
- It would be good if we can dismiss the modal by clicking anywhere outside the modal, as opposed to having to click the close button.
- Not sure if this was mentioned in the specs for this issue but I think it would be better to show "Reactions on the map" instead of "All reactions in the model". For example if a compartment has 200 total reactions and 100 missing on the map, it would make it easier to identify the reactions shown in the map if the user doesn't have to manually compare the two lists.
- Certain maps do not show the comparison, even though it looks like its list of reactions is included in API response. For example: `/explore/Human-GEM/map-viewer/acylglycerides_metabolism?dim=2d`
- Clicking the card content (see below) shows the modal. Perhaps we could make it more intuitive for users to discover this interaction. For example a little info icon (https://bulma.io/documentation/elements/icon/#colors).

image

nanjiangshu:
Hi Shan @e0 Thanks very much for the detailed and valuable comments

> * It seems that there are more reactions displayed than exists in the subsystem. Not sure if there is a good solution for this, but wanted to point it out.

There are reactions in the SVG map but not found in the model as retrieved from the neo4j database. We have raised this problem in the issue #597 

If this is normal, i.e. there could be reactions in the SVG map but not in the model and vice versa (could @mihai-sysbio please verify that as well?), then we should rephrase the description to 
```
XX reactions on the map are not found in the model
YY reactions in the model are not found on the map
``` 

> 
> * The "Missing reactions on the map" shows "40 more" but the list is empty. Clicking it doesn't show any either.
Good catch! In this case, the missing number of reactions on the map is zero. So the `show more` button should not be displayed.

> * It would be nice with some padding to wrap the list of ids. It looks like there is no top and left spacing for the tags in "All reactions in the model".

Good catch! We will add the padding. 

> * It would be good if we can dismiss the modal by clicking anywhere outside the modal, as opposed to having to click the close button.

I personally prefer to keep the feature that one can only close the popup window by clicking the `close` button but maybe you are right, to keep consistent with the `Repository` page, we should allow close function by clicking anywhere outside of the modal window.

> * Not sure if this was mentioned in the specs for this issue but I think it would be better to show "Reactions on the map" instead of "All reactions in the model". For example if a compartment has 200 total reactions and 100 missing on the map, it would make it easier to identify the reactions shown in the map if the user doesn't have to manually compare the two lists.

Sounds good. We will change it. 

> * Certain maps do not show the comparison, even though it looks like its list of reactions is included in API response. For example: `/explore/Human-GEM/map-viewer/acylglycerides_metabolism?dim=2d`

The comparison is not shown if there is no missing reaction. 

> * Clicking the card content (see below) shows the modal. Perhaps we could make it more intuitive for users to discover this interaction. For example a little info icon (https://bulma.io/documentation/elements/icon/#colors).

Good suggestion! We added a hover, but maybe it is not obvious enough.

mihai-sysbio:
Was this PR really meant to be closed?

e0:
> The comparison is not shown if there is no missing reaction.

Ah that makes sense. Do you think it would be nice to add a message like "All 123 reactions are included in the map."? On one hand I like the consistency in UI but maybe it's just unnecessary information. 

---

> Good suggestion! We added a hover, but maybe it is not obvious enough.

The hover is nice but I think it's good to give the user a nudge to let them know it's hover-able, as well as helping users using touch devices.

frontend/src/components/explorer/mapViewer/SidebarDataPanels.vue
```diff
@@ -7,9 +7,67 @@
           <i>{{ currentMap.name }}</i>
         </p>
       </header>
-      <footer v-if="currentMap.type !== 'custom'" class="card-footer">
+      <div v-if="currentMap.reactionList && mapReactionList.length !== modelNumberOfReactions"
+      class="card-content p-4 sidebarCardHover">
```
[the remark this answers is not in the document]
nanjiangshu: Will take a look at it

api/src/neo4j/queries/map.js
```diff
@@ -16,6 +16,11 @@ CALL apoc.cypher.run("
   MATCH (:Compartment${m} {id: $cid})-[${v}]-(:CompartmentalizedMetabolite)-[${v}]-(r:Reaction)
   RETURN { id: $cid, reactionCount: COUNT(DISTINCT(r)) } as data
```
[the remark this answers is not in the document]
nanjiangshu: Good point!

frontend/src/store/modules/maps.js
```diff
@@ -38,6 +38,7 @@ const data = {
   backgroundColor: BG_COLORS.light,
   loading: true,
   loadingElement: false,
+  svgReactionsIdList: null,
```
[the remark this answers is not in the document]
nanjiangshu: Sure. I agree it is better to initialize an array as [] instead of null, in case e.g. `array.length` is called.
However, It a bit weird that I did not see any Vue error with `/explore/Human-GEM/map-viewer/acylglycerides_metabolism?dim=2d`

frontend/src/components/explorer/mapViewer/Svgmap.vue
```diff
@@ -153,7 +153,18 @@ export default {
...
+      const svgWrapper = document.querySelector('#svg-wrapper');
+      const svgReactionsIdList = [];
+      Object.entries(svgWrapper.children).forEach((item) => {
+        const nodesChildren = item[1].children.nodes.children;
+        Object.entries(nodesChildren).forEach((child) => {
+          if (child[1].className.animVal === 'rea') {
+            svgReactionsIdList.push(child[1].id);
+          }
+        });
+      });
```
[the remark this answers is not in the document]
nanjiangshu: Good point!

nanjiangshu:
> Was this PR really meant to be closed?

No, it must be done by mistake. Thanks for re-opening it.

api/src/neo4j/queries/map.js
```diff
@@ -36,6 +41,11 @@ CALL apoc.cypher.run("
   
   MATCH (:Subsystem${m} {id: $sid})-[${v}]-(r:Reaction)
   RETURN { id: $sid, reactionCount: COUNT(DISTINCT(r)) } as data
```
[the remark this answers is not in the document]
nanjiangshu: We have added a new issue to solve this problem

frontend/src/components/explorer/mapViewer/SidebarDataPanels.vue
```diff
@@ -7,7 +7,65 @@
           <i>{{ currentMap.name }}</i>
         </p>
       </header>
-      <footer v-if="currentMap.type !== 'custom'" class="card-footer">
+      <div v-if="currentMap.reactionList && missingReactionList.length > 0"
+           class="card-content p-4 sidebarCardHover">
+        <div class="content mb-0" @click="showModal = true">
+          Please note that {{ missingReactionList.length }}
+          of the reactions in the model are not shown on the map
+          <span class="icon"><i class="fa fa-info-circle"></i></span>
```
mihai-sysbio, line 15: This icon is not showing up for me
image

frontend/src/components/explorer/mapViewer/SidebarDataPanels.vue
```diff
@@ -7,7 +7,65 @@
...
+        </div>
+        <div v-if="showModal" class="modal is-active">
+          <div class="modal-background" @click="closeModal"></div>
+          <div class="modal-content p-5 column is-6-fullhd is-8-desktop is-10-tablet is-full-mobile
+          has-background-white">
+            <h4 class="title is-size-4 m-0 mb-2"> List of missing and total reactions on the map </h4>
+            <p class="pb-4"> There are {{ missingReactionList.length }}
+              reactions not shown on the map as compared to the model since
+              the map is manually curated and some reactions are not possible
+              to add to the map. The number of reaction shown on the map is
```
mihai-sysbio, line 25: The part of the text that aims to provide an explanation should be moved to the Documentation page:

```suggestion
              reactions not shown on the map as compared to the model.
              The number of reaction shown on the map is
```

frontend/src/components/explorer/mapViewer/SidebarDataPanels.vue
```diff
@@ -160,13 +218,62 @@ export default {
       showMapCardContent: true,
       showSelectionCardContent: true,
       messages,
+      showModal: false,
+      showFullReactionListMissing: false,
+      showFullReactionListMap: false,
+      displayedReaction: 40,
+      limitReaction: 999999999,
```
mihai-sysbio, line 225: This limit is so high, might as well not have it at all.

frontend/src/components/explorer/mapViewer/SidebarDataPanels.vue
```diff
@@ -160,13 +218,62 @@ export default {
...
+          || i === this.limitReaction) {
+          break;
+        }
+        const customLink = buildCustomLink({ model: this.model.short_name, type: 'reaction', id: r, title: r, cssClass: 'target="_blank"' });
+        l.push(
+          `<span id="${r}" class="tag">${customLink}</span>`
+        );
+      }
+      l.push('</span>');
+      return l.join('');
```
mihai-sysbio, line 252: Could this be don in a Vue template/component instead?

frontend/src/components/explorer/mapViewer/SidebarDataPanels.vue
inghylt: 1. Should we link to the Documentation?
2. Currently, the documentation says _"While a very high percentage of the reactions in the model are represented on the 2D maps, some may be unavailable."_ Should we elaborate why some may be unavailable, e.g.
_"While a very high percentage of the reactions in the model are represented on the 2D maps, some may be unavailable. 
The maps are manually curated and some reactions are not possible to add."_

mihai-sysbio: Yes, and yes:
> While a very high percentage of the reactions in the model are represented on the 2D compartment and subsystem maps, some may be missing. This is partly due to the model being updated much more often than the maps. Also, as the maps are manually curated, in rare occasions some reactions are cannot be added.

e0: Didn't we decide to not link to the Documentation page? Below is a screenshot from this issue's description. I thought the text within the parenthesis meant that we should show it in the popup and not link to the Documentation page.

image

frontend/src/components/explorer/mapViewer/SidebarDataPanels.vue
e0: This pattern is also prevalent in other places such as [Subsystem.vue](https://github.com/MetabolicAtlas/MetabolicAtlas/blob/4e46aeea9605df5040c7bc02aa9c2d97bfebb878/frontend/src/components/explorer/gemBrowser/Subsystem.vue#L139). I remember there was a problem with performance when trying to render too many `router-link` elements but not sure if the problem was exactly here. Either way we should probably open up a new issue if we want to consider having these in template/component.

frontend/src/components/explorer/mapViewer/SidebarDataPanels.vue
nanjiangshu: Maybe we can set it to a smaller number, e.g. 1000? Not sure how the popup window looks like or if it will be slow to load if there are several thousand reactions.

frontend/src/components/explorer/mapViewer/SidebarDataPanels.vue
nanjiangshu: @inghylt and me considered about it but then we thought it might be awkward to have two links, one to the modal table and another to the documentation, almost at the same place.

mihai-sysbio: That must be a misunderstanding on my side - I imagined the `icon` to act as the link to the Documentation.
So, the user should be taken to the Documentation page, and to keep things the same as the `Data overlay` and to make it clear that it's not functionality related, the link should be achieved through an `icon`.

frontend/src/components/explorer/mapViewer/SidebarDataPanels.vue
mihai-sysbio: >open up a new issue 

@e0 or @nanjiangshu please do so, and feel free to link to the above comment of this PR. I'm marking this discussion as resolved.

frontend/src/components/explorer/mapViewer/SidebarDataPanels.vue
mihai-sysbio: One way to test this would be to associate the _Mitochondrion_ map to another compartment through the association file in `data-files`, leading to none of the thousand reactions being correctly associated.
I'm hoping that the loading is fast enough to not require any limit.

frontend/src/components/explorer/mapViewer/SidebarDataPanels.vue
inghylt: Weird, I tried on both Firefox and Chrome and it shows. Not sure why it wouldn't on yours. @e0 do you have any idea?

e0: It shows up for me too! Maybe try clearing your cache @mihai-sysbio

image

mihai-sysbio: It did show up after a rebuild. Thanks for checking!

frontend/src/components/explorer/mapViewer/SidebarDataPanels.vue
inghylt: @nanjiangshu me and Mihail experimented with this on localhost and did not run in to any performance/visual issues. Thinking that is better to remove it for now and add it again if we encounter a problem in the future.

api/src/neo4j/queries/map.js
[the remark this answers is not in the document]
inghylt: We created a separate issue for this as it might affect other parts of the code than what we are looking into right now, see  #603

mihai-sysbio (approved):
Great work!
It looks like the only remaining piece of feedback is related to the Documentation, but that doesn't stop me from already approving.
@inghylt it would be good if you could also link to this a new issue regarding the exporting of the table of reactions as CSV (similar to how we have it in other places on the website).

e0 (approved):

frontend/src/components/explorer/mapViewer/SidebarDataPanels.vue
```diff
@@ -7,7 +7,57 @@
...
+        </div>
+        <div v-if="showModal" class="modal is-active">
+          <div class="modal-background" @click="closeModal"></div>
+          <div class="modal-content p-5 column is-6-fullhd is-8-desktop is-10-tablet is-full-mobile
+          has-background-white">
+            <h4 class="title is-size-4 m-0 mb-2"> List of missing and total reactions on the map </h4>
+            <p class="pb-4">
+              There are {{ missingReactionList.length }} reactions not shown on the map. Some reactions
+              are missing as the {{ currentMap.type }} is being updated much more often than the maps.
+              Also, as the maps are manually curated, occasionally some reactions are cannot be added.
```
mihai-sysbio, line 25:
```suggestion
              Also, as the maps are manually curated, occasionally some reactions cannot be added.
```

PR #600, opened 2021-05-26
fix: improve cypher query for compartments so they don't get timed out

### LINKED ISSUE #599: Compartment not loading in gembrowser
[Compartment extracellular](https://metabolicatlas.org/explore/Human-GEM/gem-browser/compartment/extracellular) and [compartment cytosol](https://metabolicatlas.org/explore/Human-GEM/gem-browser/compartment/cytosol) are not working in the GEM Browser, a `not found` error is returned.

e0:
This closes #599 

For the compartment `extracellular` in Human-GEM, the response time (when testing locally) decreased from ~27s to ~0.5s. The `cytosol` compartment, which is one of the largest, takes just under 1s now.

### DIFF
--- a/api/src/neo4j/queries/compartment.js
+++ b/api/src/neo4j/queries/compartment.js
@@ -11,7 +11,7 @@ CALL apoc.cypher.run("
   
   UNION
     
-  MATCH (:Compartment${m} {id: '${id}'})-[${v}]-(cm:CompartmentalizedMetabolite)-[${v}]-(r:Reaction)-[${v}]-(s:Subsystem)-[${v}]-(ss:SubsystemState)
+  MATCH (:Compartment${m} {id: '${id}'})-[${v}]-(:CompartmentalizedMetabolite)-[${v}]-(r:Reaction)-[${v}]-(s:Subsystem)-[${v}]-(ss:SubsystemState)
   USING JOIN on r
   RETURN { subsystems: COLLECT(DISTINCT({id: s.id, name: ss.name})) } as data
   
--- a/api/src/neo4j/queries/compartment.js
+++ b/api/src/neo4j/queries/compartment.js
@@ -11,23 +11,19 @@ CALL apoc.cypher.run("
   
   UNION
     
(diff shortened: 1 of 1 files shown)


mihai-sysbio (approved):
Awesome stuff! Random idea: would it be possible to reuse `r` from the first `MATCH` in the second one?

e0:
> Awesome stuff! Random idea: would it be possible to reuse `r` from the first `MATCH` in the second one?

I think we had some problems trying this before but perhpaps it's worth a revisit now that we have learnt new tricks with Cypher. If it works, there is quite a few other places where we could do this. Please open a new issue for this.

inghylt (commented):
image

Really nice to cut the response time so drastically! However, I get different resulting data for Reactions and Metabolites between branch `develop` and this branch, see attached image. Also compared http://localhost/explore/Human-GEM/gem-browser/compartment/cytosol, difference between Reactions and Metabolites there as well. Not sure this is intended

e0:
> image
> 
> 
> 
> Really nice to cut the response time so drastically! However, I get different resulting data for Reactions and Metabolites between branch `develop` and this branch, see attached image. Also compared http://localhost/explore/Human-GEM/gem-browser/compartment/cytosol, difference between Reactions and Metabolites there as well. Not sure this is intended

Good catch @inghylt ! This is mostly likely due to me trying to consolidate some sub-queries. Will try to add a new commit that addresses this soon.

nanjiangshu (commented):
This is absolutely awesome @e0! We should definitely try to use the techniques used here to improve the query speed at other places. I believe it will be a great boost for the user experience if we can reduce the loading time for all pages to <0.5ms.

I can approve this PR when the problem mentioned by @inghylt is addressed.

e0:
I have added a commit that should return the correct counts now. Please have another look when you have time. Thanks again @inghylt ! I tried to consolidate the `r` used to fetch `Subsystem` as well and it seems to be fine with no impact on performance. I also tried to include `Gene` in the same sub-query with an `OPTIONAL MATCH`. While that returned the correct counts it made things slower, about 25% slower for `extracellular` and 100% slower for `cytosol`.

inghylt (approved):
Nice, good job fixing it so quickly! I tried it out on cytosol, extracellular, golgi_apparatus and nucleus and they all return the same values as on develop :+1:

PR #621, opened 2021-06-08
fix: code ql warnings

### LINKED ISSUE #606: Fix CodeQL warnings
CII Best Practice:
_All medium and higher severity exploitable vulnerabilities discovered with static code analysis MUST be fixed in a timely way after they are confirmed_

The list of the issues is available here https://github.com/MetabolicAtlas/MetabolicAtlas/security/code-scanning

e0:
This PR closes #606. Looks like this clears all the alert (see screenshot below).

image

For the "Prototype-polluting assignment" warnings (`CACHED_3D_NETWORKS[model][version].whole = n;`), I simply removed the custom cache for 3D maps. For servers these should already being cached by nginx now. The potential downside is that when working locally, one might need to wait longer for working with larger 3D maps. Or maybe it is good if the developer can always see the actual load time for a 3D map. If we want to make sure that 3D networks are cached when working locally I suggest we create a new issue for caching them using nginx.

### DIFF
--- a/api/src/endpoints/hpaRna.js
+++ b/api/src/endpoints/hpaRna.js
@@ -11,7 +11,7 @@ routes.get('/all', async (req, res) => {
 
     res.json(HPA_JSON);
   } catch (e) {
-    res.status(400).send(e);
+    res.status(400).send(e.message);
   }
 });
 
--- a/api/src/endpoints/neo4j.js
+++ b/api/src/endpoints/neo4j.js
@@ -25,7 +25,6 @@ import {
 } from 'neo4j/index';
 
 const neo4jRoutes = express.Router();
-const CACHED_3D_NETWORKS = {};
 
 const fetchWith = async (req, res, queryHandler) => {
   const { id } = req.params;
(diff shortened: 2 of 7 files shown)


mihai-sysbio (changes requested):
Very nice to have started with this @e0!

> I simply removed the custom cache for 3D maps. For servers these should already being cached by nginx now.

If my memory serves me well, the caching can take around a minute per map, and the intention with caching in the backend was that we could trigger the caching from code after startup, instead of trying to squeeze these API calls among the first, in order to be cached. Regardless, I support the decision to remove the caching.

However, I think there's a timeout that needs updating - at least several compartment maps give throw an error, eg http://localhost/explore/Human-GEM/map-viewer/nucleus?dim=3d

e0:
> However, I think there's a timeout that needs updating - at least several compartment maps give throw an error, eg http://localhost/explore/Human-GEM/map-viewer/nucleus?dim=3d 
> 

The map still needs to be loaded from scratch the first time it's requested even with the cache. For the server this change should not introduce any changes in performance at all. For local development, I believe you would get this timeout regardless of having the cache or not too. Do you want to open up a new issue for this?

inghylt (commented):
Nice work, great to get this fixed!

All 3d maps worked for me except localhost/explore/Human-GEM/map-viewer/extracellular?dim=3d

That map works when building from develop

mihai-sysbio:
Commit daf85f2 increases the 60s timeout to 300s for the requests Nginx passes on to the API. All 3D maps should load successfully now, which I confirmed by loading the extracellular compartment. @inghylt could you have another look?

mihai-sysbio (approved):

inghylt (approved):
Nice, everything seem to work. The changes made to proj.sh requires a certain version of Docker, could be wise to update README.md regarding this

mihai-sysbio:
@e0 would you prefer an updated Readme.md or a reversal of the changes in `proj.sh`?

e0:
Thanks for testing this and helping out. Still not sure why the map loading time would be different but adding the timeout does not hurt.

I think we can get by without the changes in proj.sh so I'm in favor of reverting it. We should probably have a new issue for consolidating on Docker version and updating this file if it makes sense (need to check if the new syntax works everywhere in proj.sh).

Issue #617, opened 2021-06-08
Use template styling components in gem browser


inghylt:
As a developer, I would like to have a template for all the components in the gem browser, so that I can avoid repetitive code in the stack and expect similar behavior for the different components

mihai-sysbio:
When using the [Vue cli](https://cli.vuejs.org/) there is a different folder called `views`.

PR #622, opened 2021-06-10
feat: new layout for maps available card MA-755

### LINKED ISSUE #584: new layout for maps-available card
As I user, it should be straightforward and simple to see which maps are available (2D, 3D).

Proposed layout:

Map-name 2D, 3D

- [x] remove the functionality related to `getAvailableMaps` in frontend/src/store/modules/maps.js as it is no longer used
- [x] the `...` to expand is weird because it expands both lists regardless of which is clicked
- [x] if the list is long then the scrolling happens inside the card, perhaps we can reduce the times this is needed or allow the cards to grow in length by narrowing the reactions table
- [x] sort maps titles (including `part x` for Cytosol maps from Human-GEM)
- [ ] from Pierre: the font size is bigger than the rest of the text in the GEM Browser
- [ ] make it clear what's a button and what not (the IP button above is blue, but the Export to TSV button is green)

inghylt:
This PR closes #584

A new layout for maps-available have been implemented, and each component has been adapted to this. Please visit http://localhost/explore/Human-GEM/gem-browser and look at each type of component to verify that everything is working well. 

Examples with and without Cytosol, i.e. a compartment with multiple maps: 
Without cytosol
http://localhost/explore/Human-GEM/gem-browser/reaction/HMR_8288

With cytosol
http://localhost/explore/Human-GEM/gem-browser/reaction/13DMTtu


mihai-sysbio (changes requested):
- [x] Very nice to use solution @inghylt and @e0 !
- [ ] Brilliant approach to handle the multiple maps associated to the cytosol 🌟 
- [ ] Conceptually, I don't see the need/benefit of requesting the available maps from each component vs having a dedicated endpoint, giving me the impression of duplicated code (across all the components).
- [ ] I'm not sure how you feel about this, but personally instead of reverting the commit I would have opted for removing it from history and force-pushing (after `git cherry-pick`ing it on another branch).
- [ ] There is a warning in the JS console:
```
[Vue warn]: You may have an infinite update loop in a component render function.
found in
---> <MapsAvailable> at src/components/explorer/gemBrowser/MapsAvailable.vue
       <Reaction> at src/components/explorer/gemBrowser/Reaction.vue
         <App> at src/App.vue
           <Root>
```
- [ ] I would suggest left-aligning the list of maps
- [ ] I'm wondering if it would look nicer have a colored rectangle behind the 2D/3D text so they would feel more like borderless buttons

I hope the checkboxes above will make it easier to keep track of the feedback, feel free to check them how you see fit @inghylt. And again, impressive stuff!

api/src/neo4j/queries/reaction.js
```diff
@@ -47,13 +47,15 @@ CALL apoc.cypher.run("
   
   MATCH (r:Reaction${m} {id: '${id}'})-[${v}]-(cm:CompartmentalizedMetabolite)
   WITH DISTINCT cm
-  MATCH (cm)-[${v}]-(:Compartment)-[${v}]-(csvg:SvgMap)
-  RETURN { compartmentSVGs: COLLECT(DISTINCT(csvg {.*})) } as data
+  MATCH (cm)-[${v}]-(c:Compartment)-[${v}]-(csvg:SvgMap)
+  WITH {compartmentId: c.id, compartmentSVGs: COLLECT(DISTINCT(csvg {.*}))} as compartmentSVG
+  RETURN { compartmentSVGs: COLLECT(compartmentSVG) } as data
```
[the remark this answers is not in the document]
inghylt: line 17 in `/frontend/src/store/modules/reactions.js`. Basically a fix to be able to group multiple maps belonging to same compartment, such as cytosol

api/src/neo4j/queries/reaction.js
```diff
@@ -47,13 +47,15 @@ CALL apoc.cypher.run("
...
   
   UNION
   
-  MATCH (r:Reaction${m} {id: '${id}'})-[${v}]-(:Subsystem)-[${v}]-(ssvg:SvgMap)
-  RETURN { subsystemSVGs: COLLECT(DISTINCT(ssvg {.*})) } as data
+  MATCH (r:Reaction${m} {id: '${id}'})-[${v}]-(s:Subsystem)-[${v}]-(ssvg:SvgMap)
+  WITH {subsystemId: s.id, subsystemSVGs: COLLECT(DISTINCT(ssvg {.*}))} as subsystemSVG
+  RETURN { subsystemSVGs: COLLECT(subsystemSVG) } as data
```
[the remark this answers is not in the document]
inghylt: line 26 in `/frontend/src/store/modules/reactions.js`

api/src/neo4j/queries/reaction.js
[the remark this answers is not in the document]
mihai-sysbio: I think the `compartmentId: c.id` can be skipped to end up with
```suggestion
  WITH {compartmentSVGs: COLLECT(DISTINCT(csvg {.*}))} as compartmentSVG
  RETURN { compartmentSVGs: COLLECT(compartmentSVG) } as data
```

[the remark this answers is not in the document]
e0: @mihai-sysbio that is almost what we had before. I think the change @inghylt made is great here because there is not enough info in the `compartmentSVG` nodes to reliably associate an SVG map with a compartment or subsystem. While it is currently technically possible to parse the compartment name out of something like `cytosol_1`, I think @inghylt's implementation is much more reliable since it's directly based on the graph database structure.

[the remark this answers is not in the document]
mihai-sysbio: Now I see, thanks for the demo @inghylt . `compartmentSVG ` is not the same as `compartmentSVGs`, my bad!

api/src/neo4j/queries/reaction.js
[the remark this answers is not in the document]
mihai-sysbio: Same misreading as above, sorry.

mihai-sysbio:
Oh btw this PR is to good to keep in `draft` mode :)

e0:
> * [ ]  Conceptually, I don't see the need/benefit of requesting the available maps from each component vs having a dedicated endpoint, giving me the impression of duplicated code (across all the components).


Great job with the feedback here @mihai-sysbio . Just wanted to comment about this point. Do you mean that instead of including the maps when requesting each component, we should add a new dedicated endpoint (something like `/api/available-maps/:type/:id`)? I think when originally implementing the API, we favored the current implementation because it would be more efficient. If you do want to look more into this, I suggest opening up a new issue as the current one is very frontend focused.

nanjiangshu (commented):
Great work! I think the new layout for the reaction page looks already great now. Keep on for other components. 

When looking at the inspector, I found one error in the Console
```
[Vue warn]: Error in render: "TypeError: _vm.searchModel is null"

found in

---> <GemSearch> at src/components/explorer/gemBrowser/GemSearch.vue
       <App> at src/App.vue
         <Root>
```
Would it be possible to solve this problem by assigning an initial value for `searchModal`
 or use `vif` to to check if it is `null`?

api/src/neo4j/shared/formatter.js
```diff
@@ -1,7 +1,25 @@
...
+
+export const reformatSubsystemSVGs = (reaction) => [...reaction.subsystems].sort((a, b) => a.id.localeCompare(b.id)).map((s) => {
+  const subsystemsWithSVGs = reaction.subsystemSVGs.find(({ subsystemId }) => subsystemId === s.id);
+  const svgMaps = subsystemsWithSVGs ? subsystemsWithSVGs.subsystemSVGs.sort((a, b) => a.id.localeCompare(b.id)) : [];
+  return {
+    id: s.id,
+    customName: s.name,
+    svgMaps,
+  };
+});
```
e0, line 25: Nice with having this in the backend @inghylt. Do you want to try further refactoring these two functions? Maybe they can both call a new function called `reformatSVGs(component, svgs)`.

frontend/src/components/explorer/gemBrowser/MapsAvailable.vue
```diff
@@ -6,34 +6,39 @@
...
-                <li :key="map.id">
-                  <router-link v-if="viewerSelectedID"
-                               :to="{ name: 'viewer',
-                                      params: { model: model.short_name, type: mapType, map_id: map.id, reload: true },
-                                      query: { dim: mapKey, search: viewerSelectedID, sel: viewerSelectedID } }">
+    <div v-if="mapsAvailable.length !== 0" class="card-content p-2">
+      <table class="table test-table">
+        <tbody class="has-text-left">
+          <tr v-for="component in mapsAvailable" :key="component.id">
+            <td> {{ component.customName }} </td>
```
e0, line 13: Since the maps available component is supposed to go in a column that `is-narrow`, which would have a flexible width, it might be good to limit the width of the component name. Not sure if there is a built-in Bulma helper for max-width but you can always access them with a CSS selector, for example: `.test-table tr td:first-child {`.

frontend/src/components/explorer/gemBrowser/MapsAvailable.vue
```diff
@@ -65,32 +70,19 @@ export default {
...
+      console.log('Zup');
+      if (this.viewerSelectedID) {
+        this.$router.push({ name: 'viewer',
+          params: { model: this.model.short_name, map_id: svgId, reload: true },
+          query: { dim: dimension, search: this.viewerSelectedID, sel: this.viewerSelectedID } });
+      } else {
+        this.$router.push({ name: 'viewer',
+          params: { model: this.model.short_name, map_id: svgId },
+          query: { dim: dimension } });
       }
```
e0, line 85: It's great that we moved this logic from markup to JavaScript.

1. Is `reload: true` necessary? 
2. It would be nice if this is further refactored to reduce repetition, for example:

```javascript
const params = { model: .. }
const query = { model: .. }

if (this. this.viewerSelectedID) {
  params.reload = true;
  query.search = ..
  ..
}
this.$router.push({ name: 'viewer', params, query });
```

e0 (commented):
Some minor changes are suggested but otherwise looks really good!

frontend/src/components/explorer/gemBrowser/MapsAvailable.vue
e0: It looks like the `reload` stuff is obsolete since [this commit](https://github.com/MetabolicAtlas/MetabolicAtlas/commit/e39ab79dd98dd61eb1fdb02349eb002e85e2eb8e#diff-eec96ec11596d102e2cf2efc9389cbe113c1bb28c5eb7db5249808a83abcab03). It also looks like it's not directly related to this issue. I can open a new issue for removing `reload` related stuff in the whole project, but I think it's safe for you to omit it from this file for now if you want to @inghylt .

frontend/src/components/explorer/gemBrowser/Metabolite.vue
```diff
@@ -75,7 +75,7 @@
                   {{ metabolite.name }} via ChEBI</a>
               </a>
             </div>
-            <div class="column is-3-widescreen is-3-desktop is-full-tablet has-text-centered">
+            <div class="column is-narrow">
```
e0, line 78: Since the new maps available card does not adapt to full-width for tablets, the spacing looks a bit weird for metabolites that have a chebi image. Not sure if this should be solved in this PR or a new issue.

image

frontend/src/store/modules/compartments.js
```diff
@@ -15,13 +15,9 @@ const actions = {
...
-    commit('maps/setAvailableMaps', {
-      '2d': {
-        compartment: compartmentSummary.compartmentSVGs,
-        subsystem: [],
-      },
-      '3d': { compartment: [{ id: compartmentSummary.info.id, customName: compartmentSummary.info.name }], subsystem: [] },
-    }, { root: true });
+    commit('maps/setAvailableMaps', [
+      ...compartmentSummary.compartmentSVGs,
+    ], { root: true });
```
e0, line 20: Would this give the same result?

```suggestion
    commit('maps/setAvailableMaps', compartmentSummary.compartmentSVGs, { root: true });
```

frontend/src/store/modules/subsystems.js
```diff
@@ -18,13 +18,9 @@ const actions = {
...
-    commit('maps/setAvailableMaps', {
-      '2d': {
-        compartment: [],
-        subsystem: subsystemSummary.subsystemSVGs,
-      },
-      '3d': { compartment: [], subsystem: [{ id: subsystemSummary.id, customName: subsystemSummary.name }] },
-    }, { root: true });
+    commit('maps/setAvailableMaps', [
+      ...subsystemSummary.subsystemSVGs,
+    ], { root: true });
```
e0, line 23: Would this give the same result?

```suggestion
    commit('maps/setAvailableMaps', subsystemSummary.subsystemSVGs, { root: true });
```

e0 (approved):
Great work @inghylt! Left some comments but I think they are negligible details so I am approving this PR.

frontend/src/components/explorer/gemBrowser/Metabolite.vue
mihai-sysbio: I'd say we should try to avoid `is-narrow` by controlling the width with the number of columns. There was some time spent on deciding these for the different widths, so please have an extra look that all the pages look good when going through the different widths.

frontend/src/components/explorer/gemBrowser/MapsAvailable.vue
```diff
@@ -33,12 +35,14 @@
               </div>
             </td>
             <td>
-              <router-link :to="{ name: 'viewer',
-                                  params: { model: model.short_name, map_id: component.id,
-                                            reload: true },
-                                  query: { dim: '3d', search: viewerSelectedID, sel: viewerSelectedID } }">
-                3D
-                </router-link>
+              <div class="button is-outlined is-small link-button">
```
mihai-sysbio: Since I suggested this, I realise it would have helped to provide a clear example - I imagined the `white` button from [here](https://bulma.io/documentation/elements/button/#colors).

mihai-sysbio:
Just to be clear, I don't have any further feedback for this - I'll approve as soon as the existing comments are resolved.

frontend/src/components/explorer/gemBrowser/MapsAvailable.vue
inghylt: image

You would prefer something like this instead of the current more buttonlike layout?

frontend/src/components/explorer/gemBrowser/Metabolite.vue
e0: Not sure what you mean by "at source" but the reason the card contents is not filling up the card in the screenshot you provided is because of the table layout used inside the maps available component. It has nothing to do with having `is-narrow` vs `is-full-tablet` on the wrapper column.

frontend/src/components/explorer/gemBrowser/MapsAvailable.vue
mihai-sysbio: Yeah, I personally liked that it was less crowded, but still felt the need for a button-like behaviour when hovering. What do you think?

frontend/src/components/explorer/gemBrowser/Metabolite.vue
e0: To add to this, what `is-narrow` does now is that it makes sure the maps available component is at 100% width and adapts the remaining column/s accordingly to fill up the remaining space.

mihai-sysbio: By "at source" I mean deciding on the card width in the source file where the component is defined, ie https://github.com/MetabolicAtlas/MetabolicAtlas/blob/2e7164739737e0e9281f57c4f16d58064f2f3c8c/frontend/src/components/explorer/gemBrowser/MapsAvailable.vue#L96-L98

Instead, it would be more reusable to decide the width in the parent component that uses it.

> makes sure the maps available component is at 100% width 

@e0 this is exactly the "inconsistent layout" that I think should be avoided, as some of the subsystem maps have really long names.

inghylt: It's great that you caught this! I think it might be better to focus on responsiveness with PR  #632 since all components will share the same file for layout. Since this is currently looking good for desktop, and that we will most likely close PR #632 within this sprint, would it be ok if I resolve this conversation and work with mobile and tablet responsiveness in that PR instead?

e0: > By "at source" I mean deciding on the card width in the source file where the component is defined, ie
> 
> https://github.com/MetabolicAtlas/MetabolicAtlas/blob/2e7164739737e0e9281f57c4f16d58064f2f3c8c/frontend/src/components/explorer/gemBrowser/MapsAvailable.vue#L96-L98
> 
> Instead, it would be more reusable to decide the width in the parent component that uses it.
> 
> > makes sure the maps available component is at 100% width
> 
> @e0 this is exactly the "inconsistent layout" that I think should be avoided, as some of the subsystem maps have really long names.

Ah ok, I think this is a good point about not having a `max-width` and it's worth rethinking how to handle that. The problem with using something like `is-3-desktop` vs `is-narrow` is that it sets a strict width for whatever is inside the column so it's difficult to work out how to layout the contents and would not play well with what we have now. I just wanted to clarify on for the cases where the component names are short, the table would still not fill up 100% of the column and that it is not due to `is-narrow`.

frontend/src/components/explorer/gemBrowser/MapsAvailable.vue
e0: I personally like the buttons and I think it's a good balance of being clear and having a good "weight". They also look nice and consistent alongside the select if there is one (cytosol). The white button would not directly help users with touch devices.

mihai-sysbio: Good points @e0.

@inghylt it's your call!

frontend/src/components/explorer/gemBrowser/Metabolite.vue
mihai-sysbio: @inghylt sure! In that case, please update #632 and/or #617 to mention that they should address the responsiveness as well.

nanjiangshu (approved):
Great job @inghylt! I like the new layout, clean and straightforward!

The warning ```
[Vue warn]: Error in render: "TypeError: _vm.searchModel is null"

found in

---> <GemSearch> at src/components/explorer/gemBrowser/GemSearch.vue
       <App> at src/App.vue
         <Root>
``` I mentioned in a previous comment is not introduced in this PR and not critical.
I can approve this PR now.

frontend/src/components/explorer/gemBrowser/MapsAvailable.vue
inghylt: To have the buttons look good together with select for components with cytosol, it looks better to keep things as they are

mihai-sysbio (approved):

PR #625, opened 2021-06-14
fix: make sure that gene ids that are within parenthesis are parsed properly

### LINKED ISSUE #590: make sure the gene page always shows related reactions
As a user, I would like to be presented with a list of reactions when visiting the gene page, so that I can explore reactions that are associated with this gene.

Bug: For some genes, for example https://metabolicatlas.org/explore/Human-GEM/gem-browser/gene/ENSG00000000419, it shows 0 related reactions. This should not be possible since a gene should always be associated with at least one reaction to be considered as a part of a model. (Thanks to @pecholleyc for pointing this out.)

### LINKED ISSUE #623: some reactions do not render
As a user, I want to click on search items (either quicksearch or global search), so that I see the details of the component I'm interested in.

After clicking on some reactions in the quicksearch, they do not render and trigger errors instead, eg HMR_9577 and HMR_9579. The API response seems fine. This issue was reported by @JonathanRob.

https://metabolicatlas.org/explore/Human-GEM/gem-browser/reaction/HMR_9491

e0:
The PR https://github.com/MetabolicAtlas/neo4j-data-generation/pull/18 is required for this to work.

This closes #623 and closes #590.

If successful, when browsing to `/explore/Human-GEM/gem-browser/reaction/HMR_9491` for example, you should see the following section:

image

### DIFF
--- a/frontend/src/api/reactions.js
+++ b/frontend/src/api/reactions.js
@@ -1,40 +1,6 @@
 import axios from 'axios';
 import { constructCompartmentStr, reformatChemicalReactionHTML } from '@/helpers/utils';
 
-const formatGeneRule = (reaction) => {
-  const { geneRule, genes } = reaction;
-
-  if (!geneRule) {
-    return null;
-  }
-  return geneRule.split(/\s+/).map((w) => {
-    if (w.match(/and|or/)) {
-      return w;
-    }
-
-    let gid = w;
-    let prefix = '';
-    let suffix = '';
-
-    if (gid.slice(0, 2) === '((') {
-      gid = gid.slice(2);
-      prefix = '((';
-    } else if (gid.slice(0, 1) === '(') {
-      gid = gid.slice(1);
-      prefix = '(';
-    } else if (gid.slice(-2) === '))') {
-      gid = gid.slice(0, -2);
-      suffix = '))';
-    } else if (gid.slice(-1) === ')') {
-      gid = gid.slice(0, -1);
-      suffix = ')';
-    }
-
-    const gene = genes.find(g => g.id === gid);
-    return `${prefix}${gene.name}${suffix}`;
-  }).join(' ');
-};
-
 const fetchReactionData = async ({ id, model, version }) => {
   const params = { model, version };
   let { data } = await axios.get(`/reactions/${id}`, { params });
@@ -44,7 +10,6 @@ const fetchReactionData = async ({ id, model, version }) => {
     compartment_str: data.compartments.map(c => c.name).join(', '),
     reactants: data.metabolites.filter(m => m.outgoing),
     products: data.metabolites.filter(m => !m.outgoing),
-    geneRule: formatGeneRule(data),
   };
 
   return {
--- a/frontend/src/api/reactions.js
+++ b/frontend/src/api/reactions.js
@@ -13,21 +13,25 @@ const formatGeneRule = (reaction) => {
     }
 
     let gid = w;
+    let prefix = '';
+    let suffix = '';
 
-    if (gid[0] === '(') {
+    if (gid.slice(0, 2) === '((') {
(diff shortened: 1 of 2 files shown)


inghylt (approved):
Nice work!

frontend/src/api/reactions.js
```diff
@@ -11,7 +11,22 @@ const formatGeneRule = (reaction) => {
     if (w.match(/and|or/)) {
       return w;
     }
-    const gene = genes.find(g => g.id === w);
+
```
nanjiangshu, line 14: As I commented in https://github.com/MetabolicAtlas/neo4j-data-generation/pull/18, due to the exists of multiple layers of parentheses, the parser needs to be able to handle such cases as well.

nanjiangshu (commented):
Nice work! However, the parser should be able to handle cases such as `/explore/Human-GEM/gem-browser/reaction/ATPasel` as well.

frontend/src/api/reactions.js
e0: [A fix](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/625/commits/f2f3933398a74e0623605526b9c72cc1a8c597b2) has been added to address this now. After you get this update as well as [the neo4j-data-generation one](https://github.com/MetabolicAtlas/neo4j-data-generation/pull/18/commits/c23570dadacd9341804a47630a8b9a15109ce859), you should be able to rebuild and test it on reactions with nested parenthesis. For example: `/explore/Human-GEM/gem-browser/reaction/ATPasel` should display the following.

image

e0: Btw, I do think there is a good refactoring opportunity to handle the gene rules but since it's a relatively serious bug I wanted to get this fix out as soon as possible.

frontend/src/components/explorer/gemBrowser/Reaction.vue
```diff
@@ -188,8 +188,15 @@ export default {
...
-            const prefix = e[0] === '(' ? '(' : '';
-            const suffix = e.slice(-1) === ')' ? ')' : '';
+            let prefix = '';
+            let suffix = '';
+            if (e[0] === '(') {
+              prefix = e[1] === '(' ? '((' : '(';
+            }
+            if (e.slice(-1) === ')') {
+              suffix = e.slice(-2) === '))' ? '))' : ')';
+            }
```
mihai-sysbio, line 198: I've `grep`-ed the `data-files/integrated-models` and it looks like there are cases of `(((` as well. How about using a regex instead,  like the one used below ``` ^\(+|\)+$/g ``` ?

mihai-sysbio (changes requested):

frontend/src/components/explorer/gemBrowser/Reaction.vue
e0: Ah, nice that you caught this and sounds like a good suggestion, thanks!

e0: I realize that there was a lot of redundant code, which has been removed now in [this commit](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/625/commits/b3af2639dcbd211b0bded9d467fe089fa1a864f6). It looks like the gene rule is only being used in `Reaction.vue`, so instead of reformatting the gene rule first in `frontend/src/api/reactions.js` and then in `frontend/src/components/explorer/gemBrowser/Reaction.vue`, it is only being formatted in the latter now. The formatting logic has also been rewritten to handle any level of nesting. There is also some legacy code involving `geneRule_wname` that has been removed since it's no longer used.

You should be able to test the reaction `/explore/Fruitfly-GEM/gem-browser/reaction/ATPasel` and see the following:

image

inghylt (approved):
Nice work and refactoring!

PR #627, opened 2021-06-15
fix: use id as fallback if name is not available for title in interaction partners page

### LINKED ISSUE #586: Null in title for genes in Interaction Partners
When looking at a gene in the Interaction Partners, the title shows "Interaction Partners for null" if you are using model mouse, fruit fly, zebrafish or rat. Seems to be related to neo4j data as the data returned from MetabolicAtlas/frontend/src/api/interactionPartners.js has null value for component.name when using the models mentioned above.

e0:
This closes #586.

### DIFF
--- a/frontend/src/store/modules/interactionPartners.js
+++ b/frontend/src/store/modules/interactionPartners.js
@@ -15,7 +15,10 @@ const data = {
 const getters = {
   component: state => state.interactionPartners.component || {},
   reactions: state => state.interactionPartners.reactions || [],
-  title: (state, _getters) => (_getters.component.type === 'metabolite' ? chemicalName(_getters.component.name) : _getters.component.name),
+  title: (state, _getters) => (_getters.component.type === 'metabolite'
+    ? chemicalName(_getters.component.name || _getters.component.id)
+    : _getters.component.name || _getters.component.id
+  ),
   reactionsSet: (state, _getters) => new Set(_getters.reactions.map(r => r.id)),
   componentName: (state, _getters) => _getters.component.name || _getters.component.id,
 };
--- a/frontend/src/store/modules/interactionPartners.js
+++ b/frontend/src/store/modules/interactionPartners.js
@@ -16,7 +16,7 @@ const getters = {
   component: state => state.interactionPartners.component || {},
   reactions: state => state.interactionPartners.reactions || [],
   title: (state, _getters) => (_getters.component.type === 'metabolite'
-    ? chemicalName(_getters.component.name || _getters.component.id)
+    ? chemicalName(_getters.component.name) || _getters.component.id
     : _getters.component.name || _getters.component.id
   ),
   reactionsSet: (state, _getters) => new Set(_getters.reactions.map(r => r.id)),


frontend/src/store/modules/interactionPartners.js
```diff
@@ -15,7 +15,10 @@ const data = {
 const getters = {
   component: state => state.interactionPartners.component || {},
   reactions: state => state.interactionPartners.reactions || [],
-  title: (state, _getters) => (_getters.component.type === 'metabolite' ? chemicalName(_getters.component.name) : _getters.component.name),
+  title: (state, _getters) => (_getters.component.type === 'metabolite'
+    ? chemicalName(_getters.component.name || _getters.component.id)
```
mihai-sysbio, line 19: IDs should not be interpreted as `chemicalName`.

```suggestion
    ? chemicalName(_getters.component.name)
```

mihai-sysbio (changes requested):
It's surprising to see that the function `chemicalName` is used in the title of the Interaction Partners, but not in the title of the GEM Browser.

frontend/src/store/modules/interactionPartners.js
e0: Nice! I was too sloppy with that. The metabolites didn't seem to have a problem anyways but do you think we should still try to account for the name possible being null? In that case it would be: 
```javascript
_getters.component.name ? chemicalName(_getters.component.name) : _getters.component.id
```

mihai-sysbio: Would this be applied to `'gene'` as well then?
`chemicalName` checks for null btw.

e0: > Would this be applied to `'gene'` as well then?

Can you elaborate on this? Not sure what you mean.

___

> `chemicalName` checks for null btw.

While it checks for null it doesn't return anything usable, so it would show "Interaction Partners for" if the name is null.

nanjiangshu: > IDs should not be interpreted as `chemicalName`.

I agree.

nanjiangshu, line 19: The interpretation of some of the chemical names are wrong, e.g. for 
`m02043c,h2s2o3  /explore/Human-GEM/interaction-partners/m02043c`. I think it should be H<sub>2</sub>S<sub>2</sub>O<sub>3</sub>.

e0: Good that you noticed this. Could you please open up a new issue for this? It's not directly related the purpose of this issue and PR.

nanjiangshu: There are more problems with the function `chemicalName`.  For example, `/explore/Human-GEM/interaction-partners/m02046c`, it should be HCO<sub>3</sub><sup>-</sup>, not HCO<sub>3</sub><sub>-</sub>. I think we should create a separate issue for that.

nanjiangshu:
> It's surprising to see that the function `chemicalName` is used in the title of the Interaction Partners, but not in the title of the GEM Browser.

chemicalName is not used in the Reaction Page neither, e.g. for `/explore/Human-GEM/gem-browser/reaction/HMR_2556`

frontend/src/store/modules/interactionPartners.js
e0: Thank you for creating the new issue!

mihai-sysbio: > > Would this be applied to `'gene'` as well then?
> 
> Can you elaborate on this? Not sure what you mean.

The Interaction Partners tool works for both genes and metabolites, and it is only the metabolites that should be passed to the `chemicalName` function. At a quick glance, it looked like the genes would also go through that parsing - it might not lead to any changes, but it should still be avoided.

mihai-sysbio: Nevermind the comment above, I see know it shouldn't be a problem!

frontend/src/store/modules/interactionPartners.js
```diff
@@ -15,7 +15,10 @@ const data = {
...
+    : _getters.component.name || _getters.component.id
+  ),
```
mihai-sysbio, line 21:
```suggestion
  title: (state, _getters) => (_getters.component.type === 'metabolite'
    ? chemicalName(_getters.component.name) || _getters.component.id
    : _getters.component.name || _getters.component.id
  ),
```

inghylt (commented):
Nice catch. Have tested out Mihails suggestion and it seems to work fine. So I'm up for approving this once that fix is committed

frontend/src/store/modules/interactionPartners.js
e0: Looks great, thanks!

mihai-sysbio (approved):

inghylt (approved):

PR #628, opened 2021-06-15
feat: make the toggle and data panels sticky in the map viewer's sidebar

### LINKED ISSUE #613: Review Map Viewer sidebar scrolling
As a user, I would (probably) like to have the 2D, 3D and Gem browser fixed, so that I don't have to scroll all the way to the top of the sidebar if I have chosen a subsystems map with a name at the end of the alphabet

e0:
This closes #613.

Please try this out and let me know if you have any thoughts on how to further improve the UI/UX.

### DIFF
--- a/frontend/src/components/explorer/MapViewer.vue
+++ b/frontend/src/components/explorer/MapViewer.vue
@@ -7,10 +7,11 @@
         </div>
       </template>
       <template v-else>
-        <div id="mapSidebar"
+        <div id="mapSidebar" ref="mapSidebar"
              class="column is-one-fifth-widescreen is-one-quarter-desktop
                     is-one-quarter-tablet has-background-lightgray om-2 pt-0
-                    fixed-height-desktop scrollable">
+                    fixed-height-desktop scrollable"
+             @scroll.passive="handleSidebarScroll">
           <div id="mapSidebar__header" class="has-background-lightgray pt-3">
             <div class="buttons has-addons is-centered padding-mobile m-0"
                  :title="`Switch to ${dimensionalState(!showing2D) }`"
@@ -26,9 +27,11 @@
                 <span class="is-uppercase">{{ dimensionalState(dim) }}</span>
               </button>
             </div>
-            <SidebarDataPanels :dim="dimensionalState(showing2D)"
+            <SidebarDataPanels ref="sidebarDataPanels"
+                               :dim="dimensionalState(showing2D)"
                                :current-map="currentMap"
-                               :selection-data="selectionData" />
+                               :selection-data="selectionData"
+                               @openSelectionCardContent="resetSidebarLayout" />
           </div>
           <div class="padding-mobile">
             <a class="button is-fullwidth is-primary is-inverted has-text-weight-bold is-hidden-tablet"
@@ -145,6 +148,7 @@ export default {
   },
   async created() {
     this.handleQueryParamsWatch = debounce(this.handleQueryParamsWatch, 100);
+    this.handleSidebarScroll = debounce(this.handleSidebarScroll, 300);
     window.onpopstate = this.handleQueryParamsWatch();
 
     EventBus.$off('loadRNAComplete');
@@ -166,6 +170,14 @@ export default {
     this.loadMapFromParams();
   },
   methods: {
+    handleSidebarScroll() {
+      if (this.$refs.mapSidebar.scrollTop > 0) {
+        this.$refs.sidebarDataPanels.hideSelectionCardContent();
+      }
+    },
+    resetSidebarLayout() {
+      this.$refs.mapSidebar.scrollTop = 0;
+    },
     dimensionalState(showing2D) {
       return showing2D ? '2d' : '3d';
     },
--- a/frontend/src/components/explorer/MapViewer.vue
+++ b/frontend/src/components/explorer/MapViewer.vue
@@ -9,25 +9,27 @@
       <template v-else>
         <div id="mapSidebar"
              class="column is-one-fifth-widescreen is-one-quarter-desktop
-                    is-one-quarter-tablet has-background-lightgray om-2
+                    is-one-quarter-tablet has-background-lightgray om-2 pt-0
                     fixed-height-desktop scrollable">
-          <div class="buttons has-addons is-centered padding-mobile m-0"
-               :title="`Switch to ${dimensionalState(!showing2D) }`"
-               @click="$store.dispatch('maps/toggleShowing2D')">
-            <button v-for="dim in [true, false]" :key="dim"
-                    class="button m-0"
-                    :class="dim === showing2D ? 'is-selected is-primary has-text-weight-bold' : 'is-light'"
-                    :disabled="currentMap && currentMap.type === 'custom'">
-              <span v-if="dim === showing2D" class="icon">
-                <i class="fa fa-check-square-o"></i>
-              </span>
-              <span v-if="dim !== showing2D">Switch to&nbsp;</span>
-              <span class="is-uppercase">{{ dimensionalState(dim) }}</span>
-            </button>
+          <div id="mapSidebar__header" class="has-background-lightgray pt-3">
+            <div class="buttons has-addons is-centered padding-mobile m-0"
+                 :title="`Switch to ${dimensionalState(!showing2D) }`"
+                 @click="$store.dispatch('maps/toggleShowing2D')">
+              <button v-for="dim in [true, false]" :key="dim"
+                      class="button m-0"
+                      :class="dim === showing2D ? 'is-selected is-primary has-text-weight-bold' : 'is-light'"
+                      :disabled="currentMap && currentMap.type === 'custom'">
+                <span v-if="dim === showing2D" class="icon">
+                  <i class="fa fa-check-square-o"></i>
+                </span>
+                <span v-if="dim !== showing2D">Switch to&nbsp;</span>
+                <span class="is-uppercase">{{ dimensionalState(dim) }}</span>
+              </button>
+            </div>
+            <SidebarDataPanels :dim="dimensionalState(showing2D)"
+                               :current-map="currentMap"
+                               :selection-data="selectionData" />
           </div>
-          <SidebarDataPanels :dim="dimensionalState(showing2D)"
-                             :current-map="currentMap"
-                             :selection-data="selectionData" />
           <div class="padding-mobile">
             <a class="button is-fullwidth is-primary is-inverted has-text-weight-bold is-hidden-tablet"
                @click="showingMapListing = !showingMapListing">
@@ -238,6 +240,14 @@ export default {
 
   #mapSidebar {
     word-wrap: break-word;
+
+    &__header {
+      @media screen and (min-width: $tablet) {
+        position: sticky;
+        top: 0;
+        z-index: 1;
+      }
+    }
 
     .buttons {
 
--- a/frontend/src/components/explorer/MapViewer.vue
+++ b/frontend/src/components/explorer/MapViewer.vue
@@ -11,7 +11,7 @@
              class="column is-one-fifth-widescreen is-one-quarter-desktop
                     is-one-quarter-tablet has-background-lightgray om-2 pt-0
                     fixed-height-desktop scrollable"
-             @scroll.passive="handleSidebarScroll">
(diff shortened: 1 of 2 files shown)


inghylt (approved):
Really an improvement of UX IMO, great work!

nanjiangshu (commented):
It's really a big improvement!

In some cases, when the height of the frozen panel is quite big, the map listing becomes very narrow. e.g. (/explore/Human-GEM/map-viewer/golgi_apparatus?dim=2d&panel=1&sel=ENSG00000151726&search=&coords=-4519.63,-40502.68,0.09,-14.07,5.41,250.28&g1=None&g2=None). I think it would be better to somehow restrict the maximum height for the frozen panel or even makes the frozen panel also scrollable or content of Gene card expandable.

nanjiangshu:
Another issue is that when one change to a different 2D map, the previously selected component is still shown. I think it is probably better to initialize this card after one changes to a new map, since the previously selected component is not valid anymore. I think this should be addressed in a separate issue.

mihai-sysbio:
>  frozen panel also scrollable

I feel that would be creating too many scrollable panels next to each other, which hampers the usability. Would it be okay instead to make just the first card sticky (the one with the map name)?

> previously selected component is still shown [...] in a separate issue

Thanks for creating #630.


- [ ] I noticed how the map listing slides under the sticky cards. I somehow felt the need for some space below the card that sits on top, otherwise there is too much white-on-white occasionally.

image

e0:
@mihai-sysbio I get what you mean and I tried to have some space below it as well. I personally think it helps that the there is no space so that it can look like the list of maps is being scrolled up and hid behind the card above. Or else it kind of feels like the list is just vanishing. Happy to consider if you have any other suggestions on how to improve this though.

e0:
> It's really a big improvement!
> 
> In some cases, when the height of the frozen panel is quite big, the map listing becomes very narrow. e.g. (/explore/Human-GEM/map-viewer/golgi_apparatus?dim=2d&panel=1&sel=ENSG00000151726&search=&coords=-4519.63,-40502.68,0.09,-14.07,5.41,250.28&g1=None&g2=None). I think it would be better to somehow restrict the maximum height for the frozen panel or even makes the frozen panel also scrollable or content of Gene card expandable.

Really nice that you tested and caught this case @nanjiangshu! I added a commit now that introduces some new behaviors.

1. When scrolling down the sidebar, the component card will now collapse.
2. Clicking the toggle to open the component card again will reset the scrolling.
3. Selecting a different component will open the component card again and reset the scrolling.

Ideally maybe it would be better to not reset the scrolling after opening the card again but I didn't come up with a clean solution to handle that for now. I personally kind of like the current behavior and hope it's ok for now, but please give this another look when you have time and let me know what you think.

frontend/src/components/explorer/mapViewer/SidebarDataPanels.vue
```diff
@@ -264,6 +264,17 @@ export default {
       return l.join('');
     },
   },
+  watch: {
+    selectionData(newData, oldData) {
+      if (
+        (newData.data && !oldData.data)
+        || (!newData.data && oldData.data)
+        || (newData.data && oldData.data && newData.data.id !== oldData.data.id)
+      ) {
```
mihai-sysbio, line 273: Hmm I don't intuitively understand, so I'm going to ask - why wouldn't this work?
```
      if (newData.data) {
```

frontend/src/components/explorer/mapViewer/SidebarDataPanels.vue
```diff
@@ -290,6 +301,20 @@ export default {
...
+    toggleSelectionCardContent() {
+      if (this.showSelectionCardContent) {
+        this.hideSelectionCardContent();
+      } else {
+        this.openSelectionCardContent();
+      }
+    },
+    openSelectionCardContent() {
+      this.showSelectionCardContent = true;
+      this.$emit('openSelectionCardContent');
```
mihai-sysbio, line 313: If there is no other way to implement the same behaviour without the `$emit`, then it would be acceptable to delete it and the subsequent scroll reset, and increase the timeout from 300ms to ~800ms.

frontend/src/components/explorer/mapViewer/SidebarDataPanels.vue
```diff
@@ -290,6 +301,20 @@ export default {
       this.showFullReactionListMissing = false;
       this.showFullReactionListMap = false;
     },
+    toggleSelectionCardContent() {
```
mihai-sysbio, line 304: This function could take in an optional variable to also achieve the same functionality as `openSelectionCardContent` and `closeSelectionCardContent`.

frontend/src/components/explorer/MapViewer.vue
```diff
@@ -145,6 +148,7 @@ export default {
   },
   async created() {
     this.handleQueryParamsWatch = debounce(this.handleQueryParamsWatch, 100);
+    this.handleSidebarScroll = debounce(this.handleSidebarScroll, 300);
```
mihai-sysbio, line 151: Instead of debouncing, could `this.$refs.mapSidebar.scrollTop` be `watch`-ed instead?

mihai-sysbio (changes requested):

frontend/src/components/explorer/mapViewer/SidebarDataPanels.vue
e0: What is wrong with `$emit`?

frontend/src/components/explorer/MapViewer.vue
nanjiangshu: @e0 This is a smart solution for the problem of too narrow map listing. I personally think the folding of the Gene card should happen as soon as a user starts to scroll the map listing, so that change the waiting time from 300ms to e.g. 50ms would be better. But if you think it is better with a longer waiting time, I'm OK with that.

e0: > @e0 This is a smart solution for the problem of too narrow map listing. I personally think the folding of the Gene card should happen as soon as a user starts to scroll the map listing, so that change the waiting time from 300ms to e.g. 50ms would be better. But if you think it is better with a longer waiting time, I'm OK with that.

Haha @mihai-sysbio actually suggested 800ms. Maybe we can book a quick meeting to talk about this or maybe in tomorrow's stand-up if there is time.

mihai-sysbio: I had suggested 800ms _if_ there is no way of keeping this behaviour after removing the scroll reset code.

frontend/src/components/explorer/mapViewer/SidebarDataPanels.vue
mihai-sysbio: Emits have effects in a non-deterministic way, ie one cannot deduce where event is consumed. For this reason, they should be avoided.

frontend/src/components/explorer/MapViewer.vue
mihai-sysbio: >  I personally think the folding of the Gene card should happen as soon as a user starts to scroll the map listing,

My previous suggestion is in line with this @nanjiangshu - as soon as `this.$refs.mapSidebar.scrollTo` becomes non-zero, the card may be collapsed (with or without delay).

frontend/src/components/explorer/mapViewer/SidebarDataPanels.vue
mihai-sysbio: [`.sync`](https://vuejs.org/v2/guide/components-custom-events.html#sync-Modifier) is a deterministic `$emit` substitute - it could be used with a `isCollapsed` prop.

frontend/src/components/explorer/mapViewer/SidebarDataPanels.vue
e0: I think this is where I personally prefer the clarity for the developers. For example when just looking at where the methods are being called (in markup or from another component), it's straightforward to know the intent without looking into the method definitions. It's not too much difference either way for this case though but that's my general stance. What do you think @inghylt and @nanjiangshu? I don't want to enforce having too many methods if you think it's unnecessary too.

frontend/src/components/explorer/MapViewer.vue
e0: > Instead of debouncing, could `this.$refs.mapSidebar.scrollTop` be `watch`-ed instead?

I'm not sure if that's possible but could look into it. What would the advantage be?

frontend/src/components/explorer/mapViewer/SidebarDataPanels.vue
mihai-sysbio: Oh right! I totally confused the `$emit`s.
But now that I looked into it - do you know if the strict child-parent communication is enforced? I couldn't find  anything about the strictness in the documentation, but I did find ways around it eg https://stackoverflow.com/questions/42615445/vuejs-2-0-emit-event-from-grand-child-to-his-grand-parent-component .

frontend/src/components/explorer/MapViewer.vue
mihai-sysbio: `debounce` basically groups all the events that happened in a desired interval. What I am after is to not trigger as many events to start with. Perhaps `watch`-ing is not the right way..

frontend/src/components/explorer/mapViewer/SidebarDataPanels.vue
e0: It's enforced in the sense that if you didn't have the binding `@openSelectionCardContent="resetSidebarLayout"` then it would not work since there are no listeners matching the emit.

There is a section about it here: https://vuejs.org/v2/guide/components.html#Listening-to-Child-Components-Events

frontend/src/components/explorer/mapViewer/SidebarDataPanels.vue
inghylt: I would say that in general, I prefer having multiple dense and straightforward methods. It's kind of context dependent though, e.g. in this particular case, it seems like the methods can be combined without loosing straightforwardness. But I would also be fine leaving it as it is right now

frontend/src/components/explorer/mapViewer/SidebarDataPanels.vue
e0: Hmm, this is a pattern I picked up previously to compare the ids and I didn't give this too much thought. I tested it out now and indeed it looks like that it is not necessary. Actually it looks like the checking could be removed completely. Fixed in: https://github.com/MetabolicAtlas/MetabolicAtlas/pull/628/commits/349a9bf0cb0cf0284313e0130966f58df1843997

frontend/src/components/explorer/MapViewer.vue
e0: I have added [a commit](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/628/commits/3c4cd536b2379548c5e3bf9da2a1698e27c8e8e2) that folds the data panel immediately after scrolling down instead of using the debounce to wait. It is also conditionally bound so that the scroll does not trigger the scroll handler once the panel is folded.

frontend/src/components/explorer/mapViewer/SidebarDataPanels.vue
e0: I just want to add that by definition, toggle means switching between two states so it doesn't 100% make sense for it to take an optional variable. By not taking a variable and using the three functions, the binary intent of each function is directly implied.

If this had to be one function maybe it should be renamed to something like `setSelectionCardContentVisibility` (probably needs a better name since visibility feels awkward).

Anyways, while this is worth a discussion, I think it is too minor of an issue to hold up this PR. So I will resolve this conversation without fixing it but I would love to discuss more about coding standards and practices.

mihai-sysbio (approved):
Great UX improvement! I did notice that the scroll gets reset after expanding the card (ie. the list of maps jumps to the top), but it's something that I don't necessarily consider has to be changed.

e0:
> Great UX improvement! I did notice that the scroll gets reset after expanding the card (ie. the list of maps jumps to the top), but it's something that I don't necessarily consider has to be changed.

Yes unfortunately that's the behavior without further complicating the implementation. So expanding the card manually or selecting another component when the card is folded have the same effect.

PR #632, opened 2021-06-18
Refactor: gem browser component template

### LINKED ISSUE #617: Use template styling components in gem browser

inghylt:
This PR closes #617

Please visit http://localhost/explore/Human-GEM/gem-browser and see if each component looks good and not too different from before.

There was some workaround needed for Metabolite ChEBI images and Reaction references, I think it looks good though

No chebi
http://localhost/explore/Human-GEM/gem-browser/metabolite/m02247l

Chebi
http://localhost/explore/Human-GEM/gem-browser/metabolite/m02604c

Chebi broken link
http://localhost/explore/Human-GEM/gem-browser/metabolite/m02576c

Reaction with references
http://localhost/explore/Human-GEM/gem-browser/reaction/HMR_4882

Also, some components used  `<maps-available :id="cName" :type="type" :element-i-d="''" />` before they were refactored. I did not add element-i-d as an option to reusable component layout since maps-available does not have or use :element-i-d prop. 

This PR also ensure proper responsiveness as discussed in https://github.com/MetabolicAtlas/MetabolicAtlas/pull/622#pullrequestreview-688373852

### DIFF
--- a/frontend/src/App.vue
+++ b/frontend/src/App.vue
@@ -379,7 +379,7 @@ html {
   }
 }
 
-#comparison-details, .model-table, .reaction-table, .gembrowser-table {
+#comparison-details, .table-template {
   .main-table tr td.td-key, #ed-table tr td.td-key {
     width: 150px;
   }
--- a/frontend/src/App.vue
+++ b/frontend/src/App.vue
@@ -379,7 +379,7 @@ html {
   }
 }
 
-#comparison-details, .metabolite-table, .model-table, .reaction-table, .subsystem-table {
+#comparison-details, .metabolite-table, .model-table, .reaction-table, .subsystem-table, .gembrowser-table {
   .main-table tr td.td-key, #ed-table tr td.td-key {
     width: 150px;
   }
--- a/frontend/src/App.vue
+++ b/frontend/src/App.vue
@@ -379,7 +379,7 @@ html {
   }
 }
 
-#comparison-details, .metabolite-table, .model-table, .reaction-table, .subsystem-table, .gembrowser-table {
+#comparison-details, .model-table, .reaction-table, .gembrowser-table {
   .main-table tr td.td-key, #ed-table tr td.td-key {
     width: 150px;
   }
--- a/frontend/src/components/explorer/gemBrowser/Compartment.vue
+++ b/frontend/src/components/explorer/gemBrowser/Compartment.vue
@@ -1,6 +1,6 @@
 <template>
   <component-layout
-    :component-type="'compartment'" :component-name="compartment.name"
+    component-type="compartment" :component-name="compartment.name"
     query-component-action="compartments/getCompartmentSummary"
     :include-reaction-table="false"
   >
@@ -52,11 +52,8 @@ export default {
   },
   data() {
     return {
-      modelNotFound: false,
       showFullSubsystem: false,
       limitSubsystem: 30,
-      componentNotFound: false,
-      showLoaderMessage: '',
     };
   },
   computed: {
--- a/frontend/src/components/explorer/gemBrowser/Compartment.vue
+++ b/frontend/src/components/explorer/gemBrowser/Compartment.vue
@@ -1,88 +1,57 @@
 <template>
-  <div class="section extended-section">
-    <div class="container is-fullhd">
-      <div v-if="modelNotFound || componentNotFound" class="columns is-centered">
-        <NotFound
-          :type="modelNotFound ? 'model' : type"
-          :component-id="modelNotFound ? $route.params.model : cName" />
-      </div>
-      <div v-else>
-        <div class="columns">
-          <div class="column">
-            <h3 class="title is-3"><span class="is-capitalized">{{ type }}</span> {{ compartment.name }}</h3>
-          </div>
-        </div>
-        <loader v-if="showLoaderMessage" :message="showLoaderMessage" class="columns" />
-        <div v-else class="columns is-multiline is-variable is-8">
-          <div class="subsystem-table column is-10-widescreen is-9-desktop is-full-tablet">
-            <div class="table-container">
-              <table v-if="compartment && Object.keys(compartment).length != 0"
-                     class="table main-table is-fullwidth">
-                <tr>
-                  <td class="td-key has-background-primary has-text-white-bis">Name</td>
-                  <td> {{ compartment.name }}</td>
-                </tr>
-                <tr>
-                  <td class="td-key has-background-primary has-text-white-bis">Subsystems</td>
-                  <td>
-                    <div v-html="subsystemListHtml"></div>
-                    <div v-if="!showFullSubsystem && subsystems.length > limitSubsystem">
-                      <br>
-                      <button class="is-small button" @click="showFullSubsystem=true">
-                        ... and {{ subsystems.length - limitSubsystem }} more
-                      </button>
-                    </div>
-                  </td>
-                </tr>
-                <tr>
-                  <td class="td-key has-background-primary has-text-white-bis">Reactions</td>
-                  <td> {{ compartment.reactionsCount }}</td>
-                </tr>
-                <tr>
-                  <td class="td-key has-background-primary has-text-white-bis">Metabolites</td>
-                  <td> {{ compartment.metabolitesCount }}</td>
-                </tr>
-                <tr>
-                  <td class="td-key has-background-primary has-text-white-bis">Genes</td>
-                  <td> {{ compartment.genesCount }}</td>
-                </tr>
-              </table>
+  <component-layout
+      :component-type="type" :component-name="compartment.name"
+      query-component-action="compartments/getCompartmentSummary"
+      :include-reaction-table="false"
+  >
+    <template v-slot:table>
+      <table v-if="compartment && Object.keys(compartment).length != 0"
+              class="table main-table is-fullwidth">
+        <tr>
+          <td class="td-key has-background-primary has-text-white-bis">Name</td>
+          <td> {{ compartment.name }}</td>
+        </tr>
+        <tr>
+          <td class="td-key has-background-primary has-text-white-bis">Subsystems</td>
+          <td>
+            <div v-html="subsystemListHtml"></div>
+            <div v-if="!showFullSubsystem && subsystems.length > limitSubsystem">
+              <br>
+              <button class="is-small button" @click="showFullSubsystem=true">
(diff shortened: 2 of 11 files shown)


mihai-sysbio:
> I did not add element-i-d as an option to reusable component layout since maps-available does not have or use :element-i-d prop.

@inghylt I think this `element-i-d` was previously used for centering and highlighting elements on the map, but now this is done via the url params. Good that your cleaned this up.

frontend/src/App.vue
```diff
@@ -379,7 +379,7 @@ html {
   }
 }
 
-#comparison-details, .metabolite-table, .model-table, .reaction-table, .subsystem-table {
+#comparison-details, .metabolite-table, .model-table, .reaction-table, .subsystem-table, .gembrowser-table {
   .main-table tr td.td-key, #ed-table tr td.td-key {
     width: 150px;
```
mihai-sysbio, line 384: I think `.gembrowser-table` should not have a fixed width, but instead leave the width to be decided by the parent component.

e0: It looks like this was done mostly to replicate the old CSS selectors of `.metabolite-table` and `.subsystem-table`, which could now be removed I think.

frontend/src/components/explorer/gemBrowser/Gene.vue
```diff
@@ -139,29 +97,7 @@ export default {
   watch: {
     '$route.params': 'setup',
   },
```
e0, line 99: These lines should be removed.

frontend/src/components/explorer/gemBrowser/Gene.vue
```diff
@@ -1,88 +1,46 @@
...
 import { mapGetters, mapState } from 'vuex';
-import NotFound from '@/components/NotFound';
-import Loader from '@/components/Loader';
-import ExtIdTable from '@/components/explorer/gemBrowser/ExtIdTable';
-import MapsAvailable from '@/components/explorer/gemBrowser/MapsAvailable';
-import ReactionTable from '@/components/explorer/gemBrowser/ReactionTable';
-import GemContact from '@/components/shared/GemContact';
+import ComponentLayout from '@/layouts/explorer/gemBrowser/ComponentLayout';
 import { generateSocialMetaTags, reformatTableKey } from '@/helpers/utils';
 import { default as messages } from '@/helpers/messages';
```
e0, line 38: This, along with `messages,` in `data` should be removed. This also applies for `Metabolite.vue`

e0 (changes requested):
Looks great @inghylt! There are some minor issues that should be fixed before merging imo.

frontend/src/components/explorer/gemBrowser/Compartment.vue
```diff
@@ -1,88 +1,57 @@
...
-                  <td class="td-key has-background-primary has-text-white-bis">Metabolites</td>
-                  <td> {{ compartment.metabolitesCount }}</td>
-                </tr>
-                <tr>
-                  <td class="td-key has-background-primary has-text-white-bis">Genes</td>
-                  <td> {{ compartment.genesCount }}</td>
-                </tr>
-              </table>
+  <component-layout
+    :component-type="type" :component-name="compartment.name"
```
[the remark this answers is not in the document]
inghylt: Did you perhaps intend to type `compartment` instead of `comparison` here? But yeah good suggestion!

frontend/src/App.vue
inghylt: Yeah  `.metabolite-table` and `.subsystem-table` should go. I'm not fond of removing `width: 150px` now though, that's kind of a major layout change which messes up the current layout. But I'm up for creating a new issue for it if you would like that

frontend/src/components/explorer/gemBrowser/Compartment.vue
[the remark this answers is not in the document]
e0: Haha indeed, sorry about that!

e0 (approved):

nanjiangshu (approved):
Great job @inghylt! The code is clearer now.

frontend/src/App.vue
mihai-sysbio: > I'm not fond of removing width: 150px now though

👍🏻 @inghylt let's leave that for some other refactoring session - I'm not tempted to record this in an issue.

However, following @e0's previous comment, it looks like all the CSS selectors on that line are used for the exact same purpose. I think it's better to have one and the same class for everything.

mihai-sysbio (changes requested):
Very nice refactoring!
I spotted that in the `Metabolite` and `Reaction` pages I'm getting this error in the console:
```
Invalid prop: type check failed for prop "referenceList". Expected Object, got Array 
```
It doesn't seem to break the functionality though.

frontend/src/components/explorer/gemBrowser/Reaction.vue
```diff
@@ -147,33 +113,8 @@ export default {
...
-        await this.$store.dispatch('reactions/getReactionData', payload);
-        this.componentNotFound = false;
-        this.showLoaderMessage = '';
-        await this.getRelatedReactions();
-      } catch {
-        this.componentNotFound = true;
-      }
-    },
-    async getRelatedReactions() {
+    async handleCallback() {
```
mihai-sysbio, line 117: Instead of triggering subsequent API calls via a callback, couldn't `query-component-action` be an array instead eg:
```
query-component-actions="['reactions/getReactionData','reactions/getRelatedReactionsForReaction']"
```

e0: Technically that would work I think but I'm not sure if it's an improvement because it doesn't really improve the intent. I might argue for renaming `handleCallback` to something like `handleComponentLoadedCallback`. What advantages do you see with having them in a list?

mihai-sysbio: I do see how the callback is needed for the metabolite page, but conceptually that is a different use case than loading data for the related reactions. As a developer I found this behaviour surprising, and less future proof - with a list it will be very easy to handle multi-request pages.

e0: I think currently it's clearer that there is one request needed for fetching a component and a subsequent function to handle stuff to do after the component has loaded. If it is a list then it is not clear the order of when things should be called. "Are the requests handled sequentially or concurrently?" Stylistically I think it would look awkward to have something like:

```javascript
await this.queryComponentActions[0];
if (this.queryComponentActions[1]) {
  await this.queryComponentActions[1];
}
```

And I think it would only be less and less intuitive how and when requests are being used if the list was to grow.

mihai-sysbio: The list would be handled concurrently, and I can't imagine it would be more than a handful of requests.

e0: I don't know if it makes sense for them to be handled concurrently, at least without rewriting a lot of the logic. For example if the URL contains an invalid metabolite id then it is unnecessary to send the `getRelatedMetabolites` request. Also since this is a refactoring issue I don't think we should go about changing the behavior of these components in this PR.

I was just saying that without looking into the `ComponentLayout` component, it wouldn't be straightforward for the developer to know how the requests are handled if they are provided in a list, whereas currently the intent of each binding is clearer.

In addition, I currently don't see any need to add a third function binding because all of the "post-component-loaded" logic can be wrapped into a single function, like we do in `Metabolite.vue`, where we make two separate requests. If we do ever want to change so we are sending all the requests concurrently, then all of them could be wrapped into the same function before passing into `ComponentLayout.vue` as well. In general I think it is difficult to try to be too future-proof because we don't know what changes we should accommodate for in the future.

mihai-sysbio: > I don't know if it makes sense for them to be handled concurrently, at least without rewriting a lot of the logic. For example if the URL contains an invalid metabolite id then it is unnecessary to send the `getRelatedMetabolites` request.

Users will rarely trigger such unnecessary requests, so it is acceptable to have to handle occasional bad requests.

> Also since this is a refactoring issue I don't think we should go about changing the behavior of these components in this PR.

I thought it was the right moment, because it is a refactoring issue, and specifically the refactoring that introduces the new behaviors of`query-component-action` and `@handleCallback`.

> I was just saying that without looking into the `ComponentLayout` component, it wouldn't be straightforward for the developer to know how the requests are handled if they are provided in a list, whereas currently the intent of each binding is clearer.

Right, now it's a first action triggering request that gets some data, and all other stuff in callback. While this is a separation in timings, I was arguing for separating the concerns: get all data requests (component-actions) together, and all non-data callbacks separately. If timing is an issue, one could always consider the first request in the array the "blocking" one, and all the rest concurrent. This would be controlled by an explicit behavior in the `setup` of `ComponentLayout` rather explicitly by using the `callback`.

> In addition, I currently don't see any need to add a third function binding because all of the "post-component-loaded" logic can be wrapped into a single function, like we do in `Metabolite.vue`, where we make two separate requests.

This wrapping together I find goes against the separation of concerns principle, and it is exactly what I think should be avoided. Perhaps it would be clearer if the props would be called `component-first-action` and `private-component-setup` or something like this, to communicate that the separation is about timing (but I still think the separation should be about concerns).

Anyway, it doesn't seem that I can change your mind about this, so I am resolving the conversation to signal that no changes are expected here.

e0: Although I was involved in some of the work here it's still @inghylt's PR and I don't want to assume authority over any of this. Was just curious about the list suggestion and provided my thoughts. So please unresolved this if you want to so others can get a chance to respond to your suggestion.

inghylt: Good suggestion. I do however feel like it will make things less clear and than they currently are, and I agree with the concerns @e0 raise. So I'm fine with keeping this conversation resolved

frontend/src/App.vue
inghylt: @mihai-sysbio  Good point. I think though that since `comparison-details` is not applied to a table, but all of the rest are, I would like to have `#comparison-details, .table-template` or similar

mihai-sysbio: sounds good!

mihai-sysbio (approved):
Well done!

PR #643, opened 2021-06-28
fix: disable custom map for 3D viewer

nanjiangshu:
This PR closes [#16](https://app.zenhub.com/workspaces/metabolic-atlas-sprint-607745622d49260019ce115a/issues/metabolicatlas/private-issues/16)

Description: Custom maps should not be shown in the sidebar for 3D viewer.

DoD: custom maps is shown on the sidebar for 2D map viewer (if available) but disappear when changing to 3D viewer.

### DIFF
--- a/frontend/src/components/explorer/mapViewer/MapsListing.vue
+++ b/frontend/src/components/explorer/mapViewer/MapsListing.vue
@@ -43,15 +43,7 @@ export default {
       mapsListing: state => state.maps.mapsListing,
     }),
     categories() {
-      const categories = [];
-      const li = Object.keys(this.mapsListing).filter(c => this.mapsListing[c].length > 0).sort();
-      for (let i = 0; i < li.length; i += 1) {
-        const category = li[i];
-        if (category !== 'customs' || this.showing2D) {
-          categories.push(category);
-        }
-      }
-      return categories;
+      return Object.keys(this.mapsListing).filter(c => this.mapsListing[c].length > 0 && (c !== 'customs' || this.showing2D)).sort();
     },
   },
   methods: {
--- a/frontend/src/components/explorer/mapViewer/MapsListing.vue
+++ b/frontend/src/components/explorer/mapViewer/MapsListing.vue
@@ -1,34 +1,32 @@
 <template>
   <div id="maps-listing">
-    <div v-for="category in Object.keys(mapsListing).filter(c => mapsListing[c].length > 0).sort()"
+    <div v-for="category in categories"
          :key="category" class="card my-3">
-      <template v-if="category !== 'customs' || showing2D">
-        <p class="is-capitalized is-size-6 has-text-weight-bold">{{ category.replace(/.$/," maps") }}</p>
-        <span v-for="item in mapsListing[category]" :key="item.id">
-          <template v-if="showing2D">
-            <template v-if="item.svgs.length === 0">
-              {{ item.name }}
-            </template>
-            <template v-else-if="item.svgs.length === 1">
-              <a @click="changeToMap(item.svgs[0].id)">
-                {{ item.svgs[0].customName }}
-              </a>
-            </template>
-            <template v-else>
-              {{ item.name }}:
-              <a v-for="svg in [...item.svgs].sort((a, b) => a.customName.localeCompare(b.customName))"
-                :key="svg.id" class="inline" @click="changeToMap(svg.id)">
-                {{ svg.customName }}
-              </a>
-            </template>
+      <p class="is-capitalized is-size-6 has-text-weight-bold">{{ category.replace(/.$/," maps") }}</p>
+      <span v-for="item in mapsListing[category]" :key="item.id">
+        <template v-if="showing2D">
+          <template v-if="item.svgs.length === 0">
+            {{ item.name }}
           </template>
-          <template v-else>
-            <a @click="changeToMap(item.id)">
-              {{ item.name }}
+          <template v-else-if="item.svgs.length === 1">
+            <a @click="changeToMap(item.svgs[0].id)">
+              {{ item.svgs[0].customName }}
             </a>
           </template>
-        </span>
-      </template>
+          <template v-else>
+            {{ item.name }}:
+            <a v-for="svg in [...item.svgs].sort((a, b) => a.customName.localeCompare(b.customName))"
+              :key="svg.id" class="inline" @click="changeToMap(svg.id)">
+              {{ svg.customName }}
+            </a>
+          </template>
+        </template>
+        <template v-else>
+          <a @click="changeToMap(item.id)">
+            {{ item.name }}
+          </a>
+        </template>
+      </span>
     </div>
   </div>
 </template>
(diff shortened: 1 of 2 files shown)


mihai-sysbio (approved):
This works fine. My approach would have been to remove the `id` property for the maps without a 3D view, and let the 2D view continue to rely on the `svgs` array. However, I don't think that would have been as easy to understand.
Also, I noticed a weird styling bug, but it might be that this will go away on its own.
image

frontend/src/components/explorer/mapViewer/MapsListing.vue
```diff
@@ -2,31 +2,33 @@
...
-      <span v-for="item in mapsListing[category]" :key="item.id">
-        <template v-if="showing2D">
-          <template v-if="item.svgs.length === 0">
-            {{ item.name }}
-          </template>
-          <template v-else-if="item.svgs.length === 1">
-            <a @click="changeToMap(item.svgs[0].id)">
-              {{ item.svgs[0].customName }}
-            </a>
+      <template v-if="category !== 'customs' || showing2D">
```
e0, line 5: Since this `v-if` is inside of the `v-for`, the `div` from line 3 is still being rendered, which is causing the visual bug @mihai-sysbio mentioned [here](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/643#pullrequestreview-694091438).

A suggestion is to combine the logic from `v-for` and `v-if` into a computed property, perhaps called `categories`.

frontend/src/components/explorer/mapViewer/MapsListing.vue
```diff
@@ -2,31 +2,33 @@
...
+            </template>
+            <template v-else-if="item.svgs.length === 1">
+              <a @click="changeToMap(item.svgs[0].id)">
+                {{ item.svgs[0].customName }}
+              </a>
+            </template>
+            <template v-else>
+              {{ item.name }}:
+              <a v-for="svg in [...item.svgs].sort((a, b) => a.customName.localeCompare(b.customName))"
+                :key="svg.id" class="inline" @click="changeToMap(svg.id)">
```
e0, line 20: The linter wants another space here but maybe this will get auto-resolved if you apply the fix mentioned in the other message since you would be able to remove the `<template v-if="showing2D">` wrapper.

e0 (changes requested):

frontend/src/components/explorer/mapViewer/MapsListing.vue
nanjiangshu: Thanks a lot @mihai-sysbio and @e0 for pointing out this problem. I will try to solve it with Shan's suggestion.

nanjiangshu: Hi @mihai-sysbio and @e0, I've added a computed `categories` to solve the problem mentioned above.

frontend/src/components/explorer/mapViewer/MapsListing.vue
nanjiangshu: Hi Shan, it seems the logic `<template v-if="showing2D">` needs to be kept since the treatment for the svg maps are slightly different from the 3D maps. I've added an extra space to solve the linter warning message anyway.

frontend/src/components/explorer/mapViewer/MapsListing.vue
```diff
@@ -42,6 +42,17 @@ export default {
...
+    categories() {
+      const categories = [];
+      const li = Object.keys(this.mapsListing).filter(c => this.mapsListing[c].length > 0).sort();
+      for (let i = 0; i < li.length; i += 1) {
+        const category = li[i];
+        if (category !== 'customs' || this.showing2D) {
+          categories.push(category);
+        }
+      }
+      return categories;
```
e0, line 54: I think the following might do the same, please test it out locally though as I'm not on that branch atm.

```javascript
return Object.keys(this.mapsListing).filter(c => this.mapsListing[c].length > 0 && (c !== 'customs' || this.showing2D)).sort();
```

e0 (approved):
This looks good, just added a comment about possible further refactoring.

frontend/src/components/explorer/mapViewer/MapsListing.vue
e0: Ok, I think we are on the same page here. I meant that the logic for `showing2D` should still be kept and moved into the computed property and then the actual `<template>` can be removed, which looks like to be exactly what you have done.

frontend/src/components/explorer/mapViewer/MapsListing.vue
nanjiangshu: @e0 Works like a charm! I'll take your suggestion for refactoring.

inghylt (changes requested):
Nice work! I tried out the latest commits and it looks good. However, when a custom map is chosen, you can click 3D even though the 3D button is disabled. That leads to the entire yeast model being rendered in 3D. I think it makes sense to fix this within this PR

e0:
> Nice work! I tried out the latest commits and it looks good. However, when a custom map is chosen, you can click 3D even though the 3D button is disabled. That leads to the entire yeast model being rendered in 3D. I think it makes sense to fix this within this PR

@nanjiangshu This was a good catch by @inghylt and sorry that I missed it in my review. I believe it is just about making the click event here conditional.

```
          <div class="buttons has-addons is-centered padding-mobile m-0"
               :title="`Switch to ${dimensionalState(!showing2D) }`"
               @click="$store.dispatch('maps/toggleShowing2D')">

```

frontend/src/components/explorer/mapViewer/MapsListing.vue
```diff
@@ -44,6 +42,17 @@ export default {
       showing2D: state => state.maps.showing2D,
       mapsListing: state => state.maps.mapsListing,
     }),
+    categories() {
+      const categories = [];
+      const li = Object.keys(this.mapsListing).filter(c => this.mapsListing[c].length > 0).sort();
+      for (let i = 0; i < li.length; i += 1) {
+        const category = li[i];
+        if (category !== 'customs' || this.showing2D) {
```
mihai-sysbio: Wouldn't it make more sense to include the check
> category !== 'customs' 

in the `filter` above?

mihai-sysbio: Sorry, now I saw this was done in a later commit!

nanjiangshu:
Thanks @inghylt for catching this problem and @e0 for pointing out the place for solving the problem. I've push a commit for addressing this issue.

e0:
> Thanks @inghylt for catching this problem and @e0 for pointing out the place for solving the problem. I've push a commit for addressing this issue.

Nice! Looks like there is a merge conflict now unfortunately, do you think you can fix it? If it looks messy just let me know and I can try to fix it.

nanjiangshu:
Hi Shan @e0 , I used `git rebase develop` to fix the merging conflict. Hope it is done in the correct way.

e0:
@nanjiangshu It's great that you used `git rebase`! I think it looks right based on the order of commits. GitHub has a problem where it doesn't update the "Files changed" section sometimes. A workaround is to edit this PR and change the target branch from `develop` to something else, save, and then change back to `develop` and save. Then it should only show the files you have worked on in this PR. Please give that a try and see if it works for you.

nanjiangshu:
Thanks for the tip @e0. It seems the `Files changed` has been updated after this trick :)

nanjiangshu:
> Nice work! I tried out the latest commits and it looks good. However, when a custom map is chosen, you can click 3D even though the 3D button is disabled. That leads to the entire yeast model being rendered in 3D. I think it makes sense to fix this within this PR

Hi @inghylt, do you think it is OK to merge now?

e0:
> Thanks for the tip @e0. It seems the `Files changed` has been updated after this trick :)

Nice, looks good!

inghylt (approved):
Nice, well done!

PR #644, opened 2021-06-29
fix: broken ascii arrows in reaction table

inghylt:
This PR fixes the broken ascii symbols and links in the Reaction Tables. 

Some examples:

/explore/Human-GEM/gem-browser/metabolite/mdz_s
/explore/Human-GEM/gem-browser/metabolite/m01202c
/explore/Human-GEM/gem-browser/metabolite/m03106c

### DIFF
--- a/frontend/src/components/explorer/gemBrowser/ReactionTable.vue
+++ b/frontend/src/components/explorer/gemBrowser/ReactionTable.vue
@@ -62,9 +62,9 @@
                 </template>
               </td>
               <td>
-                <template v-for="(RP, i) in r.compartment_str.split(' => ')">
+                <template v-for="(RP, i) in r.compartment_str.split(new RegExp(['&#8658;|&#8660;']))">
                   <template v-if="i !== 0">{{ r.reversible ? ' &#8660; ' : ' &#8658; ' }}</template>
-                  <template v-for="(compo, j) in RP.split(' + ')">
+                  <template v-for="(compo, j) in RP.split(' ,')">
                     <template v-if="j != 0"> + </template>
                     <!-- eslint-disable-next-line vue/valid-v-for vue/require-v-for-key max-len -->
                     <a :href="`/explore/${model.short_name}/gem-browser/compartment/${idfy(compo)}`" @click="handleRouterClick">{{ compo }}</a>
--- a/frontend/src/components/explorer/gemBrowser/ReactionTable.vue
+++ b/frontend/src/components/explorer/gemBrowser/ReactionTable.vue
@@ -62,12 +62,12 @@
                 </template>
               </td>
               <td>
-                <template v-for="(RP, i) in r.compartment_str.split(/&#8658;|&#8660;/)">
-                  <template v-if="i !== 0">{{ r.reversible ? ' &#8660; ' : ' &#8658; ' }}</template>
-                  <template v-for="(compo, j) in RP.split(' ,')">
+                <template v-for="(w, i) in r.compartment_str.split(/⇔|⇒/)">
+                  <template v-if="i !== 0">{{ r.reversible ? ' ⇔ ' : ' ⇒ ' }}</template>
+                  <template v-for="(comp, j) in w.split(' + ')">
                     <template v-if="j != 0"> + </template>
                     <!-- eslint-disable-next-line vue/valid-v-for vue/require-v-for-key max-len -->
-                    <a :href="`/explore/${model.short_name}/gem-browser/compartment/${idfy(compo)}`" @click="handleRouterClick">{{ compo }}</a>
+                    <a :href="`/explore/${model.short_name}/gem-browser/compartment/${idfy(comp)}`" @click="handleRouterClick">{{ comp }}</a>
                   </template>
                 </template>
               </td>
--- a/frontend/src/components/explorer/gemBrowser/ReactionTable.vue
+++ b/frontend/src/components/explorer/gemBrowser/ReactionTable.vue
@@ -62,7 +62,7 @@
                 </template>
               </td>
               <td>
-                <template v-for="(RP, i) in r.compartment_str.split(new RegExp(['&#8658;|&#8660;']))">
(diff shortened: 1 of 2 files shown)


frontend/src/components/explorer/gemBrowser/ReactionTable.vue
```diff
@@ -62,9 +62,9 @@
                 </template>
               </td>
               <td>
-                <template v-for="(RP, i) in r.compartment_str.split(' => ')">
+                <template v-for="(RP, i) in r.compartment_str.split(/&#8658;|&#8660;/)">
```
e0, line 65: Just a note that variable names shouldn't start with capital letters. @inghylt I know that you didn't do this but just wanted to mention it. We might want to have some code cleaning PR for this type of things.

frontend/src/components/explorer/gemBrowser/ReactionTable.vue
```diff
@@ -62,9 +62,9 @@
...
                   <template v-if="i !== 0">{{ r.reversible ? ' &#8660; ' : ' &#8658; ' }}</template>
-                  <template v-for="(compo, j) in RP.split(' + ')">
+                  <template v-for="(compo, j) in RP.split(' ,')">
```
e0, line 67: I find that it's a bit difficult to comprehend with the nested `v-for`. Personally I would prefer something like the following but I'm not sure if that's an improvement.

```html
<td v-html="reformatCompartmentStr(r.compartment_str)" />
```

```javascript
reformatCompartmentStr(compartmentStr) {
  // matches any sequence that is not whitespace
  return compartmentStr.replace(/[^\s]+/g, (w) => {
    if (w === ',') {
      return '+';
    }

    if (w.match(/(&#8658;|&#8660)/)) {
      return w;
    }

    return `<a href="/explore/${this.model.short_name}/gem-browser/compartment/${idfy(w)}" @click="handleRouterClick">${w}</a>`;
  });
},
```

e0 (approved):
Looks like the bug is solved so I'm approving this but added a few notes for some potential refactoring.

frontend/src/components/explorer/gemBrowser/ReactionTable.vue
inghylt: @nanjiangshu  @mihai-sysbio What's your preference regarding this? I think this function is cleaner but possibly slightly easier to read the code as it is right now inside the template

frontend/src/components/explorer/gemBrowser/ReactionTable.vue
nanjiangshu: I agree. A variable name like `RP` is often used to define a constant, which is not the case here. But I think we can create a separate issue for this kind of refactoring.

frontend/src/components/explorer/gemBrowser/ReactionTable.vue
nanjiangshu: Hi @inghylt. I think both are OK but prefer the use of a function :)

nanjiangshu (approved):
Hi @inghylt, great work! I can approve the PR for whatever option you use for the template.

frontend/src/components/explorer/gemBrowser/ReactionTable.vue
mihai-sysbio: At the same time, these lines were already modified in this PR, so it wouldn't be wrong to rename the variable now.

mihai-sysbio (changes requested):
I think the focus of the PR should be to resolve the issue at source. It seems to come from  `frontend/helpers/utils.js`:
```
export function equationSign(isReversible) {
  return isReversible ? '&#8660;' : '&#8658;';
}
```
So instead of patching this by regex-ing the constructed link, we should probably look into how the link it put together in the first place, and have a special function/condition for constructing this text specifically for this use case.

Also, in the example for `m01202c`, the last compartment before the arrow is `peroxisome_` which leads to a blank (broken) page.

e0 (approved):

frontend/src/components/explorer/gemBrowser/ReactionTable.vue
```diff
@@ -62,12 +62,12 @@
                 </template>
               </td>
               <td>
-                <template v-for="(RP, i) in r.compartment_str.split(/&#8658;|&#8660;/)">
-                  <template v-if="i !== 0">{{ r.reversible ? ' &#8660; ' : ' &#8658; ' }}</template>
-                  <template v-for="(compo, j) in RP.split(' ,')">
+                <template v-for="(w, i) in r.compartment_str.split(/⇔|⇒/)">
+                  <template v-if="i !== 0">{{ r.reversible ? ' ⇔ ' : ' ⇒ ' }}</template>
```
mihai-sysbio, line 66: Would it make sense to define `⇔` and `⇒` as constants, since they need to be the same as in the `equationSign` function?

mihai-sysbio (approved):
Nice, I see this PR also fixes the broken arrows in the Interaction Partners, which wasn't spotted when the issue was described.

frontend/src/components/explorer/gemBrowser/ReactionTable.vue
mihai-sysbio: Like @inghylt, I too have a preference for keeping as much of the html creation in the `<template>` section of the Vue component. This has benefits in terms of eg syntax highlighting.

frontend/src/components/explorer/gemBrowser/ReactionTable.vue
inghylt: Since we don't use global constants elsewhere, I don't feel like it makes sense to add that just for this quite small case

mihai-sysbio: We use `messages.js` like global constants (strings), and we use `utils.js` for shared functions. But indeed, there is no dedicated place for this type of constants.

nanjiangshu (approved):
Great work!

Issue #648, opened 2021-07-02
Missing reaction popup modal not shown as expected


nanjiangshu:
As a user, I would like the modal window popup to the front when clicking the `See comparison` in the left sidebar to 2D map viewer. However, as shown in the attached screenshots, the modal window is displayed behind some components of the map viewer. 

This issue might be a good application for `git bisect`

(for Chrome)
![Screenshot 2021-07-02 at 11.30.08.png](https://images.zenhubusercontent.com/58ca67a651aa2c305778a403/a1f7e0a6-f626-480e-a81b-674c0f62c682) 

(for Firefox and Safari)
![Screenshot 2021-07-02 at 19.50.56.png](https://images.zenhubusercontent.com/58ca67a651aa2c305778a403/2f858347-c1d9-40a7-9928-d8485d0cd712)

inghylt:
Seems like commit [349a9bf0cb0cf0284313e0130966f58df1843997](https://github.com/MetabolicAtlas/MetabolicAtlas/commit/349a9bf0cb0cf0284313e0130966f58df1843997) introduced the bug

e0:
> Seems like commit [349a9bf0cb0cf0284313e0130966f58df1843997](https://github.com/MetabolicAtlas/MetabolicAtlas/commit/349a9bf0cb0cf0284313e0130966f58df1843997) introduced the bug

Interesting. I can imagine if the associated PR to this commit causes the bug but it's pretty wild to me that this commit itself would cause it.

inghylt:
Yeah it's weird. The commit before [76fbf2cceaaffca8dc12ba6e8f917a2ebf447e4b](https://github.com/MetabolicAtlas/MetabolicAtlas/commit/76fbf2cceaaffca8dc12ba6e8f917a2ebf447e4b) and all previous ones works fine though so it legit seems to be introduced by that

e0:
> Yeah it's weird. The commit before [76fbf2cceaaffca8dc12ba6e8f917a2ebf447e4b](https://github.com/MetabolicAtlas/MetabolicAtlas/commit/76fbf2cceaaffca8dc12ba6e8f917a2ebf447e4b) and all previous ones works fine though so it legit seems to be introduced by that

That's some great detective work. I guess we could just revert this commit or maybe try [the variation](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/628#discussion_r656197559) @mihai-sysbio suggested and see if it goes away.

[...]

PR #651, opened 2021-07-09
fix: broken missing reactions modal

inghylt:
This PR closes issue #648

Please test out MapViewer and see if everything is working as it should. Please try out edge cases, such as switching between 2d and 3d, selecting another map while inside 2d and 3d and so on, to see if there are any sneaky bugs I have missed.

The modal (now in a separate component called MissingReactionModal) was moved to not be a child under `<div id="mapSidebar__header"...>` to avoid the bug introduced by the new styling in commit 8f43b0df6104b255bd9f9a60b9be8b46c683bb33. Some related information is available at https://stackoverflow.com/questions/31646746/z-index-behaviour-is-different-in-chrome-to-firefox/31710969#31710969

### DIFF
--- a/frontend/src/components/explorer/mapViewer/SidebarDataPanels.vue
+++ b/frontend/src/components/explorer/mapViewer/SidebarDataPanels.vue
@@ -9,16 +9,15 @@
       </header>
       <div v-if="dim==='2d' && currentMap.reactionList && missingReactionList.length > 0"
            class="card-content p-4">
-        <div v-if="currentMap.mapReactionIdSet.length == 1" class="content mb-0">
+        <template v-if="currentMap.mapReactionIdSet.length == 1">
           Please note that {{ missingReactionList.length }}
           of the reactions in the {{ currentMap.type }} are not shown on the map.
-          <a @click="$emit('update:showModal', true)"> See comparison </a>
-        </div>
-        <div v-else class="content mb-0">
+        </template>
+        <template v-else>
           Please note that {{ missingReactionList.length }} of the reactions in the
           {{ currentMap.type }} are not shown on any of the {{ currentMap.name }} maps.
-          <a @click="$emit('update:showModal', true)"> See comparison </a>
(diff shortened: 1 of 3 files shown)


frontend/src/components/explorer/mapViewer/SidebarDataPanels.vue
```diff
@@ -12,69 +12,12 @@
...
-                        ... and {{ mapReactionList.length - displayedReaction }} more
-                      </button>
-                    </div>
-                  </td>
-                </tr>
-              </tbody>
-            </table>
-          </div>
-          <button class="modal-close is-large" @click="closeModal"></button>
+          <a @click="$emit('update:showModal', true)"> See comparison </a>
```
e0, line 20: Minor thing and I don't think it was exactly introduced in this PR but we can avoid a bit more repetition with the following.

```suggestion
          <template v-if="currentMap.mapReactionIdSet.length == 1">
            Please note that {{ missingReactionList.length }}
            of the reactions in the {{ currentMap.type }} are not shown on the map.
          </template>
          <template v-else>
            Please note that {{ missingReactionList.length }} of the reactions in the
            {{ currentMap.type }} are not shown on any of the {{ currentMap.name }} maps.
          </template>
          <a @click="$emit('update:showModal', true)"> See comparison </a>
```

e0 (approved):
Looks great @inghylt! Really clean refactoring. I left a minor suggestion but it's not needed so I'm approving it for now. Have tested in Firefox, Chrome, and Safari on macOS.

frontend/src/components/explorer/mapViewer/SidebarDataPanels.vue
inghylt: Nice suggestion! Fixed with 95d825942bddc73f0a804e0751727527fa0e85db

mihai-sysbio (approved):
Excellent refactoring!

PR #652, opened 2021-07-12
fix: use styling to fix the height of the 3D viewer on mobile

mihai-sysbio:
The fix proposed by @e0 has proven to work - it was tested from the dev server and Android, and locally with Firefox/Chrome and responsive mode. Since this introduced a second use of the height value, a variable was defined.

### DIFF
--- a/frontend/src/style/vars.scss
+++ b/frontend/src/style/vars.scss
@@ -28,3 +28,5 @@ $black: hsl(0, 0%, 4%);
 // Custom variables for fullheight Map Viewer, same as Bulma defaults
 $navbar-height: 4rem;
 $footer-height: 3.5rem;
+
+$viewer-height: 450px;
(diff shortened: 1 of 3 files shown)


e0 (approved):

frontend/src/style/vars.scss
```diff
@@ -28,3 +28,5 @@ $black: hsl(0, 0%, 4%);
 // Custom variables for fullheight Map Viewer, same as Bulma defaults
 $navbar-height: 4rem;
 $footer-height: 3.5rem;
+
+$viewer-height: 450px;
```
inghylt, line 32: Should we perhaps add a newline?

mihai-sysbio: Ah you mean at the end of the file? The linter does not seem to complain as far as I can see.
Another idea would be to add a comment explaining what this variable is for.

inghylt: Github made a red remark (hidden by these discussion I think), not a major thing obv. but maybe nice to fix. Adding a comment is a good idea!

PR #653, opened 2021-07-12
fix: SVG maps not loading on mobile 

e0:
This closes https://github.com/MetabolicAtlas/private-issues/issues/14

### DIFF
--- a/frontend/src/components/explorer/mapViewer/Svgmap.vue
+++ b/frontend/src/components/explorer/mapViewer/Svgmap.vue
@@ -298,7 +298,7 @@ export default {
 
         this.processSelSearchParam();
         this.$store.dispatch('maps/setLoading', false);
-      }, 0);
+      }, 50);
     },
     handleWheelEvent(event) {
       event.preventDefault();
--- a/frontend/src/components/explorer/mapViewer/Svgmap.vue
+++ b/frontend/src/components/explorer/mapViewer/Svgmap.vue
@@ -55,7 +55,7 @@ export default {
       panzoom: null,
       panzoomOptions: {
         maxScale: 1,
-        minScale: 0.03,
+        minScale: 0.01,
         step: 0.1,
         canvas: true,
       },
--- a/frontend/src/components/explorer/mapViewer/Svgmap.vue
+++ b/frontend/src/components/explorer/mapViewer/Svgmap.vue
@@ -253,6 +253,11 @@ export default {
         return;
       }
 
+      if (window.scrollY > 0) {
+        // for optimal SVG positioning, the calculation expects window.scrollY to be 0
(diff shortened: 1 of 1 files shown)


mihai-sysbio:
I tried the fix on desktop in responsive mode. The original bug can be reproduced, and the fix seems to mostly work. However, it still doesn't work for some maps, eg when going from the home page, click on the map viewer big button, and then choosing the Part 3 Cytosol map.

e0:
> I tried the fix on desktop in responsive mode. The original bug can be reproduced, and the fix seems to mostly work. However, it still doesn't work for some maps, eg when going from the home page, click on the map viewer big button, and then choosing the Part 3 Cytosol map.

Thanks for testing. I'm looking into this now. It's interesting because it's not exactly the same issue. Previously the maps would not be centered correctly after selecting from the list, but if you reload the page without the coords in the params (e.g. `/explore/Human-GEM/map-viewer/aminoacyl_trna_biosynthesis?dim=2d`), the map would appear. For Cytosol and some other larger compartments, they do not load after this type of reload either.

e0:
> I tried the fix on desktop in responsive mode. The original bug can be reproduced, and the fix seems to mostly work. However, it still doesn't work for some maps, eg when going from the home page, click on the map viewer big button, and then choosing the Part 3 Cytosol map.

Addressed in: https://github.com/MetabolicAtlas/MetabolicAtlas/pull/653/commits/379953689e6d6372cce15385c903efd1266b8562

The problem was that the minimum scale was preventing the larger maps from being properly fit and centered on smaller screen sizes.

mihai-sysbio:
> The problem was that the minimum scale was preventing the larger maps from being properly fit and centered on smaller screen sizes.

Oh how unexpected. Great that you figured it out.

e0:
Thanks to @inghylt for mentioning that there was occasionally still times where the map would not get centered properly. It seemed a bit random, I was able to spot it when switching maps a few times but difficult to replicate the issue. I added a new fix that adds a slight delay (50ms) before performing the calculations and I haven't been able to trigger the same issue.

mihai-sysbio (approved):
Nice fixes! I can confirm they work both locally (responsive mode) and on the dev server (Android, Firefox). I did notice that the largest map (mitochondria) does not center correct on mobile when loading from the dev server, but I'm not sure if it make sense to increase the 50ms delay for this reason.

mihai-sysbio:
> Thanks for testing and deploying to the dev server. I tested on my phone (mobile Safari) now and Mitochondria seems to center correctly.

It took a few tries but in the end it loaded for me as well. It's definitely good enough, these maps need a desktop screen anyway.

inghylt (approved):
Nice work!

PR #658, opened 2021-07-19
fix: change layout of 2d viewer (to like in 3d) and remove previous workaround fix for enabling search

### LINKED ISSUE #638: text selection in Map Viewer search bar
As a user, I want to select the text in the search bar of the Map Viewer, so that I can edit it conveniently or copy-paste it. This was reported by @johan-gson.

e0:
This closes #638 

This also fixes problems where the mouse "clicks through" the search and map controls. For example if you dragged over the search bar it would pan the map, this is no longer the behavior so interacting with the search and map controls do not interfere with the map itself.

### DIFF
--- a/frontend/src/components/explorer/mapViewer/MapSearch.vue
+++ b/frontend/src/components/explorer/mapViewer/MapSearch.vue
@@ -33,7 +33,6 @@
 
 import { mapState } from 'vuex';
 import { debounce } from 'vue-debounce';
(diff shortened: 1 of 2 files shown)


mihai-sysbio (approved):
Interesting refactoring. It worked great in all scenarios tested (2D/3D switching, overlay, mobile). I'm really happy to see this one solved, especially while removing another `event-bus` call 🤩

PR #659, opened 2021-07-19
fix: clear error message map is found

### LINKED ISSUE #645: Warning message "2D map is not available" remains once it appears
As a user, I would like the warning message appears when a 2D map is not available but disappears after selecting an existing 2D map. 

How to reproduce the bug:

* Choose "Yeast Gem" in Map Viewer
* Select `Fatty acid biosynthesis`
* Switch to 3D viewer
* Select `sce00791 atrazine degradation`
* Switch to 2D viewer 
The warning message will appear 
* select `Fatty acid biosynthesis` again 
The warning message remains

e0:
This closes #645

### DIFF
--- a/frontend/src/components/explorer/mapViewer/Svgmap.vue
+++ b/frontend/src/components/explorer/mapViewer/Svgmap.vue
@@ -150,6 +150,7 @@ export default {
         this.errorMessage = messages.mapNotFound;
         return;
       }
+      this.errorMessage = '';
       this.$store.dispatch('maps/setLoading', true);
       const payload = { model: this.model.short_name, svgName: this.mapData.svgs[0].filename };
       await this.$store.dispatch('maps/getSvgMap', payload);
--- a/frontend/src/components/explorer/mapViewer/Svgmap.vue
+++ b/frontend/src/components/explorer/mapViewer/Svgmap.vue
@@ -147,6 +147,7 @@ export default {
     async init() {
       this.$refs.mapsearch.reset();
       if (this.mapData.svgs.length === 0) {
+        this.$store.dispatch('maps/clearSvgMap');
         this.errorMessage = messages.mapNotFound;
         return;
       }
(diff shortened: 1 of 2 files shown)


mihai-sysbio (approved):
The fix works. I did notice a related bug, however; here is how to reproduce it: select a 2D map for Human-GEM, switch to 3D, select `Arachidonic acid metabolism` (a subsystem with no 2D map). and then switch back to 2D. While the error shows as expected, I did also see a part of the previously loaded map. This could (should?) be reported as a separate issue though.

e0:
> The fix works. I did notice a related bug, however; here is how to reproduce it: select a 2D map for Human-GEM, switch to 3D, select `Arachidonic acid metabolism` (a subsystem with no 2D map). and then switch back to 2D. While the error shows as expected, I did also see a part of the previously loaded map. This could (should?) be reported as a separate issue though.

I tried to trigger this bug following the steps but don't think I saw what you described. Could you add a screenshot or quick screen recording of what you see? I first tested this in the production site just to help see if it was something that was introduced in this PR but didn't see it there. Then I tested in my local machine using this branch and wasn't able to trigger it either.

mihai-sysbio:
It's definitely not something introduced in this PR, as it is reproducible in production, in Firefox, Chrome and Safari (all desktop). Maybe it's another quick fix, maybe not.

https://user-images.githubusercontent.com/23480589/126544140-75b7cb0d-6557-4130-8265-7c0f83e3710d.mp4

e0:
@mihai-sysbio Thanks for the video. I can verify now that I can trigger the same error from my side. I didn't look carefully enough and my browser window was not tall enough to show any visible part of the map, so it was just a white background. I think the bug is closely related enough that we fix it in the same PR. I will try to fix it now and request a new review.

mihai-sysbio (approved):
Awesome. Works as advertised 😁

PR #661, opened 2021-07-23
feat: content updates

mihai-sysbio:
This PR shares the news between the home page and the about page as described in https://github.com/MetabolicAtlas/private-issues/issues/94, and updates other content relevant for the 2.0 release listed in https://github.com/MetabolicAtlas/private-issues/issues/77.


.gitignore
```diff
@@ -51,14 +51,7 @@ bh_unicode_properties.cache
 GitHub.sublime-settings
 
 # local env files
-.env.local
-.env.*.local
-.env
-*.env
-
-static/
```
e0, line 59: This makes it so that the `api/static/logo.png` appears untracked. If it is meant to be included in the repo then I think it should be added (`git add`).

frontend/src/components/Home.vue
```diff
@@ -250,35 +256,8 @@ export default {
...
-          text: 'Metabolic Atlas is upgraded to v1.2 with Human-GEM updated to v1.1' },
-        { date: '2019.05.29',
-          hash: '29-May-2019',
-          text: 'Metabolic Atlas is upgraded to v1.1' },
-        { date: '2019.05.17',
-          hash: '17-May-2019',
-          text: 'Metabolic Atlas is publicly available as v1.0' },
-      ],
+      newsItems:
+        Object.values(newsItems).reverse().flat(1).slice(0, 12),
```
e0, line 260: Nice!

frontend/src/components/Home.vue
```diff
@@ -68,8 +73,8 @@
                   <tr v-if="el.date">
                     <td>{{ el.date }}</td>
```
e0, line 74: The new date format (dash instead of dot as separator) allows the dates to wrap into multiple lines. One way to fix this could be to have a `white-space:nowrap;` rule on the relevant elements.

image

image

frontend/src/components/About.vue
```diff
@@ -17,35 +22,56 @@
...
-          <center>The SysBio group 2019</center>
+          <center><a href="https://sysbio.se" target="_blank">The SysBio group 2019</a></center>
           <br><br>
-          <h4 class="title is-4">Scientific Advisors</h4>
-          <nav id="advisors" class="level">
-            <div v-for="advisor in advisors" :key="advisor.name" class="level-left">
-              <a :href="advisor.url" target="_blank">
-                <span class="level-item">
+          <h4 class="title is-4">Contact</h4>
+          <p>For any questions, comments or suggestions please email us at <a href="mailto:">contact [at] metabolicatlas [dot] org</a>.</p>
```
e0, line 44: Since we are using Cloudflare now, we could start including the email addresses in the mailto link. Not a big difference either way, just wanted to mention it if we would be interested in the future. https://support.cloudflare.com/hc/en-us/articles/200170016-What-is-Email-Address-Obfuscation-

e0 (approved):
Looks great! Left some minor comments but they are not necessary to fix imo.

Another minor question since I couldn't leave it on the file. How come the `orcid` image is a gif?

.gitignore
mihai-sysbio: I don't have this file, nor is it part of the repository as far as I can tell. Maybe it's not needed?

frontend/src/components/About.vue
mihai-sysbio: Great idea! This is now achieved through 8d70a05.

mihai-sysbio:
> Another minor question since I couldn't leave it on the file. How come the `orcid` image is a gif?

I chose `gif` because of the slightly smaller size; `png` was also available.

.gitignore
e0: Ah ok, I thought it was needed for swagger but it's probably just some leftover file from building the project before.

e0:
> > Another minor question since I couldn't leave it on the file. How come the `orcid` image is a gif?
> 
> I chose `gif` because of the slightly smaller size; `png` was also available. Does this have any unwanted side-effects?

Ah I see. I'm pretty sure it's fine. Was just wondering because I associated `gif` with images with multiple frames but I don't think it would cause any problems here.

frontend/src/components/Home.vue
mihai-sysbio: By the way, I didn't notice this problem in Firefox, so I had to change to Chrome to trigger it.

PR #662, opened 2021-07-26
chore: update dependencies

mihai-sysbio:
This PR aims to resolve security issues reported by GitHub's dependabot. Instead of updating specific dependencies, all dependencies have been updated.

### DIFF
--- a/frontend/src/components/explorer/mapViewer/DataOverlay.vue
+++ b/frontend/src/components/explorer/mapViewer/DataOverlay.vue
@@ -173,10 +173,10 @@ export default {
       return this.HPATissues.length !== 0 && this.tissue2 !== 'None';
     },
     isSelectedCustomtissue1() {
-      return !this.disabledCustomSelectData && NOFILELOADED !== 'None' && NOFILELOADED !== this.customTissue1;
+      return !this.disabledCustomSelectData && ![NOFILELOADED, 'None'].includes(this.customTissue1);
     },
     isSelectedCustomtissue2() {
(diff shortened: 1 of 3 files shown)


e0 (approved):

frontend/src/components/explorer/mapViewer/DataOverlay.vue
```diff
@@ -173,10 +173,10 @@ export default {
       return this.HPATissues.length !== 0 && this.tissue2 !== 'None';
     },
     isSelectedCustomtissue1() {
-      return !this.disabledCustomSelectData && !(NOFILELOADED, 'None').includes(this.customTissue1);
+      return !this.disabledCustomSelectData && NOFILELOADED !== 'None' && NOFILELOADED !== this.customTissue1;
```
e0, line 176: Was this the original intended syntax?

```javascript
      return !this.disabledCustomSelectData && ![NOFILELOADED, 'None'].includes(this.customTissue1);
```

mihai-sysbio: I suspect so. In this particular case, I find the use of parentheses very confusing. Would you suggest to use square brackets instead?

e0: I prefer the square brackets since it's shorter and I find it easier to read but it's fine with your implementation too. I mostly wanted to mention it because I'm pretty sure the original syntax with parenthesis is actually wrong and maybe it was meant to be square brackets from the beginning.

mihai-sysbio: Makes sense! The square brackets have been adopted via 5a6d6d7.

PR #663, opened 2021-07-26
refactor: show chebi image only if exists with a single request

mihai-sysbio:
The original version created two requests in order to verify that a ChEBI image exists https://github.com/MetabolicAtlas/MetabolicAtlas/pull/589.

In this PR the refactoring is focused on triggering a single request by using the technique described in [point 2 here](https://blog.imagekit.io/how-to-handle-loading-images-that-may-not-exist-on-your-website-92e6c3c6ea63) which also resolves bugs such as mentioned in https://github.com/MetabolicAtlas/private-issues/issues/117. While it has been reported that `onerror` is deprecated for `img`, it is still a valid HTML attribute, and as commented in various forums, it is still okay to use. Since the request to ChEBI was done prior to requesting the related metabolites, the website might also appear to load faster by having this request handled in parallel by the browser.

As in the original PR, here are the test cases:

>Fix so no broken ChEBI images are shown
With image
http://localhost/explore/Human-GEM/gem-browser/metabolite/MAM03362p
With broken image
http://localhost/explore/Human-GEM/gem-browser/metabolite/MAM02576c
With no ChEBI entry at all
http://localhost/explore/Human-GEM/gem-browser/metabolite/MAM02865c

### DIFF
--- a/frontend/src/components/explorer/gemBrowser/Metabolite.vue
+++ b/frontend/src/components/explorer/gemBrowser/Metabolite.vue
@@ -53,12 +53,16 @@
       </table>
     </template>
     <template v-slot:chebi>
-      <div v-if="chebiImageLink"
+      <div v-if="metabolite.externalDbs && metabolite.externalDbs.ChEBI"
            class="column is-3-widescreen is-2-desktop is-full-tablet has-text-centered px-2">
         <a :href="metabolite.externalDbs.ChEBI[0].url" target="_blank">
-          <img id="chebi-img" :src="chebiImageLink" class="hoverable" />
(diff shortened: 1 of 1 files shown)


e0 (approved):
This seems to work nicely!

I remember that @inghylt and I initially explored using `onerror` but thought that it wouldn't work since the link returns a successful `200` response. (For example: https://www.ebi.ac.uk/chebi/displayImage.do?defaultImage=true&imageIndex=0&chebiId=17119&dimensions=400) Seems like browsers still deem `200` with no body in the response as an error for images.

PR #665, opened 2021-07-27
fix: update the selected model for search when the global model is updated

e0:
This closes https://github.com/MetabolicAtlas/private-issues/issues/17

### Steps to verify
1. Visit the landing page (`/`).
2. Click on the search icon in the menubar.
3. Verify that the selected model for search is `Fruitfly-GEM`, which is the first one in the list of models.
4. Dismiss the search by clicking outside the search bar.
5. Click on the `GEM Browser` link (below the hero block), which leads to the `/explore/Human-GEM/gem-browser` page.
6. Click on the search icon in the menubar.
7. Verify that the selected model for search is now `Human-GEM` instead of staying as `Fruitily-GEM`.


mihai-sysbio (approved):
Nice implementation, works perfectly 🌟

PR #666, opened 2021-07-27
fix: dispose 3D viewer when it's not in use

### LINKED ISSUE #664: unmount 3D viewer when not in use
As a user, I want to the 3D viewer to be unmounted when it is not in use, so that it does not take up computing resources unnecessarily.

It seems that once the 3D viewer is loaded for a component, it does not unmount when changing to another page (unless the new page is reloaded). In the attached video, see the browser console, which keeps logging out 3D viewer related activity after the gem browser page is loaded.

https://user-images.githubusercontent.com/423498/126984248-4beba0b2-a1e4-4e4b-8f09-b9c60ef0f307.mov

e0:
This closes #664 

This was a bit of a tricky one. Had to do a deep dive in [how to dispose things in three.js](https://threejs.org/docs/#manual/en/introduction/How-to-dispose-of-objects). If I understand correctly, without manually disposing things as [now implemented](https://github.com/MetabolicAtlas/3d-network-viewer/blob/main/src/met-atlas-viewer.js#L1063) in the 3d-network-viewer library, there is no automatic garbage collection for resources used in three.js and this could potentially lead to memory leak in the RAM and GPU. Hope the fixes here covered the problem and can bring much better performance to the users.

### DIFF
--- a/frontend/src/components/explorer/mapViewer/ThreeDviewer.vue
+++ b/frontend/src/components/explorer/mapViewer/ThreeDviewer.vue
@@ -123,7 +123,6 @@ export default {
     },
     async renderNetwork(customizedNetwork) {
       this.resetNetwork();
-
       this.controller = MetAtlasViewer('viewer3d');
 
       const graphData = customizedNetwork || this.network;
--- a/frontend/src/components/explorer/mapViewer/ThreeDviewer.vue
+++ b/frontend/src/components/explorer/mapViewer/ThreeDviewer.vue
@@ -56,6 +56,7 @@ export default {
       controller: null,
       isFullscreen: false,
       searchedNodesOnMap: [],
+      currentLevels: {},
     };
   },
   computed: {
@@ -74,7 +75,6 @@ export default {
   },
   watch: {
     async currentMap() {
-      this.resetNetwork();
       await this.loadNetwork();
     },
     dataOverlayPanelVisible() {
@@ -88,6 +88,9 @@ export default {
   },
   async mounted() {
     await this.loadNetwork();
+  },
+  beforeDestroy() {
+    this.resetNetwork();
   },
   methods: {
     async loadNetwork() {
@@ -120,6 +123,7 @@ export default {
     },
     async renderNetwork(customizedNetwork) {
       this.resetNetwork();
+
       this.controller = MetAtlasViewer('viewer3d');
 
       const graphData = customizedNetwork || this.network;
@@ -177,6 +181,13 @@ export default {
       }
     },
     async applyColorsAndRenderNetwork(levels) {
+      if (this.currentLevels === levels
+          || (this.currentLevels === {} && Object.keys(levels).length > 0)
+      ) {
+        return;
+      }
+      this.currentLevels = levels;
+
       const nodes = this.network.nodes.map((node) => {
         let color = colorToRGBArray('#9df');
 
@@ -219,8 +230,10 @@ export default {
       this.$store.dispatch('maps/setCoords', payload);
     },
     resetNetwork() {
-      const viewer = document.getElementById('viewer3d');
-      viewer.innerHTML = '';
+      if (this.controller) {
+        this.controller.dispose();
+        this.controller = null;
+      }
     },
     zoomIn() {
       this.zoomBy(50);
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -19,7 +19,7 @@
     "cytoscape-cola": "2.3.0",
     "file-saver": "^2.0.2",
     "jquery": "3.3.1",
-    "@metabolicatlas/3d-network-viewer": "^0.1.17",
+    "@metabolicatlas/3d-network-viewer": "^0.1.18",
     "@panzoom/panzoom": "^4.3.2",
     "vue": "^2.6.10",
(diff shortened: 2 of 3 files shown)


frontend/src/components/explorer/mapViewer/ThreeDviewer.vue
```diff
@@ -219,8 +230,10 @@ export default {
       this.$store.dispatch('maps/setCoords', payload);
     },
     resetNetwork() {
-      const viewer = document.getElementById('viewer3d');
-      viewer.innerHTML = '';
+      if (this.controller) {
+        this.controller.dispose();
+        this.controller = null;
```
mihai-sysbio, line 235: Isn't setting to `null` redundant?

frontend/src/components/explorer/mapViewer/ThreeDviewer.vue
```diff
@@ -120,6 +123,7 @@ export default {
     },
     async renderNetwork(customizedNetwork) {
       this.resetNetwork();
+
```
mihai-sysbio, line 126: This extra line here was probably not needed.

frontend/src/components/explorer/mapViewer/ThreeDviewer.vue
```diff
@@ -177,6 +181,13 @@ export default {
       }
     },
     async applyColorsAndRenderNetwork(levels) {
+      if (this.currentLevels === levels
+          || (this.currentLevels === {} && Object.keys(levels).length > 0)
+      ) {
+        return;
+      }
+      this.currentLevels = levels;
+
```
mihai-sysbio, line 190: I have a feeling this might change in the updated data overlay, so it doesn't matter so much how things are done now as long as they work.

mihai-sysbio (changes requested):
Could you please have a look if you also see a weird behavior with the labels on the 3D viewer? Here's a screenshot of the labels being off-center and quite "deep" in the view:

image

frontend/src/components/explorer/mapViewer/ThreeDviewer.vue
e0: This probably has something to do with the [`getInverse` -> `invert` changes](https://github.com/MetabolicAtlas/3d-network-viewer/pull/33/files#diff-35463b24cc26f87cd73e22540e83a61576bb2d35b6226a88461608713f4eb477L506-R506 ). Need to take a second look, thanks!

frontend/src/components/explorer/mapViewer/ThreeDviewer.vue
e0: This was added because the [`EventBus`](https://github.com/MetabolicAtlas/MetabolicAtlas/blob/fix/dispose-3d-viewer/frontend/src/components/explorer/mapViewer/ThreeDviewer.vue#L87) was triggering `applyColorsAndRenderNetwork` too many times, which means the entire 3D viewer was being constructed multiple times. This takes up a lot of resources unnecessarily and it would also also make the disposing of things more cumbersome if we don't add this.

frontend/src/components/explorer/mapViewer/ThreeDviewer.vue
e0: Similar to the problem with the `EventBus` triggering too many times, it can also lead to the `resetNetwork` being called too many times. Setting it to `null`, as well as the initial `if` condition is to prevent this problem. I understand that it's not the best solution long-term but it solves the problem with disposing the viewer. Like you mentioned it is likely that the `EventBus` would get removed completely during the data overlay change so I hope this is ok for now.

e0: > This probably has something to do with the [`getInverse` -> `invert` changes](https://github.com/MetabolicAtlas/3d-network-viewer/pull/33/files#diff-35463b24cc26f87cd73e22540e83a61576bb2d35b6226a88461608713f4eb477L506-R506). Need to take a second look, thanks!

I have made a [PR](https://github.com/MetabolicAtlas/3d-network-viewer/pull/36) to address this problem. Once it's merged and the package is published we can update this repo.

mihai-sysbio: Sounds good. It might be useful to link this discussion too in the refactoring issue.

e0: > > This probably has something to do with the [`getInverse` -> `invert` changes](https://github.com/MetabolicAtlas/3d-network-viewer/pull/33/files#diff-35463b24cc26f87cd73e22540e83a61576bb2d35b6226a88461608713f4eb477L506-R506). Need to take a second look, thanks!
> 
> I have made a [PR](https://github.com/MetabolicAtlas/3d-network-viewer/pull/36) to address this problem. Once it's merged and the package is published we can update this repo.

Addressed in https://github.com/MetabolicAtlas/MetabolicAtlas/pull/666/commits/e1064fe8dec9aacef1711c1cecf232c54b6ab3c1

mihai-sysbio (approved):

mihai-sysbio:
Nicely done!

PR #669, opened 2021-08-05
feat: show model in error message for component id errors

mihai-sysbio:
As a user, I want to easily understand why an ID is not found, so that I find out how to look for it, or where the problem is.

At the moment, visiting an url with a wrong identifier, for example [this](https://metabolicatlas.org/explore/Human-GEM/gem-browser/reaction/m00524), shows:

image

With the changes in this PR, it will show:

image

This will help clarify that the ID was not found in the selected model, which can be different than what the user intended to select.


nanjiangshu (approved):
Looks great! I think the message is more informative with the addition of GEM name.

e0 (approved):
Looks good!

PR #671, opened 2021-08-11
feat: add API endpoints for data overlay sources

e0:
This is related to MetabolicAtlas/private-issues#113.

After building and starting the project, the new endpoints can be reached at (examples):

- Index metadata for all data sources of a model: `/api/data-overlay/Human-GEM`.
- Data source file: `/api/data-overlay/Human-GEM/transcriptomics/protein1.mock.tsv`

After https://github.com/MetabolicAtlas/data-files/pull/14, https://github.com/MetabolicAtlas/neo4j-data-generation/pull/20, as well as this PR are approved and merged, MetabolicAtlas/private-issues#113 can be closed.

### DIFF
--- a/api/src/endpoints/dataOverlay.js
+++ b/api/src/endpoints/dataOverlay.js
@@ -1,0 +1,35 @@
+import express from 'express';
+import { readFileSync } from 'fs';
+
+const routes = express.Router();
+
+routes.get('/:model', async (req, res) => {
+  const { model } = req.params;
+  try {
+    const indexJson = readFileSync(
+      `/project/dataOverlay/${model}/index.json`,
+      'utf8',
+    );
+    res.json(JSON.parse(indexJson));
+  } catch (e) {
+    console.error(e.message);
+    res.sendStatus(404);
+  }
+});
+
+routes.get('/:model/:dataType/:filename', async (req, res) => {
+  const { model, dataType, filename } = req.params;
+  try {
+    const dataSourceFile = readFileSync(
+      `/project/dataOverlay/${model}/${dataType}/${filename}`,
+      'utf8',
+    );
+    res.setHeader('Content-Type', 'text/tsv');
+    res.send(dataSourceFile);
+  } catch (e) {
+    console.error(e.message);
+    res.sendStatus(404);
+  }
+});
+
+export default routes;
--- a/proj.sh
+++ b/proj.sh
@@ -6,7 +6,8 @@ function generate-data {
   echo 'Data generation started.'
   source .env && yarn --cwd $DATA_GENERATOR_PATH start $DATA_FILES_PATH "$@"
   /bin/cp -rf $DATA_GENERATOR_PATH/data/* neo4j/import
-  /bin/cp  -f $DATA_GENERATOR_PATH/data/hpaRna.json api/src/data/
+  /bin/cp  -f $DATA_GENERATOR_PATH/data/hpaRna.json api/src/data/ # this line should be removed after HPA data is included in the dataOverlay folder
+  /bin/cp -rf $DATA_GENERATOR_PATH/data/dataOverlay api/
   /bin/cp  -f $DATA_FILES_PATH/integrated-models/integratedModels.json api/src/data/
   /bin/cp  -f $DATA_FILES_PATH/gemsRepository.json api/src/data/
   /bin/cp -rf $DATA_FILES_PATH/svg api/
--- a/.gitignore
+++ b/.gitignore
@@ -86,3 +86,6 @@ svg/
 
 # data folder in API
 api/src/data/*.json
+
+# dataOverlay folder in API
+api/dataOverlay
--- a/api/src/endpoints/index.js
+++ b/api/src/endpoints/index.js
@@ -2,6 +2,7 @@ import express from 'express';
 import neo4jRoutes from 'endpoints/neo4j';
 import repoRoutes from 'endpoints/repository';
 import hpaRoutes from 'endpoints/hpaRna';
+import dataOverlayRoutes from 'endpoints/dataOverlay';
 import swaggerRoutes from 'endpoints/swagger';
 
 const router = express.Router();
@@ -9,6 +10,7 @@ const router = express.Router();
 router.use(neo4jRoutes);
 router.use('/repository', repoRoutes);
 router.use('/rna', hpaRoutes);
+router.use('/data-overlay', dataOverlayRoutes);
 router.use(swaggerRoutes);
 
 export default router;
--- a/docker-compose-local.yml
+++ b/docker-compose-local.yml
@@ -43,6 +43,7 @@ services:
       - ./api/svg:/project/svg
       - ./api/repository:/project/repository
       - ./api/public:/project/public
+      - ./api/dataOverlay:/project/dataOverlay
     command: yarn develop
 
   ftp:


mihai-sysbio:
> About the [CodeQL warnings](https://github.com/MetabolicAtlas/MetabolicAtlas/security/code-scanning/12?query=ref%3Arefs%2Fpull%2F671%2Fmerge+ref%3Arefs%2Fpull%2F671%2Fhead+ref%3Arefs%2Fheads%2Ffeat%2Fapi-data-overlay-sources), I think it's ok to ignore them since we are caching all of the requests so there should be some automatic rate-limiting. Alternatively we could add a middleware according to [this recommendation](https://codeql.github.com/codeql-query-help/javascript/js-missing-rate-limiting/#recommendation). Thoughts?

Rate limit is already implemented in Nginx based on url, below is the example for the `/api`:
https://github.com/MetabolicAtlas/MetabolicAtlas/blob/ecba3f35ddc1acbfeaa8679a600a875f642db6b2/nginx/prod.nginx.conf#L69
Based on the consensus, I dismissed 2 of the warnings (can be re-opened if we change our minds).

However, the 2 remaining warnings are about requesting data about a model which may not exist. I think it would be okay to let these cases fail if the target model folders do not exist on disk, but I am wondering if we do not already have a function that checks if a requested model exists?

e0:
> However, the 2 remaining warnings are about requesting data about a model which may not exist. I think it would be okay to let these cases fail if the target model folders do not exist on disk, but I am wondering if we do not already have a function that checks if a requested model exists?

Thanks, I didn't notice those actually. Thought it was all the same kind of error. We have the [model checking in frontend](https://github.com/MetabolicAtlas/MetabolicAtlas/blob/develop/frontend/src/store/modules/models.js#L37) but I'm not sure if we currently have something like that in the api. It wouldn't be difficult to implement it though and I think it's a good suggestion.

api/src/endpoints/dataOverlay.js
```diff
@@ -0,0 +1,35 @@
+import express from 'express';
+import { readFileSync } from 'fs';
+
+const routes = express.Router();
+
+routes.get('/:model', async (req, res) => {
+  const { model } = req.params;
+  try {
+    const indexJson = readFileSync(
+      `/project/dataOverlay/${model}/index.json`,
```
mihai-sysbio, line 10: Would it be possible to avoid hard-coding the `/project` folder name, e.g. by using a relative path?

api/src/endpoints/dataOverlay.js
```diff
@@ -0,0 +1,35 @@
...
+    console.error(e.message);
+    res.sendStatus(404);
+  }
+});
+
+routes.get('/:model/:dataType/:filename', async (req, res) => {
+  const { model, dataType, filename } = req.params;
+  try {
+    const dataSourceFile = readFileSync(
+      `/project/dataOverlay/${model}/${dataType}/${filename}`,
```
mihai-sysbio, line 24: Same path-related comment here.

proj.sh
```diff
@@ -6,7 +6,8 @@ function generate-data {
   echo 'Data generation started.'
   source .env && yarn --cwd $DATA_GENERATOR_PATH start $DATA_FILES_PATH "$@"
   /bin/cp -rf $DATA_GENERATOR_PATH/data/* neo4j/import
-  /bin/cp  -f $DATA_GENERATOR_PATH/data/hpaRna.json api/src/data/
+  /bin/cp  -f $DATA_GENERATOR_PATH/data/hpaRna.json api/src/data/ # this line should be removed after HPA data is included in the dataOverlay folder
```
mihai-sysbio, line 9: It would make more sense to me if this comment would be a checklist item in https://github.com/MetabolicAtlas/private-issues/issues/124 . 

My feeling is that to-dos and to-do style comments in code are incompatible with planning practices (including scrum). In line with this thinking, I had created an [issue to clean up to-dos](https://github.com/MetabolicAtlas/private-issues/issues/34).

proj.sh
```diff
@@ -6,7 +6,8 @@ function generate-data {
...
+  /bin/cp -rf $DATA_GENERATOR_PATH/data/dataOverlay api/
```
mihai-sysbio, line 10: I have made a comment about this in `neo4j-data-generation`, suggesting:

```suggestion
  /bin/cp -rf $DATA_GENERATOR_PATH/dataOverlay api/
```

mihai-sysbio (approved):
Great stuff!
I have only minor comments, the outcome of which will not affect the merging, so I am approving the PR. However, I would appreciate if we could go through the comments before merging.

mihai-sysbio:
>  I'm not sure if we currently have something like that in the api. It wouldn't be difficult to implement it though and I think it's a good suggestion.

It might be that this type of check should be implemented "globally" in the API, which to me would be a different issue.

api/src/endpoints/dataOverlay.js
e0: The relative path might be a bit difficult since I think they would be different for `local` and `prod`. It should be possible but I'm not sure if the implementation would be less rigid than the current one. I would also consider having an environment variable to replace `/project/`. If we do decide to change anything, [this file](https://github.com/MetabolicAtlas/MetabolicAtlas/blob/develop/api/src/neo4j/queries/map.js#L9) should also be updated.

proj.sh
e0: Sounds good! I have updated the issue description now.

proj.sh
e0: Looks good. Leaving this as pending until the other related discussions and possible changes are resolved.

api/src/endpoints/dataOverlay.js
mihai-sysbio: The name `/project` is decided in a single place:
https://github.com/MetabolicAtlas/MetabolicAtlas/blob/ecba3f35ddc1acbfeaa8679a600a875f642db6b2/api/Dockerfile#L3
and so both the `local` and `prod` docker-compose files use the same folder name.

e0: If we want to use a relative path, I think `local` is in `src/endpoints` and `prod` is in `dist`.

mihai-sysbio: > If we want to use a relative path, I think `local` is in `src/endpoints` and `prod` is in `dist`.

😕  sorry I don't get it

e0: Maybe I got confused before. Did you mention the `WORKDIR` to use in the absolute file path? If so then it makes sense but if it's still about the relative path I'm not sure how it helps.

mihai-sysbio: The suggestion below seems to work:

```suggestion
      `./dataOverlay/${model}/index.json`,
```

api/src/endpoints/dataOverlay.js
mihai-sysbio: Here is the same suggestion as above:
```suggestion
      `./dataOverlay/${model}/${dataType}/${filename}`,
```

api/src/endpoints/dataOverlay.js
e0: Ahh right, in this case it does work since the file reading happens from the place of execution `api`. I thought it would be like when we do a `require` or `import` like `../../dataOverlay` for `local` and `../dataOverlay` for `prod`.

e0:
> Nice work! Tried with the renaming changes and seems to work fine 👍

Thanks for testing it so fast! Nice to have this confirmed to wrap up the work for today. 😊

nanjiangshu (approved):
I struggled a bit with the API for `dataOverlay` since I just got `Not Found` error when trying to access the API point Shan @e0  mentioned in the description.  Finally, I figured out it was because the files in `dataOverlay` were not all copied due to permission change when mounted by Docker. I guess this is just an annoying feature for Docker on Linux. 

It works great and all setups looks good to me! 

One thing I do not fully understand is why do we create an API for accessing the dataOverlay. Is that only for the convenience of developers or will MetAtlas server also retrieve data of `dataOverlay` via the API? I guess it is for the latter.

e0:
> I struggled a bit with the API for `dataOverlay` since I just got `Not Found` error when trying to access the API point Shan @e0 mentioned in the description. Finally, I figured out it was because the files in `dataOverlay` were not all copied due to permission change when mounted by Docker. I guess this is just an annoying feature for Docker on Linux.
> 

Sorry about this. I think @inghylt had this same problem yesterday. We should do a better job at either the import script or at least mention this in the documentation.

---

> 
> One thing I do not fully understand is why do we create an API for accessing the dataOverlay. Is that only for the convenience of developers or will MetAtlas server also retrieve data of `dataOverlay` via the API? I guess it is for the latter.

Do you mean why not have these files directly in the frontend? If so I think that could increase the bundle size by a lot. In addition, we also want to serve the `tsv` files and then do all the processing (validation and possible transformation) in the frontend because we want to use the same processing logic for handling custom data the users can upload.

nanjiangshu:
> > I struggled a bit with the API for `dataOverlay` since I just got `Not Found` error when trying to access the API point Shan @e0 mentioned in the description. Finally, I figured out it was because the files in `dataOverlay` were not all copied due to permission change when mounted by Docker. I guess this is just an annoying feature for Docker on Linux.
> 
> Sorry about this. I think @inghylt had this same problem yesterday. We should do a better job at either the import script or at least mention this in the documentation.
> 
> > One thing I do not fully understand is why do we create an API for accessing the dataOverlay. Is that only for the convenience of developers or will MetAtlas server also retrieve data of `dataOverlay` via the API? I guess it is for the latter.
> 
> Do you mean why not have these files directly in the frontend? If so I think that could increase the bundle size by a lot. In addition, we also want to serve the `tsv` files and then do all the processing (validation and possible transformation) in the frontend because we want to use the same processing logic for handling custom data the users can upload.

Great explanation! I agree that it is good to keep the data processing logic consistent. 
Do you know how will the custom upload data be stored? Will it be stored temporary in the docker volume or will the file exist even after the stack is redeployed?

e0:
> Do you know how will the custom upload data be stored? Will it be stored temporary in the docker volume or will the file exist even after the stack is redeployed?

Since all the data processing should be done in the frontend the custom data is meant to be used directly inside the browser. Since there is no user login we can't store different custom data for specific users. If someone wishes to share their data and make it available for all users, they could reach out to the us maintainers and then we can make a PR to `data-files` to add this new data source.

nanjiangshu:
> > Do you know how will the custom upload data be stored? Will it be stored temporary in the docker volume or will the file exist even after the stack is redeployed?
> 
> Since all the data processing should be done in the frontend the custom data is meant to be used directly inside the browser. Since there is no user login we can't store different custom data for specific users. If someone wishes to share their data and make it available for all users, they could reach out to the us maintainers and then we can make a PR to `data-files` to add this new data source.

I see. I think it should work well for now. If in the future custom data is frequently used, it might be good to store the data (e.g. under a randomly created ID) so that the users can share the 2D map with dataOvderlay by just sharing the link, without the need to upload the data again by people who was shared.

e0:
> I see. I think it should work well for now. If in the future custom data is frequently used, it might be good to store the data (e.g. under a randomly created ID) so that the users can share the 2D map with dataOvderlay by just sharing the link, without the need to upload the data again by people who was shared.

That makes sense. We need to make sure we comply with GDPR for storing this type of data server side as well as and think about how best to store the files so they are persistent after redeploys. Sounds like a good candidate for a future issue.

mihai-sysbio:
Sharing of custom data overlay will be achievable through the notebook-based advanced data overlay, which allows us to avoid storing any data.

Issue #605, opened 2021-06-01
testing


inghylt:
CII Best Practice:
- _The project MUST use at least one automated test suite that is publicly released as FLOSS (this test suite may be maintained as a separate FLOSS project). The project MUST clearly show or document how to run the test suite(s) (e.g., via a continuous integration (CI) script or via documentation in files such as BUILD.md, README.md, or CONTRIBUTING.md)_
- _A test suite SHOULD be invocable in a standard way for that language_
- _It is SUGGESTED that the test suite cover most (or ideally all) the code branches, input fields, and functionality_
- _It is SUGGESTED that the project implement continuous integration (where new or changed code is frequently integrated into a central code repository and automated tests are run on the result)_
- _The project MUST have a general policy (formal or not) that as major new functionality is added to the software produced by the project, tests of that functionality should be added to an automated test suite._
- _The project MUST have evidence that the test_policy for adding tests has been adhered to in the most recent major changes to the software produced by the project._
- _It is SUGGESTED that this policy on adding tests (see test_policy) be documented in the instructions for change proposals._

As a developer, I would like to have test validating my code, so that I know that changes I make don't create bugs that are missed during PR reviews

As a user, I would like to have test validating the code, so that I know that the software I use is functional and that I use valid data

mihai-sysbio:
The testing framework `cypress` is implemented - a first test in written at `frontend/cypress/testing`. To run the tests see the dev docs

Potentially use  https://chrome.google.com/webstore/detail/cypress-recorder/glcapdcacdfkokcmicllhcjigeodacab

mihai-sysbio:
@e0 mentioned these frameworks in https://github.com/MetabolicAtlas/MetabolicAtlas/pull/673#issuecomment-908275439 :

* https://github.com/lukeed/uvu
* https://jestjs.io

inghylt:
Issues
#849
#1082
#1083

are actionable issues that cover what we have left to fulfill, closing this epic

PR #673, opened 2021-08-18
feat: validate uploaded data overlay files

inghylt:
This is an MVP of a component that accepts an uploaded file, and checks that
- the uploaded file is a TSV file
- that there are at least two columns and two rows
- that column names have max length of 20 characters
- that id:s have max length of 20 characters
- that non id column values are between 0-1

Small test files were created as well to test the functionality.

As the DataOverlay.vue component will be redesigned, as well as the RNA expression logic, the component is not incorporated into any existing components in this PR.

To test the new component, import the component into DataOverlay.vue and add:

`<DataOverlayValidation @errorCustomFile="handleErrorCustomFile($event)" />`

And try out the different test files. The results are available in the console.


mihai-sysbio (changes requested):
This looks great,
As is, I'm unsure what should be done next to integrate this into the codebase. I'm expecting either an adjustment of this PR, or a link to another issue that describes the integration. In the same line, I was kind of expecting a visual indicator of success/failure as in the current implementation, and another way of seeing the status instead of the console.

inghylt:
Yes, I didn't develop it any further, including the success/failure indicator, as I did not know how the main data overlay component would look like. Maybe keep this PR in draft mode and wait until the main data overlay one is done?

e0:
> And try out the different test files. The results are available in the console.

Looks good! I think you included all the essential test cases and great that you included the test files as well. Just one minor comment, I had to add `console.log(errorMsg);` in the `handleErrorCustomFile` function in `DataOverlay.vue` component to see the results in the console. Not sure if I imported it wrong but just wanted to leave this here in case anyone else was wondering.

@inghylt We also previously discussed a bit about adding unit tests for these. Here are two good options:
- https://github.com/lukeed/uvu
- https://jestjs.io

mihai-sysbio:
As commented in https://github.com/MetabolicAtlas/private-issues/issues/115#issuecomment-908995401, the testing does not need to be part of this PR. There is also a dedicated issue for setting up (automated? unit/e2e?) testing in https://github.com/MetabolicAtlas/MetabolicAtlas/issues/605.

inghylt:
Commit 58c01deb310246728de312d00833b1b14cd16995 provides error messaging based on our discussion previous today.

I'm thinking, maybe the validation code should just be inside DataOverlay.vue even though that one is quite big already? Feels like DataOverlatValidation.vue is kind of small + creates a bit less clean code. Or maybe keep it like this for now and incorporate into the changes made through #676 at a later stage

frontend/src/components/explorer/mapViewer/DataOverlayValidation.vue
```diff
@@ -0,0 +1,63 @@
...
+</template>
+
+<script>
+export default {
+  name: 'DataOverlayValidation',
+  methods: {
+    validateFile(e) {
+      if (e.target.files.length !== 0) {
+        if (e.target.files[0].type !== 'text/tab-separated-values') {
+          this.$emit('errorCustomFile', 'Error: expected TSV file', e.target.files[0].name);
```
mihai-sysbio, line 24:
```suggestion
          this.$emit('errorCustomFile', 'Error: expected a TSV file', e.target.files[0].name);
```

mihai-sysbio (commented):
This is looking great!

I did notice some minor things that you might want to think about:
- [x] the error messages could go in `messages.js`
- [ ] when there are multiple errors, it would be more readable with a line break between them
- [ ] the error text should be justified instead of centered
- [ ] the left-right spacing around the box with the file name does not appear symmetrical
- [ ] the dismiss button `x` has a tendency to break on the next line like this:
image

#676 also contains some validation code that should be merged, so I would favor keeping the DataOverlay component separate. Plus, as you say @inghylt, this can always be done later.

frontend/src/components/explorer/mapViewer/DataOverlayValidation.vue
```diff
@@ -0,0 +1,63 @@
...
+              }
+            });
+            lines.slice(1).forEach((line) => {
+              const columnValues = line.split('\t');
+              if (columnValues[0].length > 20) {
+                errors += `Error: Id:s should have max length of 20 characters, ${columnValues[0]} has ${columnValues[0].length}\n`;
+              }
+              columnValues.slice(1).forEach((value) => {
+                if (Number.isNaN(Number(value)) || value < 0 || value > 1) {
+                  errors += `Error: ${value} is not a number between 0-1\n`;
```
nanjiangshu, line 45: I think it is better to use `between 0 and 1`

nanjiangshu (changes requested):
It looks great @inghylt! I agree with the [minor points](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/673#pullrequestreview-747611074) mentioned by @mihai-sysbio, probably the unevenness of the space around the box of input file name is not a big issue. 

One scenario that might be missing is when the input file does not contain the Id column, e.g. 

```
gland 1	gland 2
0.1	0.8
0.22	0.7
```
It could be good to validate such data files as well.

MalinAhlberg:
I have added a test for the scenario mentioned by Nanjiang (https://github.com/MetabolicAtlas/MetabolicAtlas/pull/673#pullrequestreview-748355818).

I also noted this bug(?): if you open an invalid file, close the error message and then reopen the same file, you do not get the error message back. If you do the same for the valid file, you do get the success message again.

inghylt:
@MalinAhlberg commit 19488681acca0ac170e76353e73befafc7b945c3 fixes the bug you discovered! The onchange event was not triggered in Chromium when opening the same file twice but setting the value of the input element to `''` onclick seems to work

inghylt:
image

@mihai-sysbio using justified instead of centered makes the spacing weird as seen on the image. But I think commit 5ed0907dda4e7c558a3d9bd0b5a05af9d7d64a4a makes the text layout work nicely

inghylt:
Commit 1092bdf07ad473ba99ad63d508044b5cef612170 makes sure that the custom file upload error message does not take up too much space and is scrollable if a lot of errors are present. The scrolling setting seems to be ignored when using mobile view with Chromium, but I would say this is acceptable given that people will probably not upload their custom data file on a phone

frontend/src/components/explorer/mapViewer/DataOverlay.vue
```diff
@@ -338,16 +327,25 @@ export default {
...
       overflow: hidden;
-      max-width: 250px;
       text-overflow: ellipsis;
-      cursor: help;
     }
   }
 }
+
+#customFileError {
+  max-height:30%;
```
mihai-sysbio, line 344:
```suggestion
  max-height: 30%;
```

frontend/src/components/explorer/mapViewer/DataOverlayValidation.vue
```diff
@@ -0,0 +1,70 @@
+<template>
+  <div class="file is-centered mb-2">
+    <label class="file-label">
+      <input id="anId" class="file-input" type="file" name="resume" @change="validateFile"
```
mihai-sysbio, line 4: I'm not sure if this id is needed?

```suggestion
      <input class="file-input" type="file" name="resume" @change="validateFile"
```

mihai-sysbio (approved):
🦄

frontend/src/components/explorer/mapViewer/DataOverlayValidation.vue
inghylt: Good catch. I'm thinking it makes sense to remove name="resume" as well?

mihai-sysbio: Yes, it looks like there is no other reference to `resume` in the codebase.

frontend/src/components/explorer/mapViewer/DataOverlayValidation.vue
```diff
@@ -15,13 +15,15 @@
...
 
 <script>
+import { default as messages } from '@/content/messages';
+
 export default {
   name: 'DataOverlayValidation',
   methods: {
     validateFile(e) {
       if (e.target.files.length !== 0) {
         if (e.target.files[0].type !== 'text/tab-separated-values') {
```
nanjiangshu, line 26: It seems fileType checking is not working now. When I try to upload a non-TSV file, either a textfile or binary file, the error message is not shown.

inghylt: Ah good catch! I'll look at it :+1:

inghylt: The bug is fixed in commit 35878b1f76a52469954cf891d64acb535c17d45a

nanjiangshu (approved):
Looks great now!

mihai-sysbio (approved):

frontend/src/components/explorer/mapViewer/Svgmap.vue
```diff
@@ -330,10 +331,12 @@ export default {
     },
     applyLevelsOnMap() {
       if (Object.keys(this.computedLevels).length === 0) {
-        $(`#svg-wrapper .${this.componentClassName} .shape`).attr('fill', this.componentDefaultColor);
+        Object.values(DATA_TYPES_COMPONENTS).forEach((dataType) => {
```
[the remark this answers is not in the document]
mihai-sysbio: It might be because of this comment I left on Slack:
> A user should be able to overlay multiple data types at the same time, so they should not be reset.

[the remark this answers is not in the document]
inghylt: Before PR #676, this section looked like this

```
    applyHPARNAlevelsOnMap(RNAlevels) {
      this.HPARNAlevels = RNAlevels;
      if (Object.keys(this.HPARNAlevels).length === 0) {
        $('#svg-wrapper .enz .shape').attr('fill', this.defaultGeneColor);
```

I think @e0 adapted it to make in applicable to more data sources. However, when switching between sources/setting data set to None, the coloring was not reset entirely on the map. The code you commented fixes this. Please play around in the data overlay section if you want to see how it looks like with and without this fix

PR #676, opened 2021-08-25
feat: frontend multiple data overlay

e0:
This PR covers refactoring the data overlay component and related logic, connecting the routes with data overlay options and renaming/removing now outdated properties and methods.

Overlaying multiple data sources will be handed in issue MetabolicAtlas/private-issues#134 to keep this PR more contained, and data validation in `frontend/src/helpers/dataOverlay.js` will be addressed in #673

For testing, please use the `biotin_metabolism` map from `Human-GEM` and the `fructose_and_mannose_metabolism` map from `Yeast-GEM` in the map viewer

Keeping the previous draft description from @e0 below for reference

_Creating this draft PR as I have a few questions. I have some ideas for how to solve them, but I would appreciate some discussion and input before deciding on the implementation._

_For testing, please use the `biotin_metabolism` map from `Human-GEM` and the `fructose_and_mannose_metabolism` map from `Yeast-GEM` in the map viewer._

_Note: _Currently, this does not include the proposed design of having data types in tabs. The functionality is included though using a select.__

_### Questions_

_1. For custom data, the uploaded data can be of different types (currently gene or metabolite). How should this type be set?
2. For the map viewer's route params, e.g. `?dim=2d&panel=1&sel=&search=&coords=-1566,-276,0.3,2.7,-52,138&g1=None`, there are more params (data type, data source, dataset name) to be considered now. In the future if more data sources can be added for comparison, that could increase the amount of params by quite a bit. Should the map route params still include the data overlay configuration or maybe they could be omitted in favor of the proposed Juypter notebook based solution?
3. Currently we have one set of colors `RNALegend`. If we want to go forward with using this color spectrum for all component types, then we should probably just rename this file. But maybe we want to investigate if we should have different color spectrums for different component types?_


frontend/src/helpers/dataOverlay.js
```diff
@@ -0,0 +1,97 @@
...
+    throw new Error('Invalid TSV format, expect at least two columns.');
+  }
+
+  // parse lines
+  // make tissues key;
+  for (let i = 0; i < tissues.length; i += 1) {
+    const tissue = tissues[i];
+    if (tissue in levels) {
+      throw new Error(`Error: duplicated column '${tissue}'.`);
+    }
```
mihai-sysbio, line 59: This is a check/validation, so it might make sense to move it to the validation component.

frontend/src/helpers/dataOverlay.js
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  let entriesCount = 0;
+  for (let k = 0; k < lines.length; k += 1) {
+    const line = lines[k];
+    if (line) {
+      const arrLine = line.split('\t');
+      if (arrLine.length !== tissues.length + 1) {
+        throw new Error(`Error: invalid number of values line ${indexLine}.`);
+      }
```
mihai-sysbio, line 70: Another example of data validation.

frontend/src/helpers/dataOverlay.js
```diff
@@ -0,0 +1,97 @@
...
+      for (let i = 1; i < arrLine.length; i += 1) {
+        if (arrLine[i]) {
+          const v = Number(arrLine[i]);
+          if (Number.isNaN(v)) {
+            throw new Error(`Error: invalid value line ${indexLine}.`);
```
mihai-sysbio, line 75: Again, data validation.

frontend/src/helpers/dataOverlay.js
```diff
@@ -0,0 +1,97 @@
...
+  let indexLine = 1;
+  // fetch Tissue
+  if (lines[0].split('\t').length !== 1) {
+    const arrLine = lines[0].split('\t');
+    const v = Number(arrLine[1]);
+    if (Number.isNaN(v)) {
+      tissues = lines[0].split('\t');
+      tissues.shift();
+      lines.shift();
+    } else {
```
mihai-sysbio, line 43: Would it be possible to have this type of check in the data validation stage?

mihai-sysbio:
Thanks for setting up this draft PR, it does make for an iterative discussion.

The terminology we have so far is: data type, data source and data set. According to my interpretation, the current use is potentially confusing, where the data source selection is missing.

>     1. For custom data, the uploaded data can be of different types (currently gene or metabolite). How should this type be set?

According to the sketched layouts, the data type is set via a vertical accordion or tabs. A straightforward way would be to upload data under each section.

>     2. For the map viewer's route params, e.g. `?dim=2d&panel=1&sel=&search=&coords=-1566,-276,0.3,2.7,-52,138&g1=None`, there are more params (data type, data source, dataset name) to be considered now. In the future if more data sources can be added for comparison, that could increase the amount of params by quite a bit. Should the map route params still include the data overlay configuration or maybe they could be omitted in favor of the proposed Juypter notebook based solution?

All URLs should be copy-pastable, so yes they should include the data overlay configuration.

>     3. Currently we have one set of colors `RNALegend`. If we want to go forward with using this color spectrum for all component types, then we should probably just rename this file. But maybe we want to investigate if we should have different color spectrums for different component types?

Renaming the component makes a lot of sense. In order to keep issues small, and this implementation as close to an MVP as possible, I propose to not investigate color spectrum changes at this stage.

frontend/src/helpers/dataOverlay.js
e0: Great that you noticed these. For reference, these are mostly untouched/slightly adjusted based on [RNAexpression.vue](https://github.com/MetabolicAtlas/MetabolicAtlas/blob/develop/frontend/src/components/explorer/mapViewer/RNAexpression.vue), which has been removed in this PR. Your comments are good to leave up until they can be resolved.

e0:
> The terminology we have so far is: data type, data source and data set. According to my interpretation, the current use is potentially confusing, where the data source selection is missing.

The data source selection should not be missing. Can you please provide a screenshot of what it looks like for you?

---

> 
> > ```
> > 1. For custom data, the uploaded data can be of different types (currently gene or metabolite). How should this type be set?
> > ```
> 
> According to the sketched layouts, the data type is set via a vertical accordion or tabs. A straightforward way would be to upload data under each section.

I was under the impression that the accordion/tabs would be constructed dynamically based on the data types available for a model. But it sounds like you want the same data types to be presented for all models regardless if there are data sources for them? It's a good solution I think, just want to double check. If we do go this route though, the models that don't have the specific data types should only show the custom file select.

inghylt:
Me and @e0 had a small meeting and concluded that

- map viewer route should be updated to contain all info
- all different models should have the same data types options (currently only transcriptomics & metabolomomics, might be more in he future) even though there might not be any data of that type on the server (disable select or similar if no data available)
- important to think of mobile view when changing design
- data fields called tissues in the code today should be renamed to dataSets or similar to be neutral and not HPA specific
- no color spectrum changes at this stage

We are thinking that maybe it is beneficial to keep the data types as a select option, only more separated and higher up in the component than it is today. Tabs might get a bit crowded if more data types are added, and select is easy to use for mobile view as well

mihai-sysbio:
@inghylt great summary

I have something to add to
> We are thinking that maybe it is beneficial to keep the data types as a select option, only more separated and higher up in the component than it is today. Tabs might get a bit crowded if more data types are added, and select is easy to use for mobile view as well

One can expect users to overlay multiple data types at the same time, which I guess wouldn't be as straightforward with a dropdown. At the same it, I can see it would be simpler to implement a dropdown, at least for now.

mihai-sysbio:
@e0 thanks for the follow-up, here are some (hopefully clearer) answers:

> > The terminology we have so far is: data type, data source and data set. According to my interpretation, the current use is potentially confusing, where the data source selection is missing.
> 
> The data source selection should not be missing. Can you please provide a screenshot of what it looks like for you?

You're right, everything looks great! (screenshot below)
image

Not sure what I missed before (but I did restart the stack in between).

> If we do go this route though, the models that don't have the specific data types should only show the custom file select.

Yes! We do want to provide users with the option of overlaying there data, even if we haven't found a data source to integrate ourselves.

nanjiangshu:
> > One can expect users to overlay multiple data types at the same time, which I guess wouldn't be as straightforward with a dropdown. At the same time, I can see it would be simpler to implement a dropdown, at least for now.
> > What I'm trying to say is that if it's not possible to overlay multiple data types at the same time via the current PR, it would be good to have a separate issue for it.
> 
> I think I totally forgot about the detail about overlaying multiple data types at the same time. If that's the case, I'm not sure if having tabs is an improvement over a select. Ideally it would be good if all of the selected data type+source+set can be displayed at the same time.

I think one can select multiple items with select option, see https://www.w3schools.com/tags/att_select_multiple.asp. It might be a bit awkward to operate on mobile devices.

frontend/src/components/explorer/mapViewer/DataOverlay.vue
```diff
@@ -147,6 +147,11 @@ export default {
   },
   async created() {
     await this.getDataSourcesIndex(this.model.short_name);
+    const datatype = this.validDataTypeInQuery() ? this.$route.query.datatype : Object.keys(this.dataSourcesIndex)[0];
```
mihai-sysbio, line 150: Just a thought: I would be tempted to define a variable for `Object.keys(this.dataSourcesIndex)[0]`.

frontend/src/components/explorer/mapViewer/DataOverlay.vue
```diff
@@ -203,6 +208,9 @@ export default {
       this.errorCustomFileMsg = errorMsg;
       this.showFileLoader = false;
     },
+    validDataTypeInQuery() {
+      return this.$route.query.datatype && Object.keys(this.dataSourcesIndex).indexOf(this.$route.query.datatype) > -1;
+    },
```
mihai-sysbio, line 213: Does this also validate the dataset?

inghylt: no, only that type `sometype` in `datatype=sometype` is present at our server

frontend/src/helpers/dataOverlay.js
inghylt: #673 is only focused on validating user uploaded data, while parseFile is used for parsing server and user input files. But come to look at it, since many of the checks in this file are related to the ones created in #673, I'm starting to wonder if they should be moved in here entirely?

inghylt: Since (this PR) https://github.com/MetabolicAtlas/MetabolicAtlas/blob/0529cddd67ef8a3c5972ab0a00d7553b82421a23/frontend/src/components/explorer/mapViewer/DataOverlay.vue#L209 uses `handleErrorCustomFile` which is also used in #673 https://github.com/MetabolicAtlas/MetabolicAtlas/blob/1aeb5155e0b8442374e00e809add79aaa591c5f1/frontend/src/components/explorer/mapViewer/DataOverlay.vue#L17

inghylt: Either way, I'll think I'll leave the validation code in this file for now and work with it more when merging PR #673

inghylt:
What has been fixed since you last looked at this PR

- the route and data overlay component are now connected
- data fields called tissues in the code are renamed to dataSets to be neutral and not HPA specific. The property has been removed entirely in some places where it is no longer needed

Overlaying multiple data sources will be handed in another issue to keep this PR more contained, and comment https://github.com/MetabolicAtlas/MetabolicAtlas/pull/676#discussion_r698327625 will be addressed in #673

frontend/src/components/Home.vue
```diff
@@ -224,7 +224,7 @@ export default {
           img: require('../assets/analyze.jpg'),
           route: { name: 'viewer',
             params: { model: 'Human-GEM', type: 'compartment', map_id: 'endoplasmic_reticulum' },
-            query: { dim: '2d', panel: '1', g1: 'skin 1', coords: '-7222.7,-4501.6,0.97,0,0,0' } },
+            query: { dim: '2d', panel: '1', coords: '-7222.7,-4501.6,0.97,0,0,0' } },
```
nanjiangshu, line 227: Do you know what is this `g1: skin 1` for @mihai-sysbio @inghylt ?

mihai-sysbio: It was the url param `skin 1` for the first tissue `1` for gene expression `g`.

frontend/src/store/modules/dataOverlay.js
```diff
@@ -0,0 +1,139 @@
...
+      const dataSource = await parseFile(file);
+      const metadata = data.index[type].find(
+        m => m.filename === filename
+      );
+      commit('setCurrentDataSource', {
+        ...metadata,
+        ...dataSource,
+      });
+    } catch (e) {
+      console.error(e);
```
nanjiangshu, line 89: I see a warning message ```warning: Unexpected console statement (no-console) at src/store/modules/dataOverlay.js:89:7:``` with console.log() here.

inghylt: Good catch. I have three ideas:

- adding  `// eslint-disable-line no-console` (lazy solution)

- replace the logging with generating a `<ErrorPanel :message="errorMessage" @hideErrorPanel="errorMessage=''" />` in the component, since it is should be more of a server problem that user action related
- remove the logging and settle with having ` commit('setCurrentDataSource', null);`

Any thoughts?

nanjiangshu: As you mentioned, this is more a server problem, so users are probably not expected to see this error message. We just tell them something like "file uploading failed, please try again". This error message is more for the developers. In this case, we could add a DEBUG mode and only when the DEBUG mode is activated, this error message will be shown in the log. I think it should be done in a separate issue if we are going add this feature. For the moment, before we have a good reason to remove this error catching, I would opt for the solution 1.

frontend/src/components/Home.vue
nanjiangshu: Thanks for the explanation, but it seems this is not in use now?

inghylt: Exactly, that is why it was removed! All relevant url params for data overlay are now grouped through `queryParams` in  `frontend/src/store/modules/dataOverlay.js`

frontend/src/helpers/dataOverlay.js
inghylt: Summary from meeting (will be edited)

frontend/src/components/explorer/mapViewer/DataOverlay.vue
mihai-sysbio: As @inghylt mentioned, this is used only once (so I don't really recall why I made this comment). Resolving this now.

frontend/src/store/modules/dataOverlay.js
mihai-sysbio: So far we don't have any other `try catch` in the store, so for simplicity I would continue with this.
Making sure our API responds well is a valid concern, so we should inform the user when that doesn't happen. This is already implemented in other parts of the interface, but not consistently.
My suggestion would be to open up a different issue for this.

frontend/src/helpers/dataOverlay.js
```diff
@@ -0,0 +1,97 @@
...
+/*
+ * @param { Blob } file
+ *
+ * @example
+ * try {
+ *   const data = await parseCustomFile(file);
+ * } catch (e) {
+ *   console.error(e)
+ * }
+ */
```
mihai-sysbio, line 26: I haven't seen this type of comment in other places in our codebase.

frontend/src/helpers/dataOverlay.js
inghylt: All validation will be moved into https://github.com/MetabolicAtlas/MetabolicAtlas/blob/35878b1f76a52469954cf891d64acb535c17d45a/frontend/src/components/explorer/mapViewer/DataOverlayValidation.vue

frontend/src/helpers/dataOverlay.js
inghylt: All validation will be moved into https://github.com/MetabolicAtlas/MetabolicAtlas/blob/35878b1f76a52469954cf891d64acb535c17d45a/frontend/src/components/explorer/mapViewer/DataOverlayValidation.vue

frontend/src/helpers/dataOverlay.js
inghylt: Fixed with commit 703c572b298e278bac79a284a932c68697268188

frontend/src/store/modules/dataOverlay.js
nanjiangshu: Sounds good!

inghylt: Issue #691 created for this and commit 4f9a3ea528c69deb9c1c95987bd3c27594caae42 disables the eslint for now

mihai-sysbio (approved):
This is so great 🤩 

Just a comment if we ever need this (again) in the future: as part of this PR the color scale for comparisons (2 data sources against each other) was removed.

nanjiangshu (approved):
Looks great to me!

inghylt:
This PR will be merged even though there is still mock data in the dataOverlay folders of Human-GEM and Yeast-GE in repository data-files. Valid data will come via https://github.com/MetabolicAtlas/private-issues/issues/124

Issue #677, opened 2021-08-27
API endpoint for 2D map thumbnails


mihai-sysbio:
As a developer for an external resource (website for database/service), I want to show thumbnails of the SVG maps, so that I can enrich the user experience on the website I work on.
Desired: automate the generation of thumbnails (dynamic means less work in future deploys).
Optional: size of the thumbnail as a parameter (px).

Any feedback from HPA will result in a new issue.

nanjiangshu:
Based on a search at npmjs.com with the keyword `thumbnail svg`, [sharp](https://www.npmjs.com/package/sharp) is one of the few javascript packages that support conversion of SVG to web-friendly image format such as png. Moreover, sharp is the most actively maintained, with 110 releases and the latest release just 20 days ago. In addition, resizing of images for sharp is typically 4-5 times faster than using the quickest mageMagick and GraphicsMagick settings due to its use of libvips. Therefore, I think `sharp` is a good solution for dynamic creation of thumbnails from the 2D SVG maps.

mihai-sysbio:
Sounds excellent @nanjiangshu!

PR #685, opened 2021-09-06
API for 2D SVG map thumbnails

nanjiangshu:
This PR adds the API endpoint for 2D map thumbnails.

Description, an thumbnail in PNG format is returned by accessing the API endpoint for a specified 2D map. 

To test:
    /api/v2/svg-thumbnail/acylglycerides_metabolism?model=Human-GEM

You will need to rebuild the stack in order to test this pull request since a new package `sharp` has been added to the package list. 

The thumbnail image is generated by the package [sharp](https://www.npmjs.com/package/sharp) in real time. 
To be discussed: 
1. Size of the thumbnail image. Currently, all thumbnails are with a fixed width of 400px. (Resolved)
    Width of the thumbnail can be set by the parameter `width` in the query, although it should be between 100 and 600.
2. Format of the return value of the API endpoint. Currently, the return value is a png image. (Resolved)
    PNG is a proper format for thumbnail.
3. In the converted thumbnail png images, characters are all displayed as squares since fonts such as`Arial` are not available in the Docker container. To install these fonts may increase the size of the container significantly, not to say some of the fonts are not free according to https://wiki.alpinelinux.org/wiki/Fonts. For thumbnails, when the width is around 400, letters can not be seen, so probably we can ignore this.

### DIFF
--- a/api/src/endpoints/index.js
+++ b/api/src/endpoints/index.js
@@ -3,6 +3,7 @@ import neo4jRoutes from 'endpoints/neo4j';
 import repoRoutes from 'endpoints/repository';
 import hpaRoutes from 'endpoints/hpaRna';
 import dataOverlayRoutes from 'endpoints/dataOverlay';
+import svgRoutes from 'endpoints/svgThumbnail';
 import swaggerRoutes from 'endpoints/swagger';
 
 const router = express.Router();
@@ -11,6 +12,7 @@ router.use(neo4jRoutes);
 router.use('/repository', repoRoutes);
 router.use('/rna', hpaRoutes);
 router.use('/data-overlay', dataOverlayRoutes);
+router.use('/svg-thumbnail', svgRoutes);
 router.use(swaggerRoutes);
 
 export default router;
--- a/api/src/endpoints/index.js
+++ b/api/src/endpoints/index.js
@@ -3,7 +3,7 @@ import neo4jRoutes from 'endpoints/neo4j';
 import repoRoutes from 'endpoints/repository';
 import hpaRoutes from 'endpoints/hpaRna';
 import dataOverlayRoutes from 'endpoints/dataOverlay';
-import svgRoutes from 'endpoints/svgThumbnail';
+import svgThumbRoutes from 'endpoints/svgThumbnail';
 import swaggerRoutes from 'endpoints/swagger';
 
 const router = express.Router();
@@ -12,7 +12,7 @@ router.use(neo4jRoutes);
 router.use('/repository', repoRoutes);
 router.use('/rna', hpaRoutes);
 router.use('/data-overlay', dataOverlayRoutes);
-router.use('/svg-thumbnail', svgRoutes);
+router.use('/svg-thumbnail', svgThumbRoutes);
 router.use(swaggerRoutes);
 
 export default router;
--- a/api/src/endpoints/svgThumbnail.js
+++ b/api/src/endpoints/svgThumbnail.js
@@ -20,7 +20,7 @@ routes.get('/:svgName', async (req, res) => {
     var w_thumb = 400;
     const isnum = /^\d+$/.test(width);
     if (!isnum) {
-      throw new Error(`Invalid width provided: ${width}. It should be an integer.`);
+      throw new Error(`Invalid width provided: ${width}. It should be an integer between ${min_w} and ${max_w}.`);
     } else {
       w_thumb =  parseInt(width) || 400;
       if (w_thumb < min_w || w_thumb > max_w){
--- a/api/src/endpoints/svgThumbnail.js
+++ b/api/src/endpoints/svgThumbnail.js
@@ -10,7 +10,22 @@ routes.get('/:svgName', async (req, res) => {
   const { model } = req.query;
 
   try {
-    const svgThumbnail = await getSvgThumbnail(svgName, model)
+    const min_w = 100;
+    const max_w = 600;
+    var w_thumb = 400;
+    const isnum = /^\d+$/.test(req.query.width);
+    if (!isnum) {
+      res.status(400).send('Error! The value of width should an integer, but \''
+        + req.query.width + '\' is given!' );
+    } else {
+      w_thumb =  parseInt(req.query.width) || 400;
+      if (w_thumb < min_w || w_thumb > max_w){
+        res.status(400).send('Error! The value of width should be between '
+          + min_w + ' and ' + max_w
+          + ' but \'' + req.query.width + '\' is given!' );
+      }
+    }
+    const svgThumbnail = await getSvgThumbnail(svgName, w_thumb, model)
     res.type('png')
     res.send(svgThumbnail);
   } catch (e) {
--- a/api/src/endpoints/svgThumbnail.js
+++ b/api/src/endpoints/svgThumbnail.js
@@ -7,22 +7,22 @@ const sharp = require('sharp');
 
 routes.get('/:svgName', async (req, res) => {
   const { svgName } = req.params;
-  const { model } = req.query;
+  const { model, version, width } = req.query;
 
   try {
     const min_w = 100;
     const max_w = 600;
     var w_thumb = 400;
-    const isnum = /^\d+$/.test(req.query.width);
+    const isnum = /^\d+$/.test(width);
     if (!isnum) {
       res.status(400).send('Error! The value of width should an integer, but \''
-        + req.query.width + '\' is given!' );
+        + width + '\' is given!' );
     } else {
-      w_thumb =  parseInt(req.query.width) || 400;
+      w_thumb =  parseInt(width) || 400;
       if (w_thumb < min_w || w_thumb > max_w){
         res.status(400).send('Error! The value of width should be between '
           + min_w + ' and ' + max_w
-          + ' but \'' + req.query.width + '\' is given!' );
+          + ' but \'' + width + '\' is given!' );
       }
     }
     const svgThumbnail = await getSvgThumbnail(svgName, w_thumb, model)
--- a/api/src/endpoints/svgThumbnail.js
+++ b/api/src/endpoints/svgThumbnail.js
@@ -4,7 +4,6 @@ import getSvgThumbnail from 'utils/2d-map';
 
 const routes = express.Router();
 const sharp = require('sharp');
-let HPA_JSON = null;
 
 routes.get('/:svgName', async (req, res) => {
   const { svgName } = req.params;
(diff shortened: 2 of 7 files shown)


api/src/endpoints/svgThumbnail.js
```diff
@@ -0,0 +1,25 @@
...
+import fs from 'fs';
+import getSvgThumbnail from 'utils/2d-map';
+
+const routes = express.Router();
+const sharp = require('sharp');
+let HPA_JSON = null;
+
+routes.get('/:svgName', async (req, res) => {
+  const { svgName } = req.params;
+  const { model } = req.query;
```
mihai-sysbio, line 11: I think a way to check that the query model is valid is already implemented in the API, so it would be good to make use of that.

api/src/endpoints/svgThumbnail.js
```diff
@@ -0,0 +1,25 @@
...
+
+  try {
+    const svgThumbnail = await getSvgThumbnail(svgName, model)
+    res.send(svgThumbnail);
+  } catch (e) {
+    if (e.message === '404') {
+      return res.sendStatus(404);
+    }
+
+    res.status(400).send(e.message);
```
mihai-sysbio, line 21: There is a CodeQL warning here.

api/src/utils/2d-map.js
```diff
@@ -0,0 +1,23 @@
...
+    .resize(w_thumb)
+    .flatten({ background: { r: 255, g: 255, b: 255 } })
+    .png()
+    .toBuffer()
+    .then(function(info) {
+      console.log(info)
+      return info
+    })
+    .catch(function(err) {
+      console.log(err)
```
mihai-sysbio, line 18: We will likely want to remove these `console.log` before merging, so I'm just making a note of them.

api/src/endpoints/svgThumbnail.js
```diff
@@ -0,0 +1,25 @@
+import express from 'express';
+import fs from 'fs';
+import getSvgThumbnail from 'utils/2d-map';
+
+const routes = express.Router();
+const sharp = require('sharp');
+let HPA_JSON = null;
```
mihai-sysbio, line 7: I'm not spotting where the `HPA_JSON` variable is used.

mihai-sysbio (changes requested):
This looks good overall. I'm happy to see that the thumbnails can be generated automatically, as desired in #677.

>     1. Size of the thumbnail image. Currently, all thumbnails are with a fixed width of 400px.

It seems relatively straightforward to parametrize the image width (say between 100 and 600), so I think we should do it. It is likely this will come in very handy when we want to use these dynamic thumbnails in our website - at some point, we were talking about an overview screen for the Map Viewer, and the lack of thumbnails was a small barrier.

In line with the second bullet point: there might be something wrong with the file type, or maybe it's just the lack of a file extension, based on the screenshot below:
image

api/src/endpoints/svgThumbnail.js
nanjiangshu: Thanks for pointing out this. This comment has been addressed in the commit #db510182.
The version of the model is also parsed when handling the query parameters with the commit #a6046c95

api/src/utils/2d-map.js
nanjiangshu: These lines are removed with the commit b6399bb7

api/src/endpoints/svgThumbnail.js
nanjiangshu: It has been removed with the commit b6399bb7

api/src/endpoints/svgThumbnail.js
nanjiangshu: I don't see any warning message in the panel `Checks`. I guess the problem is solved now.

nanjiangshu: This has been used at multiple places for the API endpoints. Shall we do a refactoring? But if so I guess it is better to be done in a new issue. 
This vulnerability is medium, which is caused by potential sending user input data from `e.message`.  On solution might be that we filter all alphanumeric characters from `e.message`.

api/src/endpoints/svgThumbnail.js
```diff
@@ -0,0 +1,44 @@
...
+    if (!VALID_MODELS.includes(model)) {
+      throw new Error(`Invalid model provided: ${model}.`);
+    }
+
+    const min_w = 100;
+    const max_w = 600;
+    var w_thumb = 400;
+    const isnum = /^\d+$/.test(width);
+    if (!isnum) {
+      throw new Error(`Invalid width provided: ${width}. It should be an integer.`);
```
inghylt, line 25:
```suggestion
      throw new Error(`Invalid width provided: ${width}. It should be an integer between ${min_w} and ${max_w}.`);
```

api/src/endpoints/svgThumbnail.js
```diff
@@ -0,0 +1,44 @@
+import express from 'express';
+import fs from 'fs';
```
inghylt, line 2: unused module?

api/src/endpoints/svgThumbnail.js
```diff
@@ -0,0 +1,44 @@
...
+import getSvgThumbnail from 'utils/2d-map';
+import models from 'data/integratedModels.json';
+
+const VALID_MODELS = models.map((m) => m.short_name);
+
+const routes = express.Router();
+const sharp = require('sharp');
```
inghylt, line 9: unused module?

api/src/utils/2d-map.js
```diff
@@ -0,0 +1,20 @@
+import fs from 'fs';
```
inghylt, line 1: unused module?

api/src/endpoints/index.js
```diff
@@ -3,6 +3,7 @@ import neo4jRoutes from 'endpoints/neo4j';
 import repoRoutes from 'endpoints/repository';
 import hpaRoutes from 'endpoints/hpaRna';
 import dataOverlayRoutes from 'endpoints/dataOverlay';
+import svgRoutes from 'endpoints/svgThumbnail';
```
inghylt, line 6: minor thing, but I would prob name this svgThumbnail instead of svgRoutes as the filename is svgThumbnail and it only relates to that

nanjiangshu: I just don't like very long variable name unless necessary. Maybe I can change it to `svgThumbRoutes`.

api/src/endpoints/svgThumbnail.js
nanjiangshu: My bad, forgot to remove it after refactoring.

api/src/endpoints/svgThumbnail.js
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    const min_w = 100;
+    const max_w = 600;
+    var w_thumb = 400;
+    const isnum = /^\d+$/.test(width);
+    if (!isnum) {
+      throw new Error(`Invalid width provided: ${width}. It should be an integer.`);
+    } else {
+      w_thumb =  parseInt(width) || 400;
```
inghylt, line 27: Is the intention of this line to have width set as 400 if no width query param is provided? The current behavior is that `Invalid width provided: undefined. It should be an integer.` if you don't provide a width query param

api/src/endpoints/svgThumbnail.js
nanjiangshu: removed

api/src/utils/2d-map.js
nanjiangshu: removed

api/src/endpoints/svgThumbnail.js
nanjiangshu: removed

inghylt (commented):
Nice work! :+1: 

Do we want to have some kind of custom error message regarding when asking for a map that's not part of a model? This is the resulting message for both

http://localhost/api/v2/svg-thumbnail/acylglycerides_metabolism?model=Fruitfly-GEM&width=600 (map not present in the model) and http://localhost/api/v2/svg-thumbnail/acylglyceridez_metabolism?model=Human-GEM&width=400 (misspelled map)
image

Obv. this is not something everyday users of the site will experience but I felt that error message is a bit misleading. But I'm also fine in letting this be as it is

api/src/endpoints/svgThumbnail.js
```diff
@@ -0,0 +1,44 @@
...
+import models from 'data/integratedModels.json';
+
+const VALID_MODELS = models.map((m) => m.short_name);
+
+const routes = express.Router();
+const sharp = require('sharp');
+
+routes.get('/:svgName', async (req, res) => {
+  const { svgName } = req.params;
+  const { model, version, width } = req.query;
```
inghylt, line 11: const version is unused?

api/src/endpoints/svgThumbnail.js
nanjiangshu: It can be removed after I added `isnum` checking.

api/src/endpoints/svgThumbnail.js
nanjiangshu: no. But it is suggested by Mihail to add it. I guess it might be used in the future.

nanjiangshu:
> be

I think it is a very good point. I have added an exception to check whether the 2D map exists or not.

api/src/endpoints/svgThumbnail.js
mihai-sysbio: > This has been used at multiple places for the API endpoints. Shall we do a refactoring? But if so I guess it is better to be done in a new issue.

Absolutely! Marking this as resolved here.

api/src/endpoints/svgThumbnail.js
mihai-sysbio: This param is needed in the case when we update the model version and we don't have the older one in the website.

mihai-sysbio (approved):
looks good to me!

api/src/endpoints/svgThumbnail.js
nanjiangshu: If we always keep one version of 2D maps on the website, then we don't need the version parameter even if we updated the model version. It is needed only we are trying to keep 2D maps of multiple model versions. In addition, for the 2D map viewer, it is important that the model version in the svg file, i.e. `data-modelversion`, matches the model version in the yaml file. However, for this thumbnail API, it doesn't matter. It just create a PNG thumbnail of the svg file and don't check the model version in it. Does it make sense @mihai-sysbio @inghylt ?

mihai-sysbio: I see. Fine by me as long as there is a note of this in https://github.com/MetabolicAtlas/private-issues/issues/39.

inghylt (approved):

PR #687, opened 2021-09-13
feat: incorporate feedback for the identifiers page

### LINKED ISSUE #637: incorporate feedback for the identifier page
After getting feedback on #575, the following items should be implemented:

- [x] the component name at the beginning of each bullet list item should be removed
- [x] the components should be grouped by model (eg in sections)
- [ ] ~for the Metabolite ids, use the metabolite name~
- [x] link from the `External databases` section

nanjiangshu:
This PR closes #637 
Some points of the issue may still need to be clarified, see comment in the issue #637 

Linking from the section External databases in the Gem Browser to the identifiers page has been implemented at the commit c36b1d8c, but not sure if it is the desired feature. 
For some external pages, e.g. `/identifier/HMR%202.0/m00357c`, there is not url to the real external web-page, I think it might be better to mention that the external webpage for HMR2.0 does not exist.

### DIFF
--- a/frontend/src/components/ExternalDb.vue
+++ b/frontend/src/components/ExternalDb.vue
@@ -8,17 +8,23 @@
         This database identifier is associated with the following Metabolic Atlas
         {{ components.length === 1 ? 'component' : 'components' }}:
       </p>
+
       <ul class="is-flex-direction-column is-align-items-flex-start mb-4 ml-5">
-        <li v-for="c in components" :key="c.id + c.model + c.version" class="my-1">
-          <span class="tag is-light is-medium">
-            <router-link :to="{ name: c.componentType.toLowerCase(), params: { model:
-              c.model, id: c.id } }">
-              {{ c.id }}
-            </router-link>
-          </span>
-          from {{ c.model }} {{ c.version }}
+        <li v-for="(components, model) in compGroupedByModel" :key="model" class="my-1">
+          {{ model }}
+          <ul class="is-flex-direction-column is-align-items-flex-start mb-4 ml-5">
+            <li v-for="c in components" :key="c.id + c.model + c.version" class="my-1">
+              <span class="tag is-light is-medium">
+                <router-link :to="{ name: c.componentType.toLowerCase(), params: { model:
+                  c.model, id: c.id } }">
+                  {{ c.id }}
+                </router-link>
+              </span>
+            </li>
+          </ul>
         </li>
       </ul>
+
       <p v-if="externalDb.url">
         For more details, visit <a :href="externalDb.url" target="_blank"> {{ externalDb.url }} </a>.
       </p>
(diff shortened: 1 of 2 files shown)


frontend/src/components/ExternalDb.vue
```diff
@@ -37,6 +42,21 @@ export default {
       components: state => state.externalDb.components,
       externalDb: state => state.externalDb.externalDb,
     }),
+    compGroupedByModel() {
+      const result = this.components.reduce((r, a) => { // eslint-disable-next-line
+        r[a.model] = r[a.model] || [];
+        r[a.model].push(a);
+        return r;
+      }, Object.create(null));
```
MalinAhlberg, line 50: Question: why do you use a null object here, rather than just `{}`?

nanjiangshu: `Object.create(null)` creates an object that doesn't inherit anything. But you are right, here the object to be create is not a map, so it is more neat to just use `{}`. I will change it.

inghylt (approved):
Nice work! I think the linking from the section External databases in the Gem Browser to the identifiers page looks good. I'm not sure if information regarding non existing external web pages, e..g. for HMR 2.0, is needed, I think it makes sense as it is now

frontend/src/components/ExternalDb.vue
MalinAhlberg: Ok, thanks for the explanation! I just got curious :)

mihai-sysbio (approved):
Looks great.
There is a discussion in the issue about introducing the metabolite names next to their IDs, but I'll review again if that implementation is needed.
The only suggestion I have is about moving the link to `identifiers.org` somewhere higher up, as it feels too out of sight now.

nanjiangshu:
> Looks great.
> There is a discussion in the issue about introducing the metabolite names next to their IDs, but I'll review again if that implementation is needed.
> The only suggestion I have is about moving the link to `identifiers.org` somewhere higher up, as it feels too out of sight now. This was not part of the original issue, so I'm not _requesting_ changes, but would appreciate if it can be considered.

Very good point! I also think the link to the external database is someone hidden when the list is long. I've moved it on top of the component list when it exists with the commit c67a00ac

I'll close this pull request if there is no more comments.

mihai-sysbio:
> I'll close this pull request if there is no more comments.

super!

frontend/src/components/ExternalDb.vue
```diff
@@ -7,22 +7,25 @@
       <p class="my-3">
         This database identifier is associated with the following Metabolic Atlas
         {{ components.length === 1 ? 'component' : 'components' }}:
+        <template v-if="externalDb.url">
+          <br>(please visit <a :href="externalDb.url" target="_blank">{{ externalDb.url }}</a> for more detailes)
```
inghylt, line 11:
```suggestion
          <br>(visit <a :href="externalDb.url" target="_blank">{{ externalDb.url }}</a> for more detailes)
```

PR #692, opened 2021-09-21
fix: work break for long names in MapsAvailable.vue

### LINKED ISSUE #675: add word break for MapsAvailable.vue
When the maps have very long names, such as in https://metabolicatlas.org/explore/Human-GEM/gem-browser/reaction/MAR08387, the 3d button does not fit.

inghylt:
This PR closes #675, i.e. fixing map names taking up too much space if very long.

Some examples links to test:

/explore/Human-GEM/gem-browser/reaction/MAR08387
/explore/Human-GEM/gem-browser/metabolite/MAM03556m
/explore/Human-GEM/gem-browser/gene/ENSG00000088179

But please play around with the gem browser as well and different screen sizes to see if you can find weird cases I've missed

### DIFF
--- a/frontend/src/App.vue
+++ b/frontend/src/App.vue
@@ -435,4 +435,8 @@ span.sc {
   white-space: nowrap;
 }
 
+.break-word {
+  word-break: break-word;
+}
+
 </style>
--- a/frontend/src/App.vue
+++ b/frontend/src/App.vue
@@ -437,6 +437,10 @@ span.sc {
 
 .break-word {
   word-break: break-word;
+  -webkit-hyphens: auto;
+  -moz-hyphens: auto;
+  -ms-hyphens: auto;
+  hyphens: auto;
 }
 
 </style>
--- a/frontend/src/App.vue
+++ b/frontend/src/App.vue
@@ -438,8 +438,6 @@ span.sc {
 .break-word {
   word-break: break-word;
   -webkit-hyphens: auto;
-  -moz-hyphens: auto;
-  -ms-hyphens: auto;
   hyphens: auto;
 }
 
--- a/frontend/src/components/explorer/gemBrowser/MapsAvailable.vue
+++ b/frontend/src/components/explorer/gemBrowser/MapsAvailable.vue
@@ -10,7 +10,7 @@
       <table class="table maps-table">
         <tbody class="has-text-left">
           <tr v-for="component in mapsAvailable" :key="component.id">
-            <td> {{ component.customName }} </td>
+            <td style="word-break: break-word;"> {{ component.customName }} </td>
             <td v-if="component.svgMaps.length===0"> </td>
             <td v-else-if="component.svgMaps.length===1">
               <button class="button is-outlined is-small link-button"
--- a/frontend/src/components/explorer/gemBrowser/MapsAvailable.vue
+++ b/frontend/src/components/explorer/gemBrowser/MapsAvailable.vue
@@ -10,7 +10,7 @@
       <table class="table maps-table">
         <tbody class="has-text-left">
           <tr v-for="component in mapsAvailable" :key="component.id">
-            <td style="word-break: break-word;"> {{ component.customName }} </td>
+            <td class="break-word"> {{ component.customName }} </td>
             <td v-if="component.svgMaps.length===0"> </td>
             <td v-else-if="component.svgMaps.length===1">
               <button class="button is-outlined is-small link-button"
--- a/frontend/src/layouts/explorer/gemBrowser/ComponentLayout.vue
+++ b/frontend/src/layouts/explorer/gemBrowser/ComponentLayout.vue
@@ -30,7 +30,7 @@
               <a href="/api/v2" target="_blank">API</a></p>
           </div>
           <slot v-if="isMetabolite" name="chebi" />
-          <div class="column is-3-widescreen is-3-desktop is-half-tablet has-text-centered">
+          <div class="column is-3-widescreen is-3-desktop is-half-tablet has-text-centered" style="min-width: 350px;">
             <router-link v-if="interactionPartner" class="button is-info is-fullwidth is-outlined"
                          :to="{
                            name: 'interaction',
--- a/frontend/src/layouts/explorer/gemBrowser/ComponentLayout.vue
+++ b/frontend/src/layouts/explorer/gemBrowser/ComponentLayout.vue
@@ -30,7 +30,7 @@
               <a href="/api/v2" target="_blank">API</a></p>
           </div>
           <slot v-if="isMetabolite" name="chebi" />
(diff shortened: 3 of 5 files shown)


frontend/src/components/explorer/gemBrowser/MapsAvailable.vue
```diff
@@ -10,7 +10,7 @@
       <table class="table maps-table">
         <tbody class="has-text-left">
           <tr v-for="component in mapsAvailable" :key="component.id">
-            <td> {{ component.customName }} </td>
+            <td style="word-break: break-word;"> {{ component.customName }} </td>
```
mihai-sysbio, line 13: The code base contains two occurrences of 
> word-wrap: break-word;

It this a chance to create a class (if there is no Bulma equivalent)?

nanjiangshu: This is the inbuilt syntax as described in https://developer.mozilla.org/en-US/docs/Web/CSS/word-break. However, I think it would be better to add a hyphen (maybe refer to https://css-tricks.com/almanac/properties/h/hyphenate/) after the break point. For example, in this screenshot Screenshot 2021-09-22 at 08 46 51, it might be confusing for the word `Glycosylphosphatidylinositol` without adding a hyphen. For English literature, it is usually expressed as 
```
Glycosylphosph-
atidylinositol
```
when breaking of the word is inevitable. 

Another thing is when there is a single letter in another line after word breaking, e.g. 

Screenshot 2021-09-22 at 08 47 48
I'm not sure if there is any easy solution to avoid such cases. If not, I guess adding hyphen rule would be enough.

frontend/src/layouts/explorer/gemBrowser/ComponentLayout.vue
```diff
@@ -30,7 +30,7 @@
               <a href="/api/v2" target="_blank">API</a></p>
           </div>
           <slot v-if="isMetabolite" name="chebi" />
-          <div class="column is-3-widescreen is-3-desktop is-half-tablet has-text-centered">
+          <div class="column is-3-widescreen is-3-desktop is-half-tablet has-text-centered" style="min-width: 350px;">
```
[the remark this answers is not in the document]
inghylt: Good idea. Commit 0cf2139b7f27ac5b734017e34b220c10b08c1590 incorporates this. but I used is-6-tablet to avoid names with more standard length(example Endoplasmic reticulum or Mitochondria) to get like 1-2 letters broken off when screen is a bit more narrow than tablet

frontend/src/components/explorer/gemBrowser/MapsAvailable.vue
inghylt: `word-break` and `word-wrap` are two different CSS properties with different effects. But I could create a class for `word-break: break-word;` in this PR if you like and create a separate issue for the `word-wrap: break-word;` cases

mihai-sysbio: Even if the CSS rules are different, the aim is the same: fit long words in tight spaces.
It might be that two solutions are needed, which would explain the use of different rules.
It might also be that we come up with a more generalized rule.

inghylt: Sorry @nanjiangshu didn't see your comment before! Commit 0cf2139b7f27ac5b734017e34b220c10b08c1590 should fix the single letter issue

Regarding the hyphen, I thinks it's a good idea! I've tried it out but I can't make it work in a good way while not having the map names overflowing. We could perhaps insert `&hyphen; or &shy; ` in long map names and use `hyphens: manual`. But I feel like that might be over engineering things?

inghylt: Commit bdada9bae56c81e9801c543c38a13a4a6bd47400 includes a generalized rule for word-break. Please check it out and see if you think it looks good

frontend/src/layouts/explorer/gemBrowser/ComponentLayout.vue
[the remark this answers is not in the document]
mihai-sysbio: Good thinking!

frontend/src/components/explorer/gemBrowser/MapsAvailable.vue
mihai-sysbio: > But I feel like that might be over engineering things?

I feel the same, especially that we would then need to be careful when using these hyphenated string in eg queries.

<br>
Commit bdada9b is neat!

mihai-sysbio (approved):
🦌

frontend/src/components/explorer/gemBrowser/MapsAvailable.vue
nanjiangshu: Hi @inghylt  It seems single letter issue still remains but I think we can ignore it since there is no easy solution.
For the hyphenation, I agree that we should not over engineering by insert &hyphen. However, I've tested with the following css 

```css
.break-word {
  word-break: break-word;
  -webkit-hyphens: auto;
  -moz-hyphens: auto;
  -ms-hyphens: auto;
  hyphens: auto;
}
``` 

And the hyphenation works at least for the web-browser `Safari`.  I would suggest we add it. Also note that ` word-break: break-word;` may not work on IE for windows users according to https://developer.mozilla.org/en-US/docs/Web/CSS/word-break.

inghylt:
@mihai-sysbio could you have a look at c3ffa3dfc19dfe40a49a875243ca5dfe4bd52004. Me and @nanjiangshu had a brief meeting looking at this. Hyphens are treated differently for each browser. It works well in Safari, i.e. all long names get hyphens.  Firefox hyphens some words, but not all. So in `/explore/Human-GEM/gem-browser/reaction/MAR08387`, `Glycosylphosphatidylinositol (GPI)-anchor biosynthesis`  does not get a hyphen but `Endoplasmic reticulum` do. Chromium ignores it completely :shrug:

Irrelevant of browser, I think it looks a bit cleaner without hyphens than with, but I'm also up for having them.

frontend/src/App.vue
```diff
@@ -435,4 +435,12 @@ span.sc {
   white-space: nowrap;
 }
 
+.break-word {
+  word-break: break-word;
+  -webkit-hyphens: auto;
+  -moz-hyphens: auto;
+  -ms-hyphens: auto;
```
mihai-sysbio, line 442: These lines didn't seem to be picked up by the target browsers (Firefox, Edge), could you have another look?
```suggestion
```

mihai-sysbio (approved):
This is a bit of an inconsistent behaviour, which I would normally prefer to avoid just because of that. If hyphenation would be consistent, I think I would prefer it for improved readability.
Given that is already implemented, I'm okay with leaving it in. Let's not veer off too much from the scope of the PR.

MalinAhlberg (approved):

nanjiangshu (approved):

PR #694, opened 2021-09-22
fix: make handleCallback methods use the new and not old id

### LINKED ISSUE #680: related reactions not refreshing
Originally reported [here](https://github.com/MetabolicAtlas/MetabolicAtlas/issues/646#issuecomment-909426341) by @JonathanRob (tagging him to keep him in the loop).

As a user, I want to see the accurate content of the model reflected on the GEM browser pages.

When navigating through consecutive Reaction pages, the `Related reactions` row from the first table does not update (stays the same as in the first Reaction page displayed).

Note: Previously the entire page has not been refreshed and it has been solved. Now only the related reaction ID is not refreshed.

inghylt:
This PR closes #680

That the related reactions was not refreshed was due to that the `handleCallback` method used to get related reactions for a new reaction (i.e, when clicking on the link on a related reaction) was not using the new reaction id. This bug was also occurring for related metabolites. 

`/explore/Human-GEM/gem-browser/reaction/MAR05029` was the reported example of a misbehaving reaction, but please play around in `/explore/Human-GEM/gem-browser` and choose some random reactions/metabolites to see that their related reactions/metabolites are functioning.

### DIFF
--- a/frontend/src/components/explorer/gemBrowser/Metabolite.vue
+++ b/frontend/src/components/explorer/gemBrowser/Metabolite.vue
@@ -130,9 +130,9 @@ export default {
     };
   },
   methods: {
-    async handleCallback() {
+    async handleCallback(model, id) {
       try {
-        const payload = { model: this.model, id: this.metaboliteId };
+        const payload = { model, id };
         await this.$store.dispatch('metabolites/getRelatedMetabolites', payload);
       } catch {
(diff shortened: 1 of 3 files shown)


mihai-sysbio (approved):
Super clean!

mihai-sysbio:
Sorry for the force-push, that commit wasn't meant for this branch.

nanjiangshu (approved):
Excellent!

PR #695, opened 2021-09-23
fix: make window scroll to coord 0,0 when loading a new 3D map in mobile view

### LINKED ISSUE #678: 3D Map Viewer does not jump to the map region after selecting a map on mobile
As a user, I would like the 3D map viewer will be re-focused at the map region after selecting a map on a mobile device, exactly as for the 2D Map viewer. 

Attached please find a video recording of the described problem. 
https://images.zenhubusercontent.com/53664497/32d49793-ce03-41bb-939e-53ade5ead459/svid_20210827_225805_1.mp4

It should be one or two lines of code change. The same problem has been solved for the 2D map viewers.

inghylt:
This PR closes #678 

Please play around with different maps in http://localhost/explore/Human-GEM/map-viewer/?dim=3d and verify that everything looks good

### DIFF
--- a/frontend/src/components/explorer/mapViewer/MapsListing.vue
+++ b/frontend/src/components/explorer/mapViewer/MapsListing.vue
@@ -33,6 +33,8 @@
 
 <script>
 import { mapState } from 'vuex';
+import mobileScrollToTop from '@/helpers/mapViewer.js';
+
 
 export default {
   name: 'MapsListing',
@@ -48,6 +50,7 @@ export default {
   },
   methods: {
     changeToMap(newMapId) {
+      mobileScrollToTop();
       this.$router.push({ params: { map_id: newMapId }, query: { dim: this.showing2D ? '2d' : '3d' } });
     },
   },
--- a/frontend/src/components/explorer/mapViewer/Svgmap.vue
+++ b/frontend/src/components/explorer/mapViewer/Svgmap.vue
@@ -36,7 +36,6 @@ import MapSearch from '@/components/explorer/mapViewer/MapSearch';
 import { default as messages } from '@/content/messages';
 import { reformatChemicalReactionHTML } from '@/helpers/utils';
 import { DATA_TYPES_COMPONENTS } from '@/helpers/dataOverlay';
-import mobileScrollToTop from '@/helpers/mapViewer.js';
 
 export default {
   name: 'Svgmap',
@@ -258,8 +257,6 @@ export default {
       if (!this.svgContent) {
         return;
       }
-
-      mobileScrollToTop();
 
       this.initialLoadWithParams = !!this.$route.query.coords;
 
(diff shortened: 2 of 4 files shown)


mihai-sysbio (approved):
Nice!
I just want to bring up that these lines are identical in `Svgmap.vue` and `ThreeDviewer.vue`. One could consider creating a helper function, since it's likely we would need it for other components as well.

mihai-sysbio (approved):
:bird:

nanjiangshu (commented):
It looks like exactly the same function `mobileScrollToTop` has been used in Svgmap and ThreeDviewer. However, when I tested on three web-browsers Chrome, Firefox and Safari. 
Jumping for the Svgmap: working on all three browsers
Jumping for the 3D map: not working on Chrome. 
🤔

mihai-sysbio:
@nanjiangshu did you test on real mobile devices?

inghylt:
@nanjiangshu jumping for the 3D map is working for me in Chrome? :thinking:

MalinAhlberg (commented):
Nice work!
Potential issues I found:
- If I scroll back down and select the map that I'm already at (which at least happens to unfocused users like me...), I don't get back to the top of the page.
- I get a module error saying that `vuex import should occur before import of @/helpers/utils at src/components/explorer/gemBrowser/Tile.vue:64`. This might be some local, unwanted code check though, since you guys seem to have no problem with it.

inghylt:
@MalinAhlberg 

_If I scroll back down and select the map that I'm already at (which at least happens to unfocused users like me...), I don't get back to the top of the page._

Interesting! To me this is expected behavior, what do you think @mihai-sysbio @nanjiangshu?

And yes I think the module error is local

nanjiangshu:
Hi @mihai-sysbio and @inghylt, it seems jumping works on Chrome when I just tested. Not sure what was the problem before. Maybe related to the cached data of the Browser.

As to the problem pointed out by @MalinAhlberg, I think it is a good cache. It would be good that jumping happens whenever they click, since it would be difficult for the mobile user to realize they have clicked the map that is shown on the Map Viewer.

mihai-sysbio:
The website behaviour should be consistent, i.e. always clicking on a map name makes the UI scroll to top. But also not a deal breaker for me if sometimes (platform/browser dependent) this doesn't happen. Maybe it's a consequence of bug #697 ?

inghylt:
> Nice work!
> Potential issues I found:
> 
>     * If I scroll back down and select the map that I'm already at (which at least happens to unfocused users like me...), I don't get back to the top of the page.
> 

Commit f5369bd10a62a1ddf0a979dae8fba5acc9622a17 fixes this. It was not a consequence of bug #697, but that bug was discovered when creating the fix :bug:

mihai-sysbio (approved):
🐅  (not quite a cheetah but close enough)

MalinAhlberg (approved):
Looks very good to me :tada:

nanjiangshu (approved):
Great fix!

PR #700, opened 2021-09-29
Feat/include old ids in search

inghylt:
This is a draft PR to get feedback for search should include old ids #118

Please have a look at the search functionality both by GemSearch (i.e the search bar always available at the top of the page which searches wihtin the chosen model) and the global search (when clicking `"search all intergated GEMs"` in the GemSearch or using `http://localhost/search?term=somesearchterm`)
 
##### Old ID:s 
Ids from `HMR 2.0`, e.g. `HMR_4797`, `m01016c`, seems to work fine both in GemSearch and global search. Though a hit for gene is returned now for m01016c in global search which was not the case before

Ids from `Recon3D` are not as neat. `r0760` becomes messy in GemSearch since there are results for not only reactions. Fine in global if you know you are looking for a reaction. Before nothing was returned in GemSearch, and only Yeast-Gem hits in global

`cspg_c` seems to work fine though.

More old ID:s  can be found by choosing a random reaction or metabolite in the GemBrowser and using their external `HMR 2.0` or `Recon3D` id)

##### MA ID:s 
Seems to work fine? Examples: `MAR07523`, `MAM03309s`

Note, if you try to search using parts of the id, e.g. 07523, nothing of value is returned in neither GemSearch or global. Before no results was returned at all.

##### Genes:
Seems to work fine for both? Examples: `POLR3F`, `HSD17B1`

##### Subsystems:
Seems to work fine? Examples `Glycerolipid metabolism` `Fatty acid biosynthesis`. Becomes messy in GemSearch since there are results for not only subsystems but that was the case before as well.

##### Compartments:
Seems to work fine? Examples `Golgi apparatus ` `Inner mitochondria`. Becomes messy in GemSearch since there are results for not only compartments but that was the case before as well.

### Questions:

- Do users use Recon3D or or mostly HMR 2.0?
- Do you think the GemSearch and global search are working as they should?

### DIFF
--- a/api/src/neo4j/queries/search.js
+++ b/api/src/neo4j/queries/search.js
@@ -309,7 +309,7 @@ OPTIONAL MATCH (node)-[${v}]-(parentNode:${model})
 WHERE node:${model} OR parentNode:${model}
 WITH DISTINCT(
 	CASE
-		WHEN EXISTS(node.id) THEN { id: node.id, labels: labelList, score: score }
+		WHEN EXISTS(node.id) AND NOT EXISTS(node.externalId) THEN { id: node.id, labels: labelList, score: score }
 		ELSE { id: parentNode.id, labels: LABELS(parentNode), score: score }
 	END
 ) as r 
@@ -334,7 +334,6 @@ LIMIT ${limit}
   }, {});
 
   const ids = Object.assign({}, ...Object.keys(uniqueIds).map(c => ({ [c]: Array.from(uniqueIds[c]) })));
-
   const [
     compartmentalizedMetabolites,
     metabolites,
--- a/api/src/neo4j/queries/search.js
+++ b/api/src/neo4j/queries/search.js
@@ -324,12 +324,14 @@ LIMIT ${limit}
 
   const results = await queryListResult(statement);
 
+  const idsToScore = {};
   const uniqueIds = results.reduce((o, r) => {
     const c = intersect(componentTypes, r.labels);
     if (!o[c]) {
       o[c] = new Set();
     }
     o[c].add(r.id);
+    idsToScore[r.id] = r.score;
     return o;
   }, {});
 
@@ -350,13 +352,33 @@ LIMIT ${limit}
     fetchCompartments({ ids: ids["Compartment"], model, version: v, includeCounts: true }),
   ]);
 
-  // formatting for simple (gem browser) search
+  const resObj = {compartmentalizedMetabolites,
+    metabolites,
+    genes,
+    reactions,
+    subsystems,
+    compartments}
+    const resWithScore = {};
+    for (const [component, result] of Object.entries(resObj)) {
+        if (result) {
+            resWithScore[component] = result.map(obj => {
+                    const objWithScore = {
+                        ...obj,
+                        score: idsToScore[obj.id]
+                    }
+                    return objWithScore;
+                })
+        } else {
+            resWithScore[component] = []
+        }
+    }
+
   return {
-    metabolite: [...compartmentalizedMetabolites || [], ...metabolites || []],
-    gene: genes || [],
-    reaction: reactions || [],
-    subsystem: subsystems || [],
-    compartment: compartments || [],
+    metabolite: [...resWithScore.compartmentalizedMetabolites, ...resWithScore.metabolites],
+    gene: resWithScore.genes,
+    reaction: resWithScore.reactions,
+    subsystem: resWithScore.subsystems,
+    compartment: resWithScore.compartments,
   };
 };
 
--- a/frontend/src/components/SearchTable.vue
+++ b/frontend/src/components/SearchTable.vue
@@ -173,7 +173,7 @@ import Loader from '@/components/Loader';
 import ExportTSV from '@/components/shared/ExportTSV';
 import 'vue-good-table/dist/vue-good-table.css';
 import { default as chemicalFormula } from '@/helpers/chemical-formatters';
-import { sortResults } from '@/helpers/utils';
+import { sortResultsScore, sortResultsSearchTerm } from '@/helpers/utils';
 import { default as messages } from '@/content/messages';
 
 export default {
@@ -497,8 +497,8 @@ export default {
       // store choice only once in a dict
       Object.keys(this.searchResults).forEach((componentType) => {
         const compoList = this.searchResults[componentType];
-        // sort
-        compoList.sort((a, b) => this.sortResults(a, b, this.searchedTerm));
+        compoList.sort((a, b) => sortResultsScore(a, b));
+        compoList.sort((a, b) => sortResultsSearchTerm(a, b, this.searchedTerm));
         compoList.forEach((el) => { // e.g. results list for metabolites
           if (componentType === 'metabolite') {
             Object.keys(filterTypeDropdown[componentType]).forEach((field) => {
@@ -673,7 +673,7 @@ export default {
       return `${header.join('\t')}\n${tsvContent}`;
     },
     chemicalFormula,
-    sortResults,
+    sortResultsScore,
   },
 };
 
--- a/frontend/src/helpers/utils.js
+++ b/frontend/src/helpers/utils.js
@@ -84,7 +84,11 @@ export function reformatChemicalReactionHTML(reaction, noLink = false, model = '
   return `${reactants} ${equationSign(reaction.reversible)} ${products}`;
 }
 
-export function sortResults(a, b, searchTermString) {
+export function sortResultsScore(a, b) {
+  return b.score - a.score;
+}
+
+export function sortResultsSearchTerm(a, b, searchTermString) {
   let matchSizeDiffA = 100;
   let matchedStringA = '';
   Object.values(a).forEach((v) => {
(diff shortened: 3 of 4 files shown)


frontend/src/components/SearchTable.vue
```diff
@@ -497,8 +497,8 @@ export default {
       // store choice only once in a dict
       Object.keys(this.searchResults).forEach((componentType) => {
         const compoList = this.searchResults[componentType];
-        // sort
-        compoList.sort((a, b) => this.sortResults(a, b, this.searchedTerm));
+        compoList.sort((a, b) => sortResultsScore(a, b));
```
mihai-sysbio, line 500: If the results are already sorted by score when the API response arrives in `frontend/src/store/modules/search.js `, then perhaps it's not needed here.

mihai-sysbio:
This is not testable for me - none of the suggested search terms return results. In the file changes I see few changes to the Neo4j queries, maybe something is missing from the commits?

mihai-sysbio:
I just looked at the API response when performing a GemSearch. The response there is too rich - it contains much more information that is needed to be displayed, so it incurs unnecessary computation costs for the database query and result processing in the API. I know the search code is reused between GemSearch and Global search, but I think we should open a separate issue to split the search. Even if it would lead to some duplicated code, it would be more maintainable imo.

frontend/src/components/SearchTable.vue
inghylt: They are not sorted when arriving. But the sorting could ofc be done in the backend instead

mihai-sysbio: Oh I must have gotten confused then by this line https://github.com/MetabolicAtlas/MetabolicAtlas/blob/555e7d24a97bdfdc19be5c5382c8ab5ca25e855e/frontend/src/store/modules/search.js#L57

inghylt:
Sorry for not mentioning the required rebuild using branch in data-generation, the description is updated now. So it should work if you follow that @mihai-sysbio (looking forward to having PR templates :upside_down_face: )

inghylt:
And yeah I also feel like the GemSearch response is a bit overwhelming, and that was the case before this PR right? I'm up for having an issue for that

api/src/neo4j/queries/search.js
```diff
@@ -351,13 +352,33 @@ LIMIT ${limit}
...
+            resWithScore[component] = result.map(obj => {
+                    const objWithScore = {
+                        ...obj,
+                        score: idsToScore[obj.id]
+                    }
+                    return objWithScore;
+                })
+        } else {
+            resWithScore[component] = []
+        }
```
mihai-sysbio, line 373: Is it really possible have no scores?

mihai-sysbio (approved):
This looks good so far!

api/src/neo4j/queries/search.js
inghylt: If there are no result returned from fetch[Component] query (line 347-354), then an empty array should be returned as the search result for that component. So this replaces the previous return statement that returned empty array if no result

```
return {
    metabolite: [...compartmentalizedMetabolites || [], ...metabolites || []],
    gene: genes || [],
    reaction: reactions || [],
    subsystem: subsystems || [],
    compartment: compartments || [],
    }
```

Valid or is it too "off context" to have it within the resWithScore creation?

frontend/src/components/SearchTable.vue
```diff
@@ -497,8 +497,15 @@ export default {
       // store choice only once in a dict
       Object.keys(this.searchResults).forEach((componentType) => {
         const compoList = this.searchResults[componentType];
-        // sort
-        compoList.sort((a, b) => this.sortResults(a, b, this.searchedTerm));
+        /* Sorted twice as sortResultsSearchTerm does not catch cases when
+        this.searchResult contains nodes that was returned because they have a relationship
```
mihai-sysbio, line 501:
```suggestion
        this.searchResult contains nodes that were returned because they have a relationship
```

api/src/neo4j/queries/search.js
mihai-sysbio: I can't think of any way to refactor this right now 😊

api/src/neo4j/queries/search.js
```diff
@@ -351,13 +352,33 @@ LIMIT ${limit}
...
+    }
```
mihai-sysbio, line 374: I think a linter would suggest some changes here:

```suggestion
  const resObj = {
    compartmentalizedMetabolites,
    metabolites,
    genes,
    reactions,
    subsystems,
    compartments
  };

  const resWithScore = {};
  for (const [component, result] of Object.entries(resObj)) {
    if (result) {
      resWithScore[component] = result.map(obj => {
        const objWithScore = {
          ...obj,
          score: idsToScore[obj.id]
        };
        return objWithScore;
      });
    } else {
      resWithScore[component] = [];
    }
  }
```

frontend/src/components/explorer/gemBrowser/GemSearch.vue
```diff
@@ -35,7 +35,7 @@
       <div v-show="searchResults.length !== 0 && !showLoader"
            class="notification is-large is-unselectable has-text-centered is-clickable py-1 mb-1"
            @mousedown="globalSearch()">
-        Limited to 50 results per type. Click here to search all integrated GEMs
+        Limited to 50 results per type. Click <span class="pseudoLink"> here </span> to search all integrated GEMs
```
mihai-sysbio, line 38: I was imagining the highlight to be applied differently, which actually surfaced a bug.

This line of text should show only when there are search results, but due to how the object `searchRsults` looks like, the line of text always shows, even if there are no results. So the condition here should be refactored to only be shown if there truly are results.

There is a place lower down in this file where `{{ messages.searchNoResult }}` is shown when there are no results. That is the place where we want our users to be encouraged to click so that they search the other models.

mihai-sysbio (changes requested):
It's really nice to have sorting now applied everywhere, as a consequence of making the old id search visible for the users 💪🏻 

The changes I'm requesting are only loosely connected to this PR: there is a bug that gets in the way of the intuitiveness of the GemSearch messages. Normally this bug should be an independent issue. To me, however, it is related to the _highlight_ part of this PR - to have a proper highlight this bug should be resolved. @inghylt we can talk about the details here, and of course I am up for setting up a separate issue.

MalinAhlberg (approved):
Looks good!
The only things that looks weird to me (although not introduced by this pr) is the list that appears if you get no matches, eg `/search?term=m222222c`

frontend/src/components/explorer/gemBrowser/GemSearch.vue
nanjiangshu: It's a bit weird that I don't see any effect of highlighting. 
Screenshot 2021-10-04 at 16 09 56

Screenshot 2021-10-04 at 16 10 28

nanjiangshu (commented):
Looks great with the new searching engine and I'm impressed that you @inghylt solved this problem nicely.  The messy results from the GemSearch (quick search) for e.g. `r0760` and `Glycerolipid metabolism` is a bit unsatisfactory though. It seems it's because the results are always shown in the order `Metabolite`, `Gene`, `Reaction` `Subsystem` and then `Compartments`. Would it be possible to rank the hits over the component types? For example, for the screenshot below, 
Screenshot 2021-10-04 at 16 19 45

Can we place the hit for subsystems on the top? But this will introduce a major change to the GemSearch, maybe @mihai-sysbio has a different opinion.

frontend/src/components/explorer/gemBrowser/GemSearch.vue
inghylt: Please have a look at aad35c954984c9715d59506fd691a59bd21a53e2 and see if that is what you are after

mihai-sysbio:
> Can we place the hit for subsystems on the top? But this will introduce a major change to the GemSearch, maybe @mihai-sysbio has a different opinion.

Issue https://github.com/MetabolicAtlas/private-issues/issues/103 has been created to improve the way search results are returned by the GemSearch - let's not focus on problem in this issue.

frontend/src/components/explorer/gemBrowser/GemSearch.vue
mihai-sysbio: This is so much better @inghylt 

image

Could you make the entire area clickable, instead of just the word `here`?

inghylt: Sure. Do you want to keep `here` styled as a link, the entire message styled as a link or no link styling at all?

mihai-sysbio: Whatever you think is best UI-wise. As long as the user doesn't have to chase a small word to click on, I'm happy.

nanjiangshu: I'm also fine with the whatever UI you think the best.

nanjiangshu:
> Issue [MetabolicAtlas/private-issues#103](https://github.com/MetabolicAtlas/private-issues/issues/103) had been created to improve the way search results are returned by the GemSearch - @nanjiangshu let's not focus on this problem in this issue.

I see. It makes sense to me.

nanjiangshu (approved):
Looks good to me now.

inghylt:
@e0 Really appreciated your comments, nice work! Merged them both

@nanjiangshu Good work catching that bug! Fixed by commit c9edb5fd77c51541d1e3f9bd90cfa673ff2ffb36. If you find the commit does the trick please resolve the comment and I am ready to merge

PR #701, opened 2021-09-30
fix: update to not use old name for data-generation repo

inghylt:


mihai-sysbio (approved):
Great you spotted this, should be merged asap since it's blocking building.

nanjiangshu (approved):

PR #708, opened 2021-10-01
Feat: landing page for Met Atlas ids

nanjiangshu:
This PR closes [#132](https://app.zenhub.com/workspaces/metatlas-sprint-607745622d49260019ce115a/issues/metabolicatlas/private-issues/132)

Test:
* A valid Metabolite ID: `/identifier/MetabolicAtlas/MAM10036c`
* A valid Reaction ID: `/identifier/MetabolicAtlas/MAR10126`  
* An in-valid ID: `/identifier/MetabolicAtlas/MAM10036cc`

It has to be noted that the deprecated IDs listed at https://github.com/SysBioChalmers/Human-GEM/tree/main/data/deprecatedIdentifiers, such as `MAM03973p` are not deprecated, i.e. exist in the database.


api/src/neo4j/queries/externalDb.js
```diff
@@ -1,10 +1,22 @@
...
 
 const getComponentsForExternalDb = async ({ dbName, externalId }) => {
-  const statement = `
+  let statement = `
 MATCH (db:ExternalDb {dbName: '${dbName}', externalId: '${externalId}'})-[v]-(c)
 RETURN { externalDb: properties(db), components: COLLECT({ component: c, version: type(v) }) }
 `;
+  if (dbName === 'MetabolicAtlas') {
+    statement = `
+MATCH (r:Reaction {id: '${externalId}'})-[v]-(c)
```
mihai-sysbio, line 10: It doesn't look like the other node's details are relevant here:
```suggestion
MATCH (r:Reaction {id: '${externalId}'})-[v]-()
```

api/src/neo4j/queries/externalDb.js
```diff
@@ -1,10 +1,22 @@
...
+RETURN { externalDb: properties(r), components: COLLECT(DISTINCT({component: r, version: type(v)}))}
+
+UNION
+
+MATCH (r:CompartmentalizedMetabolite {id: '${externalId}'})-[v]-(c)
```
mihai-sysbio, line 15: or here:

```suggestion
MATCH (r:CompartmentalizedMetabolite {id: '${externalId}'})-[v]-()
```

api/src/neo4j/queries/externalDb.js
```diff
@@ -21,6 +33,12 @@ RETURN { externalDb: properties(db), components: COLLECT({ component: c, version
     };
   });
 
+  if (dbName === 'MetabolicAtlas') {
+    externalDb.dbName = 'MetabolicAtlas';
```
mihai-sysbio, line 37: To reduce the amount of hard-coded strings, I would suggest:
```suggestion
    externalDb.dbName = dbName;
```

frontend/src/components/ExternalDb.vue
```diff
@@ -53,12 +63,23 @@ export default {
...
-      externalId: this.$route.params.identifierId,
-    });
+    try {
+      await this.$store.dispatch('externalDb/getComponentsForExternalDb', {
+        dbName: this.$route.params.dbName,
+        externalId: this.$route.params.identifierId,
+      });
+    } catch {
+      const regex = /^MA[MR]/;
+      const found = this.$route.params.identifierId.match(regex);
```
mihai-sysbio, line 78: I'm wondering if it wouldn't be more reliable here to use `this.$route.params.dbName`.

frontend/src/content/messages.js
```diff
@@ -3,6 +3,7 @@ export default {
   tooManyInteractionPartner: 'There are too many interaction partners to be displayed, use our API to get the results.',
   modelNotFound: 'The url contains a model identifier that does not exist in the database. Browse the available models on the Explore page.',
   mapNotFound: 'Sorry, it seems that the selected 2D map is not available because the SVG is missing.<br>If you think this is an accident, drop us a line at <a href="mailto:contact@metabolicatlas.org?subject=SVG error&body=There was an error when choosing to see the 2D map for ..">contact@metabolicatlas.org</a>.',
+  maIDNotFound: 'It looks like this identifier is no longer part of any of the models integrated into our website. For more details about this id please look <a href="https://github.com/SysBioChalmers/Human-GEM/tree/main/data/deprecatedIdentifiers">here</a>.',
```
mihai-sysbio, line 6:
```suggestion
  maIDNotFound: 'The identifier you are looking for cannot be found. It might be that this identifier is no longer part of any of the models integrated into our website. For more details about this id please look <a href="https://github.com/SysBioChalmers/Human-GEM/tree/main/data/deprecatedIdentifiers">here</a>.',
```

mihai-sysbio (changes requested):
This is exact what I imagined, and I'm very happy with it.
I left some minor comments, but I'm requesting changes because of the error page that is shown when an invalid id is queried for:

image

frontend/src/components/ExternalDb.vue
```diff
@@ -27,14 +27,24 @@
         </li>
       </ul>
     </div>
+    <div v-else>
```
e0, line 30:
```suggestion
    <div v-else-if="errorMessage.length > 0">
```

The error notification appears to flash as the page is loading, this should make it not show by default.

frontend/src/components/ExternalDb.vue
```diff
@@ -27,14 +27,24 @@
...
+      <p class="notification has-background-danger-light" style="margin-top: 5%;" v-html="errorMessage"></p>
```
e0, line 31:
```suggestion
      <p class="notification has-background-danger-light mt-6" v-html="errorMessage"></p>
```

Not sure exactly what we decided on avoiding inline styles, but is this an okay replacement?

e0 (commented):
Looks good, nice job on the `UNION` query! Left a few minor suggestions here. I'm happy to approve after the other issues pointed out by @mihai-sysbio are addressed.

api/src/neo4j/queries/externalDb.js
inghylt: Minor detail, but I would rename `r` to `cm`

frontend/src/content/messages.js
inghylt, line 6:
```suggestion
  maIDNotFound: 'It looks like this identifier is no longer part of any of the models integrated into our website. For more details about deprecated identifiers please look <a href="https://github.com/SysBioChalmers/Human-GEM/tree/main/data/deprecatedIdentifiers">here</a>.',
```

inghylt:
Nice job!

I'm wondering, would it make sense to have `/identifier/metatlas/MAM10036c` instead of `/identifier/MetabolicAtlas/MAM10036c` since this prefix is mentioned in https://github.com/MetabolicAtlas/private-issues/issues/50? (@mihai-sysbio I think you have input on this question)

MalinAhlberg (approved):
:+1:

mihai-sysbio:
> I'm wondering, would it make sense to have `/identifier/metatlas/MAM10036c` instead of `/identifier/MetabolicAtlas/MAM10036c` since this prefix is mentioned in [MetabolicAtlas/private-issues#50](https://github.com/MetabolicAtlas/private-issues/issues/50)? (@mihai-sysbio I think you have input on this question)

I'm happy you remembered that @inghylt.
The pattern we're using for all databases is a `database/id`. For other databases, we are using the full name, so we should do the same for our database too.
The note for `metatlas` is regarding when applying for the identifier service at identifiers.org.

api/src/neo4j/queries/externalDb.js
nanjiangshu: It seems one can not use different column names when `UNION` is used @inghylt. I got this error ```All sub queries in an UNION must have the same column names (line 5, column 1 (offset: 147)) "UNION" ^``` if I change `r` to `cm`

frontend/src/content/messages.js
nanjiangshu: Addressed by b6711b6d

api/src/neo4j/queries/externalDb.js
nanjiangshu: Sorry that I missed this scenario. It has been fixed by 22495836.
The refactoring of the neo4j querying has been done by 037371d0 and  4c33332d

api/src/neo4j/queries/externalDb.js
nanjiangshu: Addressed by 4c33332d

frontend/src/components/ExternalDb.vue
nanjiangshu: I added checking for both to make sure the error message is displayed only when querying for MetAtlas IDs and when the ID matches MAR or MAM; fixed in ab38b8ef

e0:
@nanjiangshu Good work on addressing the changes. Can you please take another look at the two suggestion I made? From what I can see it looks like the commit message is meant for the other suggestion and also did not remove the inline styling.

frontend/src/components/ExternalDb.vue
nanjiangshu: Thanks @e0 for pointing this out. It has been fixed by ab38b8ef with a similar if condition.

api/src/neo4j/queries/externalDb.js
inghylt: https://neo4j.com/docs/cypher-manual/current/clauses/union/

Yup, that seems to be the case! Good to know

frontend/src/components/ExternalDb.vue
nanjiangshu: Good point. I've changed it in a2b055cc

frontend/src/content/messages.js
nanjiangshu: It makes more sense. I've changed it in f0414776

nanjiangshu:
> @nanjiangshu Good work on addressing the changes. Can you please take another look at the two suggestion I made? From what I can see it looks like the commit message is meant for the other suggestion and also did not remove the inline styling.

Sorry Shan @e0, it should be this commit 2249583  that fixed your first comment. And the inline styling has been fixed in a2b055cc just after you posted the message :)

frontend/src/components/ExternalDb.vue
e0: Slightly less ideal with a potential empty div but acceptable.

e0 (approved):

frontend/src/components/ExternalDb.vue
mihai-sysbio: As shared in another screenshot, the message appears wider than our usual `container` width, so one idea could be to add it here:

```suggestion
    <div v-else class="container">
```

mihai-sysbio (approved):
Good job!

inghylt (approved):

Issue #696, opened 2021-09-27
Templates for issues and PR:s


inghylt:
As a developer, I would like to have default template for issues and PR:s, so that I follow standards we have agreed upon and don't forget to add items such as print screens, how to recreate a problem, how to test, relative links etc.

These should be applied to all repositories (4).

mihai-sysbio:
> _Please edit this to-do list of things to include_

Issues:
- [ ] user story (rarely not applicable)
- [ ] in case of a bug, a description of how to reproduce it, with screenshots/video

PRs:
- [ ] relative url that can be reused both for production and local testing
- [ ] screenshot of the fix

MalinAhlberg:
Here are some additional suggestions:

Issues:
- [ ] in case of a bug, expected behaviour
- [ ] proposed solution
- [ ] definition of done

PRs:
- [ ] what type of change the PR does (bug fix, feature, documentation...)
- [ ] related issues and PRs
- [ ] list of changes made

As I think there might be discussions about which of these that should be used, I did not edit your post above @mihai-sysbio , but feel free to put them there if you approve of any of them.

mihai-sysbio:
Thanks @MalinAhlberg, I added in your suggestions.

The only thing I left out is the _definition of that_. According to scrum, this is something that should be equally applied to all issues, so it should be defined outside each individual issue. It would be really nice to have this DoD defined in the team.

inghylt:
Below follows a suggestion for a PR template:

### Related issue(s) and PR(s)
This PR closes [issue number] and relates to [issue or PR number]

### Description of the changes proposed in the pull request
#### Type of change
Please delete options that are not relevant.
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] [Other]
 
#### List of changes made
Specify what changes have been made and why

#### Screenshot of the fix
Attach screenshot if relevant

### How to test
Please delete options that are not relevant
- A certain branch in another repo is needed
- A rebuild is needed to due new dependencies
- Relative url that can be reused both for production and local testing
- Instructions on how to test


### Further comments
If relevant, specify questions or related information

### Checklist
- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
[...]

Issue #684, opened 2021-09-06
Facilitate for new users


MalinAhlberg:
As a new user, the following would help getting started:

- add a note to the Readme of this project that the repo `data-generation` also needs to be installed (not only cloned).
- add bash (?) script to check that correct versions of node and yarn are installed.
- consider adding templates for issues

PR #713, opened 2021-10-11
docs: updated readme for new users

MalinAhlberg:
This addresses the first two points of #684.

Regarding the special instructions for Linux users, I'm not sure if this is what @inghylt and @nanjiangshu had in mind?

--- 
As for the other points of #684: 

> add a note to the Readme of this project that the repo data-generation also needs to be installed (not only cloned)."
- I'm planning to create a PR to the `/data-generation` repo for this.

> describe how to work with Zenhub (linking PRs with unclosed issues)

- Should this be done in the wiki instead of here?

> consider adding templates for issues

- Addressed by #696.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,7 +95,7 @@ When rebuilding the stack, you might have to change the ownership of the directo
 sudo chown -R <user> neo4j
 ```
 Replace `<user>` with your user name.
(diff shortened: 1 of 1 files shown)


inghylt:
Nice!

PR #715, opened 2021-10-15
fix: do not show NavigationDuplicated error when clicking a map twice

### LINKED ISSUE #697: Fix error NavigationDuplicated when clicking map twice
Error _Uncaught (in promise) NavigationDuplicated: Avoided redundant navigation to current location: "/explore/Human-GEM/map-viewer/_ occurrences when a user clicks on the same map twice. This happens for both 2d and 3d maps. See attached video for example

nanjiangshu:
This PR closes #697 

DoD: The error message `NavigationDuplicated` does not show in the console when clicking a map twice in the Map viewer for both 2D and 3D maps.

### DIFF
--- a/frontend/src/components/explorer/mapViewer/MapsListing.vue
+++ b/frontend/src/components/explorer/mapViewer/MapsListing.vue
@@ -55,7 +55,7 @@ export default {
       const oldDim = this.$router.currentRoute.query.dim;
       const newDim = this.showing2D ? '2d' : '3d';
       if (oldMapId !== newMapId || oldDim !== newDim) {
-        this.$router.push({ params: { map_id: newMapId }, query: { dim: this.showing2D ? '2d' : '3d' } });
+        this.$router.push({ params: { map_id: newMapId }, query: { dim: newDim } });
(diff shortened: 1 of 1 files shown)


mihai-sysbio:
> DoD: The error message `NavigationDuplicated` does not show in the console when clicking a map twice in the Map viewer for both 2D and 3D maps.

To me, a more robust solution would be to not trigger these errors at all, for example by avoiding navigation when `from` and `to` urls are the same, or if the `params` for the route are the same as the current ones. Would such an approach be possible?

nanjiangshu:
The new solution by checking if there is any change in the params is implemented with the commit a3ec9c9d

mihai-sysbio:
Thanks for having another go at this @nanjiangshu . Was it not possible to use `<router-link>` instead of `<a>`?

https://github.com/MetabolicAtlas/MetabolicAtlas/blob/431a7caf495eb7fbcd77fa27717cbdc50b51770f/frontend/src/components/explorer/mapViewer/MapsListing.vue#L12-L27

PR #716, opened 2021-10-15
fix: only show gem-search when appropriate

inghylt:
This PR implements:

_As a user, I would like the window switches to the global search directly after clicking the button "Search all integrated GEMs" in the GemSearch. Currently, the web-page is directed to the global search after clicking the button but it is grey and deactivated._

To test, try different searches in the gem search. When searching for things not existing, i.e. `POLR3F` in `Yeast-GEM`, make sure that the global search page is shown and that the gem search disappears

### DIFF
--- a/frontend/src/components/explorer/gemBrowser/GemSearch.vue
+++ b/frontend/src/components/explorer/gemBrowser/GemSearch.vue
@@ -277,10 +277,6 @@ export default {
     border-top: 0;
     z-index: 30;
 
-    .globalSearchLink{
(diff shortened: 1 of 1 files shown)


MalinAhlberg:
> typing too fast without waiting for the search makes some letters disappear

I can confirm this, if I start typing, wait for the search to start and then continue typing before it's finished, letters disappear.

inghylt:
@mihai-sysbio I would say create one/several issues as you see fit, no need to have more people confirming

PR #717, opened 2021-10-15
feat: add template for issues and PR:s

inghylt:
Please have a look at the suggested templates for issues and PR:s.

I'm thinking that we only really need two types of issues, bugs (called Bug report, see attached image) and all others (called Feature request, see attached image) 

This is how the Issues section would look like in GitHub after you press New issue. Note that you can create a so called blank issue if you don't want to have a template. In  ZenHub it'll be a dropdown, and it is confirmed that it can handle multiple templates.

image

- Anything I should add/remove or other comments?
- Is it annoying to have to remove template text every time? An alternative could for example be to have things quite bare bone, such as `This PR closes []` instead of `This PR closes [issue number] and relates to [issue or PR number]` and thus not enforcing the creator to remove text
- It looks like we can add the templates for all repositories at  once following [these instructions](https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/creating-a-default-community-health-file). So I'm having this in draft mode and then we can  create a repository for default files and add them there

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -33,3 +33,4 @@ Please delete options that are not relevant.
 - [ ] I have commented my code, particularly in hard-to-understand areas
 - [ ] I have made corresponding changes to the documentation
 - [ ] My changes generate no new warnings
+- [ ] I have checked so that the same data as before is returned by the API


mihai-sysbio:
Not sure if/how to use this @inghylt, but enclosing text within the html comment tags `<--` and `-->` seems to work in these templates.

inghylt:
@e0 They layout of the md files is actually what github suggests by default when creating a new issue template. When creating an issue it will look like this so I don't think it is a problem

image

image

PR #723, opened 2021-10-19
feat: reuse data overlay component in interaction partners page

### LINKED ISSUE #709: Implement the dataOverlay as in the 2D Map Viewer for Interaction Partners tool
As a user, I want the new features in the dataoverlay for Map Viewer be implemented in the Interaction Partners.

> Original text #706 _As a user, I want to use the data overlay feature in the Interaction Partners tool in the same way as in the Map Viewer, so that I don't need to understand a new interface._

This issue is also related to #706

There's no custom data for the Interaction Partners before but now we want to implement it. The same parser as in the Map viewer can be used.

e0:
This PR closes #709 

To test this PR, please use the following interaction partners page: `/explore/Human-GEM/interaction-partners/MAM01357c` to test metabolites that are included in the sample metabolomics data. Any page should work fine for transcriptomics.

Please also pay attention to the map viewer using the following page for example `/explore/Human-GEM/map-viewer/golgi_apparatus?dim=2d&panel=1&datatype=transcriptomics&datasource=hpaRna.tsv&dataSet=none`. There seems to be a difference in how the data overlay is rendered (please see comments of this PR for more details). Please let me know if the data overlay looks ok in the map viewer or if it looks similar to the one [posted by @mihai-sysbio](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/723#issuecomment-947722719).

---

### Draft PR description

Creating a draft PR in case anyone wants to take a look at the progress so far. Currently I have added the `DataOverlay` component to the interaction partners page. The data overlay seems to be working as well. You can test with `transcriptomics` -> `Human Protein Atlas 20.1` -> any level/dataset on the following page (any other interaction partners page should work too):  `/explore/Human-GEM/interaction-partners/ENSG00000153015`.

Please don't mind the code too much right now. I would like to continue with some refactoring and clean-up after we clear up the questions below.

## Questions

- I had made some basic modifications to adapt the `DataOverlay` component so it looks acceptable on the interaction partners page. Do we think this is ok or would we like to improve it? Keep in mind that it should satisfy both the interaction partners page and the map viewer.
- So far only gene data works. Do we want to make sure metabolites data overlay works as well? I think it makes sense to only include gene data types for the interaction partners page's graph is very gene-centric since very few (usually 1 or 2) metabolites are displayed in the graph. 
- I have removed the `margin-right: -1rem;` in the `DataOverlay` component, which affects the map viewer as well. Does anyone remember/know why this was necessary? It was actually cropping the contents of the `DataOverlay` component in the map viewer so this would be an improvement there as well unless there is a good reason for not removing it.

### DIFF
--- a/frontend/src/components/explorer/mapViewer/DataOverlay.vue
+++ b/frontend/src/components/explorer/mapViewer/DataOverlay.vue
@@ -1,7 +1,7 @@
 <template>
   <div id="dataOverlayBox"
-       class="column is-one-fifth-widescreen is-one-quarter-desktop
-         is-one-quarter-tablet has-background-lightgray">
+       :class="{ 'is-one-fifth-widescreen is-one-quarter-desktop is-one-quarter-tablet': position === 'absolute'}"
+       class="column has-background-lightgray">
     <div class="title is-size-4 has-text-centered">Expression data</div>
     <div class="has-text-centered"
          title="Load a TSV file with IDs and TPM values.
@@ -107,6 +107,10 @@ export default {
     mapType: String,
     mapName: String,
     dim: String,
+    position: {
+      type: String,
+      default: 'absolute',
+    },
   },
   data() {
     return {
@@ -232,11 +236,6 @@ export default {
 </script>
 
 <style lang="scss">
-@media screen and (min-width: $tablet) {
-  #dataOverlayBox {
-    margin-right: -1rem;
-  }
-}
 
 #fileNameBox {
   display: flex;
--- a/frontend/src/components/explorer/InteractionPartners.vue
+++ b/frontend/src/components/explorer/InteractionPartners.vue
@@ -944,6 +944,7 @@ export default {
 
     .dropdown-menu {
       display: block;
+      min-width: unset;
     }
   }
 
--- a/frontend/src/components/explorer/InteractionPartners.vue
+++ b/frontend/src/components/explorer/InteractionPartners.vue
@@ -166,42 +166,40 @@
                   </div>
                 </div>
                 <div id="cy" ref="cy" class="card p0">
+                  <div id="dropdownMenuExport" class="dropdown">
+                    <div class="dropdown-trigger">
+                      <a v-show="showNetworkGraph" class="button is-white" aria-haspopup="true"
+                         aria-controls="dropdown-menu" @click="showMenuExport=!showMenuExport">
+                        <span class="icon is-large"><i class="fa fa-download"></i></span>
+                        <span>Export</span>
+                        <span class="icon is-large"><i class="fa fa-caret-down"></i></span>
+                      </a>
+                    </div>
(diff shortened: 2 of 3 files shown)


mihai-sysbio:
I'm trying to have a look this now as well.

frontend/src/components/explorer/mapViewer/DataOverlay.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <div id="dataOverlayBox"
-       class="column is-one-fifth-widescreen is-one-quarter-desktop
-         is-one-quarter-tablet has-background-lightgray">
+       :class="{ 'is-one-fifth-widescreen is-one-quarter-desktop is-one-quarter-tablet': position === 'absolute'}"
```
mihai-sysbio, line 3:
```suggestion
       :class="{ 'is-one-fifth-widescreen is-one-quarter-desktop is-one-quarter-tablet': this.position === 'absolute'}"
```

Without the suggestion I got this:
image

mihai-sysbio (changes requested):
Very nice refactoring! This is a pretty lightweight review from my side. I'm requesting changes for the suggestion below, really minor thing but with significant impact on the interface.
I particularly liked the handling of the `Export` button for the graph. I think it could be further improved by making it look like the rest of the `Export` buttons (with green rounded box and such) but let's leave that for another time. One particular suggestion here is to remove the `min-width` on the submenu of `Export` dropdown so that it attempts to stay within the graph box.
Again, very nice work!

frontend/src/components/explorer/mapViewer/DataOverlay.vue
e0: I don't have this visual problem here. Which browser is this from?

mihai-sysbio: Firefox 93. I also freshly rebuilt the stack with a new `yarn.lock`, perhaps you can try that.

mihai-sysbio: Sorry it looks like it was a cache thing - the issue has disappeared without my suggestion.

e0:
> Very nice refactoring! This is a pretty lightweight review from my side. I'm requesting changes for the suggestion above, really minor thing but with significant impact on the interface.
> I particularly liked the handling of the `Export` button for the graph. I think it could be further improved by making it look like the rest of the `Export` buttons (with green rounded box and such) but let's leave that for another time. One particular suggestion here is to remove the `min-width` on the submenu of `Export` dropdown so that it attempts to stay within the graph box.
> Again, very nice work!

Thanks @mihai-sysbio! The `min-width` change is done [here](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/723/commits/91e7e7b61958ba88e5a7558957f12bd1fad4144e).

mihai-sysbio (approved):
Great!

PR #724, opened 2021-10-22
feat: count genes, reactions and metabolits dynamically

### LINKED ISSUE #656: dynamic model component counts
As a user, I want to see the model component counts, so that I know how large a model is.

As a developer I want the reaction, metabolite and gene counts for each model to be automatically updated, so that I don't need to manually update these.

At the moment, the counts are statically stored in the `data-files` repository, in `integratedModels.json`.

nanjiangshu:
This PR closes #656 

### Changes made
The number of genes, reactions and metabolites are obtained dynamically from the neo4j database instead of reading from [integratedModels.json](https://github.com/MetabolicAtlas/data-files/blob/main/integrated-models/integratedModels.json)

### Test
You can compare the counts at `/explore` to those listed at https://metabolicatlas.org/explore 

### Problem 
The current code requires querying the database 7x3 times and thus it takes 1-2 seconds to refresh the page. Maybe you have some ideas to UNION the querying?  

I have tried to combine the querying in various ways. You can try out with branches `test/count-binding-1` or `test/count-binding-2`, but it seems the speed improvement is not obvious. 
The branch `test/count-binding-1` merges all matches for a model.
The branch `test/count-binding-2` merges all 21 matches.  

However, it is 3 times faster when `version` is not matched, as done in the branch `test/count-binding-3-no-version`.

### DIFF
--- a/api/src/endpoints/repository.js
+++ b/api/src/endpoints/repository.js
@@ -9,6 +9,23 @@ import {
 
 const routes = express.Router();
 
+const addCountToModel = async (model) => {
+  const { apiName, apiVersion } = model;
+
+  const [gene_count, reaction_count, metabolite_count] = await Promise.all([
+    getGeneCount(apiName, apiVersion),
+    getReactionCount(apiName, apiVersion),
+    getMetaboliteCount(apiName, apiVersion),
+  ]);
+
+  return {
+    ...model,
+    gene_count,
+    reaction_count,
+    metabolite_count,
+  };
+};
+
 routes.get('/integrated_models', async (req, res) => {
   try {
     const models = integratedGemsRepoJson.map(model => ({
@@ -17,13 +34,8 @@ routes.get('/integrated_models', async (req, res) => {
       apiVersion: model.version.split('.').join('_'),
     }));
 
-    for ( const model of models ){
-      model.gene_count = await getGeneCount(model.apiName, model.apiVersion);
-      model.reaction_count = await getReactionCount(model.apiName, model.apiVersion);
-      model.metabolite_count = await getMetaboliteCount(model.apiName, model.apiVersion);
-    }
-
-    res.json(models);
+    const modelsWithCount = await Promise.all(models.map(m => addCountToModel(m)));
+    res.json(modelsWithCount);
   } catch (e) {
     res.status(400).send(e.message);
   }


mihai-sysbio:
I'm wondering if this couldn't be done while parsing the `yml` and/or `tsv` files. It feels like the model components are already there in some data structures, so fetching the counts for everything (including subsystems and compartments) would be instant. What do you think @nanjiangshu ?

e0:
> I'm wondering if this couldn't be done while parsing the `yml` and/or `tsv` files. It feels like the model components are already there in some data structures, so fetching the counts for everything (including subsystems and compartments) would be instant. What do you think @nanjiangshu ?

It feels like we might have discussed this before and decided against it based on [this comment](https://github.com/MetabolicAtlas/MetabolicAtlas/issues/656#issuecomment-882536586).

mihai-sysbio:
> It feels like we might have discussed this before and decided against it based on [this comment](https://github.com/MetabolicAtlas/MetabolicAtlas/issues/656#issuecomment-882536586).

There might be a difference of interpretation there. To me, that is the issue scoping, and I read the comment

> This does not **involve** modifying the `integratedModels.json` file during import.

as

> This does not **require** modifying the `integratedModels.json` file during import.

So yeah, while I believe it would be technically to complicated to version the changes in that file, there are others ways to modify that file without versioning.

e0:
> > It feels like we might have discussed this before and decided against it based on [this comment](https://github.com/MetabolicAtlas/MetabolicAtlas/issues/656#issuecomment-882536586).
> 
> There might be a difference of interpretation there. To me, that is the issue scoping, and I read the comment
> 
> > This does not **involve** modifying the `integratedModels.json` file during import.
> 
> as
> 
> > This does not **require** modifying the `integratedModels.json` file during import.
> 
> So yeah, while I believe it would be technically to complicated to version the changes in that file, there are others ways to modify that file without versioning.

Ok, seems like there was a misunderstanding. Either way, since this is already merged and working I personally feel like implementing a new solution may not be the best allocation of time considering the potential benefits (hundreds of ms faster for the first request after a new deployment). If you still feel like there are other benefits that make this clearly the superior solution please elaborate or create a new issue.

For the future, it would be nice if things like this could be discussed and clarified during backlog refinement if possible to help save time.

mihai-sysbio:
> since this is already merged and working I personally feel like implementing a new solution may not be the best allocation of time considering the potential benefits (hundreds of ms faster for the first request after a new deployment).

I see your point of view @e0, sorry for giving you the impression that I was _requesting_ changes.

At this stage I'm more interested in discussing the technical feasibility of a different solution, especially that @nanjiangshu is very familiar (expert) with the `tsv` parser. My intuition is that it would be less code, potentially easier to maintain, which is always a concern, and fast, which was @nanjiangshu's concern above.

nanjiangshu:
Hi @mihai-sysbio, I see your point to obtain the count by parsing the tsv or yaml files. However, since the files `genes.tsv` are mostly empty for most GEMs, we have to relay on the YAML file to obtain these counts and I'm afraid it won't take less time to parse YAML files than the current solution. There might be other better solutions but I agree with @e0  that the current solution is good enough for the moment.

mihai-sysbio:
Hmm good thinking with the empty `genes.tsv` @nanjiangshu . What about this line though?

https://github.com/MetabolicAtlas/data-generation/blob/e28d3034d46b6dd8cbfe4df487c8e00b132aff6d/parser.js#L8

Having this code under it seems to do the job:
```
console.log(genes.genes.length);
```

Issue #725, opened 2021-10-26
Interaction partners page not working


MalinAhlberg:
This page `explore/Human-GEM/interaction-partners/MAM02751c` seems to be broken.

Possibly caused by `The query has returned many nodes` (noted by Nanjiang).

PR #727, opened 2021-10-26
fix: use v-html for error messages

MalinAhlberg:
This PR closes [#39](https://app.zenhub.com/workspaces/metatlas-sprint-607745622d49260019ce115a/issues/metabolicatlas/3d-network-viewer/39) and relates to [#725 ]

#### Type of change

 
#### List of changes made
Using `v-html` for error messages, as these might contain html elements.

#### Screenshot of the fix
2021-10-26-141432_644x179_scrot

### Testing
- Instructions on how to test
To test this, you can add
```
    this.errorMessage = messages.unknownError;
```
at line 437 in `InteractionPartners.vue` and visit `/explore/Human-GEM/interaction-partners`.
Note that the change has been done on multiple places, so please click around / experiment to verify that it's working in all places.

# Checklist:

### DIFF
--- a/frontend/src/components/Repository.vue
+++ b/frontend/src/components/Repository.vue
@@ -1,8 +1,7 @@
 <template>
   <section class="section extended-section">
     <div class="container is-fullhd">
-      <div v-if="errorMessage">
-        {{ errorMessage }}
+      <div v-if="errorMessage" v-html="errorMessage">
       </div>
       <div v-else>
         <h3 id="Integrated-models" class="title is-3">Integrated models</h3>
--- a/frontend/src/components/Repository.vue
+++ b/frontend/src/components/Repository.vue
@@ -1,8 +1,7 @@
 <template>
   <section class="section extended-section">
     <div class="container is-fullhd">
-      <div v-if="errorMessage" v-html="errorMessage">
-      </div>
+      <div v-if="errorMessage" v-html="errorMessage" />
       <div v-else>
         <h3 id="Integrated-models" class="title is-3">Integrated models</h3>
         <p class="has-text-justified">
--- a/frontend/src/components/explorer/GemBrowser.vue
+++ b/frontend/src/components/explorer/GemBrowser.vue
@@ -3,8 +3,7 @@
     <div class="container is-fullhd">
       <template v-if="errorMessage">
         <div class="column notification is-danger is-half is-offset-one-quarter has-text-centered"
-             v-html="errorMessage">
-        </div>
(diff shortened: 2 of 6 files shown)


MalinAhlberg:
> Looks good!
> 
> I'm not sure if we have any guidelines about this but I personally prefer to use single html tags when possible. What do other people think?
> 
> ```js
>  <div v-if="errorMessage" v-html="errorMessage" />
> ```

Good suggestion, thanks! It is fixed by 82745f32da9ccb013be3e0042913459bd06ca184

MalinAhlberg:
> I think it is neat to use `v-html` here, maybe this can be applied in more places?

Yes, I changed all places I could find that involves error messages, but there might of course be even more places. Maybe that's a new issue :) ?

mihai-sysbio:
As a general practice, it is safer to go for the double mustaches `{{ }}` in templates, according to the [documentation](https://vuejs.org/v2/guide/syntax.html):

> Only use HTML interpolation on trusted content and never on user-provided content.

Using double mustaches is done consistently in the codebase, with the few exceptions where this was not possible. Therefore, to me everything that can use the double mustaches should continue to do so.

mihai-sysbio:
I just noticed, we seem to have quite a few occurrences of the `errorMessage` with similar styling. That could be a refactoring issue.

MalinAhlberg:
> As a general practice, it is safer to go for the double mustaches `{{ }}` in templates, according to the [documentation](https://vuejs.org/v2/guide/syntax.html):
> 
> > Only use HTML interpolation on trusted content and never on user-provided content.
> 
> Using double mustaches is done consistently in the codebase, with the few exceptions where this was not possible. Therefore, to me everything that can use the double mustaches should continue to do so.

Is this a comment regarding the suggestion about using `v-html` more in general, or do you think it's a non-optimal solution as used in this pr?

mihai-sysbio:
> Is this a comment regarding the suggestion about using `v-html` more in general, or do you think it's a non-optimal solution as used in this pr?

To my knowledge, `v-html` is the _only_ way to solve this, it's the optimal solution. Expanding this to other situations should be done more carefully.

From what I'm seeing in `messages.js`, we have some plaintext errors  like `modelNotFound` and some html errors like `unknownError`. If from now on we use `v-html` for all errors, it would be very nice to emphasize the connection more clearly - perhaps by appending `Html` at the end? I just feel something should be done to avoid this exact type of bug in the future. But feel free to mark that as a separate issue.

PR #728, opened 2021-10-27
feat: neo4j parallelization for interaction partners

inghylt:
This is PR fixes issue #725 by @inghylt and @e0 

 ### A description of the changes proposed in the pull request.

#### Type of change

 
#### List of changes made
The neo4j query for Interaction Partners have been modified to make use of `apoc.cypher.mapParallel2`. This cuts the response time for id:s associated with a large number of reactions to around 25% of the previous response time.

### Testing
- A rebuild is needed if with the latest changes from 1828df6f6bda27dcd58f7ec8402930f948574dc1 if not already used on local setup
- Since we have no framework in place for testing, the best way to verify is to compare the data returned from `localhost` with `dev-metatlas.csbi.chalmers.se.` Recommended to sort the Reaction table on the reaction id. Some example links to use:

  - /explore/Human-GEM/interaction-partners/MAM02751c (ID with 731 associated reactions)
  - /explore/Human-GEM/interaction-partners/ENSG00000011083
  - /explore/Fruitfly-GEM/interaction-partners/ND-B18
  - /explore/Zebrafish-GEM/interaction-partners/MAM02085g
  - /explore/Yeast-GEM/interaction-partners/YNR030W

  But please explore other id:s and models to make sure everything is working correctly
- Note that bug #770 will show for this PR as well since it is not fixed
  
### Checklist:
- [x] My code follows the [style guidelines](https://github.com/NBISweden/development-guidelines)

### Further comments
- I am keeping the text from the draft PR below for context.
- I was not able to use neo4j `UNION` functionality within `apoc.cypher.mapParallel2`. We could create a new issue for that if we want to explore that further
- The current configuration of mapParallel2, i.e. `mapParallel2(queryString, {}, reactions, 50)`, seems to be the best configuration. Experimented with different parameters for the params map and partitions but could not find anything that made an improvement over the current configuration. See [this](https://community.neo4j.com/t/how-best-to-do-parallel-processing/13342) and [this](https://community.neo4j.com/t/using-parallel-queries-to-sum-value-of-bitcoin-outputs-connected-to-bitcoin-address-nodes/5774) example for examples of other configuration settings. We could create a new issue for this if we want to explore that further

This is a potential fix to issue #725 by @inghylt and @e0 

The reason why it takes ages to load some id:s in Interactions Partner is that they are associated with a very large amount of reactions. For example, `MAM02751c` match with 731 reactions on line 23 in `MetabolicAtlas/api/src/neo4j/queries/interactionPartners.js`. Thus, the query on line 27-52 will be run 731 times. This results in a huge network and a huge reaction table.

Example of huge network
image

### A description of the changes proposed in the pull request.
If the number of reactions are limited as in this draft PR, the response time is improved. Some different values on the limit:
50: 4169 ms
100: 8026 ms
250: 20649 ms
500: 41041 ms

### Further comments
Questions:
- Does this limit makes sense? We are limiting the amount of associated reactions/metabolites/genes returned to the user. At the same time, is any user interested in obtaining such a huge amount of associated components? Maybe for the reaction table, but probably not the network. We already have a limit on the GemBrowser (there the limit is 200, see bottom page of [example](http://localhost/explore/Human-GEM/gem-browser/metabolite/MAM02552c). Without a limit, it is probably hard to fix the query time. In this case one solution could be instead to have a info message to pop up saying "a large amount of components are associated to this id, it will take some time to calculate" if the query time exceeds some limit and increase the timeout limit in prod
- Either way we choose to go forward, we realized that paginating the reaction table both in Gem Browser and Interactions Partner is probably a good thing. It already exists a related issue to this for Gem Browser at https://github.com/metabolicAtlas/private-issues/issues/30


MalinAhlberg:
Nice work!!
I think a limit makes sense, as a larger network is so hard to see and use anyway. A future feature could be maybe be to allow the user to filter the node in some way instead of just applying a fixed limit?

mihai-sysbio:
I'm not sure I am happy with what this solution provides the users with:
- what gets limited is not guaranteed to be biologically/scientifically relevant
- a limit of 50 is based on tech specs, so it appears random for the user (could just as well be 40 or 60)
- the current IP viewer is problematic irrespective of this particular issue, and is marked for replacement in https://github.com/MetabolicAtlas/private-issues/issues/84 
- `50: 4169 ms` doesn't sound amazing to me anyway

inghylt:
@mihai-sysbio  Yes, the limit of 50 is arbitrary and as you say and it is not certain that was is kept is scientifically relevant. But if we use the limit of 50, the network can still be expanded in the following ways:

1. Use the “expand network” feature in the interaction partners.
2. If we add pagination, browsing to a new page could append the network with genes and metabolites from the new page.

If a limit cannot be accepted still, we don't really know how we could speed up the query that will be run 731 times. One option could be to explore https://neo4j.com/labs/apoc/4.0/overview/apoc.cypher/apoc.cypher.mapParallel/. @e0 has already done this before though without succeeding to implement it. But we could spend some hours at least and see where we end up.

Another thing to solve this for now would be to increase the timeout limit in prod and maybe add an info message to pop up saying "a large amount of components are associated to this id, it will take some time to calculate" if the query includes a large number of nodes or similar

Or do you have any other suggestions on how to move forward?

mihai-sysbio:
>     1. Use the “expand network” feature in the interaction partners.

Sadly that would still be a random limited set ("first" 50), so I'm not sure if it would help much.
 
>     2. If we add pagination, browsing to a new page could append the network with genes and metabolites from the new page.

If possible, I would like to not get into this because (1) I am not sure how the Interaction Partners feature will look in the future and (2) we have not used server-side pagination before.

> One option could be to explore https://neo4j.com/labs/apoc/4.0/overview/apoc.cypher/apoc.cypher.mapParallel/.

> increase the timeout limit in prod

Both valid options, not mutually exclusive. I guess you have already considered the possibility of splitting the query?

Another option would be to exclude the currency metabolites (with high count) from the Interaction Partners.

inghylt:
@mihai-sysbio do you have a suggestion regarding splitting the query? We are not really sure how that would be done

mihai-sysbio:
> @mihai-sysbio do you have a suggestion regarding splitting the query? We are not really sure how that would be done

The data required to construct the graph is different from the data required to populate the table, for example. So, would there be any win in separating the current query into one with minimal information to construct the graph, and the other with all the relational-database style content?

nanjiangshu:
I think it is reasonable to set a limit to reduce the querying time, maybe even less than 50, since any irresponsible time more than 1 second would be bad for a GUI user. 
The messy full IP network for the metabolite `MAM02751c` in this case is very hard to see and operate and I would say it is practically unable to use. However, for those users are interested in the full IP network (maybe for illustration of an extreme case) or the complete reaction table, I think adding an `Expand` tab would be useful. In addtion, we could also add a notification to the user that one could also retrieve the complete list of related-reactions through the API, e.g. `/api/v2/interaction-partners/MAM02751c?model=HumanGem'` in this case.

nanjiangshu (commented):
Hi @inghylt , it's really impressive that the querying is now more than 4 times faster. I've checked IP for many random components and did not find any weird behavior. However, I have two concerns.

1. The output of the API query is different when run diff, could you please double check if it is just because the ordering of the content is different?
2. The message is still not changed when the network is not generated at the beginning. Should we improve the message or just keep as it is? I think when we discussed, we mentioned the option to improve the message.

inghylt:
@nanjiangshu Thanks for the feedback!

1. The ordering of the content is indeed different due to the parallelization. I have validated with some ugly code locally that I did not want to push, and also by comparing the the data in the frontend in `localhost` with `dev-metatlas.csbi.chalmers.se`, since you are able to neatly organize the data the same way there. Is that enough or do you want proper test files?
2.  Yes we have talked about that. I feel like it might be better to do in another issue though to keep things separate and concise. Do you think I should create an issue for it?

MalinAhlberg (approved):
Very good work :clinking_glasses: !
The examples I've tried works well (but I have not made any proper comparisons of the output data).

Regarding the message when the network is not generated, I wonder if the limit should be changed? It seems likes network that would have been relatively fast to compute now are not shown, for example `/explore/Human-GEM/interaction-partners/MAM02360c`.

inghylt:
@MalinAhlberg I think that limit might be because the network itself is too crowded to provide much value:

MalinAhlberg:
> @MalinAhlberg I think that limit might be because the network itself is too crowded to provide much value:
> 

Yes, that could be it. But still, for examples like this, I think it would be nicer to just show the graph 
generated-network

Anyhow, that's irrelevant for this PR :upside_down_face: .

mihai-sysbio:
Just a thought, has anyone tried this for other models?

nanjiangshu:
> @mihai-sysbio Yes, the limit of 50 is arbitrary and as you say and it is not certain that was is kept is scientifically relevant. But if we use the limit of 50, the network can still be expanded in the following ways:
> 
>     1. Use the “expand network” feature in the interaction partners.
> 
>     2. If we add pagination, browsing to a new page could append the network with genes and metabolites from the new page.
> 
> 
> If a limit cannot be accepted still, we don't really know how we could speed up the query that will be run 731 times. One option could be to explore https://neo4j.com/labs/apoc/4.0/overview/apoc.cypher/apoc.cypher.mapParallel/. @e0 has already done this before though without succeeding to implement it. But we could spend some hours at least and see where we end up.
> 
> Another thing to solve this for now would be to increase the timeout limit in prod and maybe add an info message to pop up saying "a large amount of components are associated to this id, it will take some time to calculate" if the query includes a large number of nodes or similar
> 
> Or do you have any other suggestions on how to move forward?

nanjiangshu:
> @nanjiangshu Thanks for the feedback!
> 
>     1. The ordering of the content is indeed different due to the parallelization. I have validated with some ugly code locally that I did not want to push, and also by comparing the the data in the frontend in `localhost` with `dev-metatlas.csbi.chalmers.se`, since you are able to neatly organize the data the same way there. Is that enough or do you want proper test files?

I think it is good enough. I have also verified the json output for several examples and they are the same. 
Sorry that I posted this comment without checking it myself since it does not take more than a minute to write the verification code :)
```python
import json
with open('t1.txt') as json_file:
    data1 = json.load(json_file)
with open('t2.txt') as json_file:
    data2 = json.load(json_file)
if (sorted(data1) == sorted(data2)):
    print("the same")
else:
    print("different")
```


> 
>     2. Yes we have talked about that. I feel like it might be better to do in another issue though to keep things separate and concise. Do you think I should create an issue for it?

Please feel free to create a new issue about improving the user experience with slow responding IP queries. Maybe we can modify the title of this PR a bit since it is not limiting the amount of reactions.

api/src/neo4j/queries/interactionPartners.js
```diff
@@ -21,41 +21,32 @@ YIELD value as v
 
 WITH v.component as component
 MATCH (${m} {id: component.id})-[${v}]-(r:Reaction)
+WITH component, collect(r) as reactions
+
+CALL apoc.cypher.mapParallel2("
+ WITH (_) as reaction
+ MATCH (reaction)-[${v}]-(cm:CompartmentalizedMetabolite)
```
mihai-sysbio, line 28: This is suprising to me, since it seems there is absolutely no filter on the starting node. I'm note sure if it's meant to be like this. I guess what I intuitively was expecting to see was something of the sort `(reaction:Reaction)`, meaning that the starting `reaction` node must have the label `:Reaction`.

inghylt:
```
WITH v.component as component
MATCH (${m} {id: component.id})-[${v}]-(r:Reaction)
WITH component, collect(r) as reactions
```

`CALL apoc.cypher.mapParallel2(querystring, {}, reactions, 50) YIELD value)`

The starting node is an item in the list called reactions, which is created by the code copy pasted above (line 22-24) . So they do have they labels

inghylt:
> Just a thought, has anyone tried this for other models? That is, other than Human-GEM.

Yes, I have tried it on all models and as you can see in the Testing section multiple models are listed

api/src/neo4j/queries/interactionPartners.js
mihai-sysbio: Ah thanks 👍🏻

inghylt:
@nanjiangshu  Nice, thanks for the validation:star:!   Created #783 regarding improving the message returned when the number of nodes are larger than the limit in `frontend/src/components/explorer/InteractionPartners.vue`.

nanjiangshu (approved):
Looks great!

PR #743, opened 2021-11-02
feat: show compartments

### LINKED ISSUE #681: Inclusion of metabolite compartments in reaction equations
I noticed that one of the recent updates to Metabolic Atlas involved the removal of metabolite compartment abbreviations (e.g., `[c]`, `[m]`, etc.) from reaction equations shown in the GEM browser. What was the reason behind this change? It was pretty convenient to have these, in particular for reactions involving more than one compartment.

Taking a simple example, the reactions for glucose transport between different compartments all look the same ([MAR05027](https://metabolicatlas.org/explore/Human-GEM/gem-browser/reaction/MAR05027), [MAR07675](https://metabolicatlas.org/explore/Human-GEM/gem-browser/reaction/MAR07675), [MAR05029](https://metabolicatlas.org/explore/Human-GEM/gem-browser/reaction/MAR05029), etc.)

_Originally posted by @JonathanRob in https://github.com/MetabolicAtlas/MetabolicAtlas/discussions/679_

MalinAhlberg:
This PR closes #681  and relates to [improve formatting of related reactions](https://app.zenhub.com/workspaces/metatlas-sprint-607745622d49260019ce115a/issues/metabolicatlas/private-issues/116).

#### Type of change

 
#### List of changes made
- Functionality for adding compartments and summary to equations
- Updated background of compartments, to remove grey background

#### Screenshot of the fix
reaction2

### Testing
See eg `/explore/Human-GEM/gem-browser/reaction/MAR01377`. The Reaction Table should not be changed, see `/explore/Human-GEM/gem-browser/gene/ENSG00000117394`.
Please test any other location which you suspect might be effected by this, or just point them out so that I can test them.

### Further comments
I've added the summary at the end of the equations, but I'll remove them if not wanted.

Code wise, I've tried to reuse and adapt already existing code, but also keeping the functions as intact as possible, in order not to mess up functionality elsewhere.
`constructCompartmentStr` at line 150 is thus unchanged.
[`formatCompartmentStr`](https://github.com/MetabolicAtlas/MetabolicAtlas/compare/feat/show-compartmets?expand=1#diff-aca785bc9b0ea8ad133ab97de0856016128fdac75a6d1d1d30f3d14c45fb18adR67) is a modified version of this.
Maybe they can be merged, or one of them deleted?

# Checklist:

### DIFF
--- a/frontend/src/helpers/utils.js
+++ b/frontend/src/helpers/utils.js
@@ -47,18 +47,6 @@ export function getSimpleEquation(reaction) {
   return `${reactants} ${equationSign(reaction.reversible)} ${products}`;
 }
 
-export function getChemicalReaction(reaction) {
-  if (reaction === null) {
-    return '';
-  }
-  const reactants = reaction.reactants.map(
-    x => `${x.stoichiometry !== 1 ? `${x.stoichiometry} ` : ''}${x.fullName}`
-  ).join(' + ');
-  const products = reaction.products.map(
-    x => `${x.stoichiometry !== 1 ? `${x.stoichiometry} ` : ''}${x.fullName}`
-  ).join(' + ');
-  return `${reactants} ${equationSign(reaction.reversible)} ${products}`;
-}
 
 const sortByName = metabolites => [...metabolites].sort((a, b) => ((a.name > b.name) ? 1 : -1));
 
--- a/frontend/src/helpers/utils.js
+++ b/frontend/src/helpers/utils.js
@@ -71,7 +71,7 @@ export const formatCompartmentStr = (reaction) => {
   /** Extract the compartements from the full names, discard
     * duplicates and return as a string  */
   function uniqueCompartments(xs) {
-    return Array.from(new Set(xs.map(r => r.fullName.match(/\[.*\]/)))).join(' + ');
+    return Array.from(new Set(xs.map(r => r.fullName.match(/\[.*\]/)[0]))).join(' + ');
   }
   const reactantsCompartments = uniqueCompartments(reactants);
   const productsCompartments = uniqueCompartments(products);
--- a/frontend/src/helpers/utils.js
+++ b/frontend/src/helpers/utils.js
@@ -71,7 +71,7 @@ export const formatCompartmentStr = (reaction) => {
   /** Extract the compartements from the full names, discard
     * duplicates and return as a string  */
   function uniqueCompartments(xs) {
-    return Array.from(new Set(xs.map(r => r.fullName.match(/\[.*\]/)[0]))).join(' + ');
+    return Array.from(new Set(xs.map(r => r.fullName.match(/\[[a-z]{1,3}\]/)[0]))).join(' + ');
   }
   const reactantsCompartments = uniqueCompartments(reactants);
   const productsCompartments = uniqueCompartments(products);
--- a/frontend/src/helpers/utils.js
+++ b/frontend/src/helpers/utils.js
@@ -51,7 +51,8 @@ export function getSimpleEquation(reaction) {
 const sortByName = metabolites => [...metabolites].sort((a, b) => ((a.name > b.name) ? 1 : -1));
 
 /** Get the compartement from a reactant or product */
-const getCompartment = component => component.fullName.match(/\[[a-z]{1,3}\]/)[0];
+const getCompartment = ({ fullName }) => fullName.match(/\[[a-z]{1,3}\]/)[0];
+
 
 
 /** Create  the compartements for the summary, as used in Equation and Related Reactions */
--- a/frontend/src/helpers/utils.js
+++ b/frontend/src/helpers/utils.js
@@ -92,11 +92,15 @@ export function reformatChemicalReactionHTML(reaction, noLink = false, model = '
   const type = 'metabolite';
   function formatReactionElement(x) {
     if (!addComp) {
-      return `${Math.abs(x.stoichiometry) !== 1 ? x.stoichiometry : ''} ${noLink ? x.name : buildCustomLink({ model, type, id: x.id, cssClass: x.id === sourceMet ? 'cms' : undefined, title: x.name })}`;
+      return `${Math.abs(x.stoichiometry) !== 1 ? x.stoichiometry : ''}
+              ${noLink ? x.name : buildCustomLink({ model, type, id: x.id, cssClass: x.id === sourceMet ? 'cms' : undefined, title: x.name })}`;
     }
     const regex = /.+\[([a-z]{1,3})\]$/;
     const match = regex.exec(x.fullName);
-    return `${Math.abs(x.stoichiometry !== 1) ? x.stoichiometry : ''} ${noLink ? x.name : buildCustomLink({ model, type, id: x.id, cssClass: x.id === sourceMet ? 'cms' : undefined, title: x.name })}<span class="sc" title="${x.compartment}">${match[1]}</span>`;
+    return `${Math.abs(x.stoichiometry !== 1) ? x.stoichiometry : ''}
+            ${noLink ? x.name : buildCustomLink({ model, type, id: x.id, cssClass: x.id === sourceMet ? 'cms' : undefined, title: x.name })}
+            <span title="${x.compartment}">[ ${match[1]} ] </span>
+            `;
   }
 
   const reactants = sortByName(reaction.reactants).map(formatReactionElement).join(' + ');
--- a/frontend/src/helpers/utils.js
+++ b/frontend/src/helpers/utils.js
@@ -53,6 +53,9 @@ const sortByName = metabolites => [...metabolites].sort((a, b) => ((a.name > b.n
 /** Get the compartement from a reactant or product */
 const getCompartment = ({ fullName }) => fullName.match(/\[[a-z]{1,3}\]/)[0];
 
+/** Extract the compartements from the full names,
+  * discard duplicates and return as a string  */
+const uniqueCompartments = xs => Array.from(new Set(xs.map(r => getCompartment(r)))).join(' + ');
 
 
 /** Create  the compartements for the summary, as used in Equation and Related Reactions */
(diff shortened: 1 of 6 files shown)


inghylt:
Nice work! As seen on the image, some compartments are on the form `[ c ]` while others are without spaces between the brackets and compartment, i.e. `[c]`. I guess we want them all formatted the same way?

Also, I wonder why the Reaction Table should not be changed. Are we fixing that in another issue?

frontend/src/helpers/utils.js
```diff
@@ -62,26 +62,52 @@ export function getChemicalReaction(reaction) {
 
 const sortByName = metabolites => [...metabolites].sort((a, b) => ((a.name > b.name) ? 1 : -1));
 
+
+/** Create  the compartements for the summary, as used in Equation and Related Reactions */
+export const formatCompartmentStr = (reaction) => {
+  const reactants = reaction.metabolites.filter(m => m.outgoing);
+  const products = reaction.metabolites.filter(m => !m.outgoing);
+
+  function fix(xs) {
```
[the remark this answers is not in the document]
MalinAhlberg: Haha, yes, forgot to change that. Will do! :+1:

MalinAhlberg:
> Nice work! As seen on the image, some compartments are on the form `[ c ]` while others are without spaces between the brackets and compartment, i.e. `[c]`. I guess we want them all formatted the same way?

Ah, well spotted! I will fix that.

> Also, I wonder why the Reaction Table should not be changed. Are we fixing that in another issue?

I just assumed it shouldn't be changed. If it should, we might as well do it in this PR I think. Are there more places where this should apply?

inghylt:
Yeah makes sense to fix it here as well. I'm not aware of any more places, @mihai-sysbio might have some input there.

When working on the Reaction table, note this comment from the issue, https://github.com/MetabolicAtlas/MetabolicAtlas/issues/681#issuecomment-954763413. The Reaction table and column 2 in it's exported tsv file should ideally have the same equation

MalinAhlberg:
>  some compartments are on the form [ c ] while others are without spaces between the brackets and compartment, i.e. [c]

Which way do you prefer, spaces or not? (maybe a question for @mihai-sysbio ?)

mihai-sysbio:
> > some compartments are on the form [ c ] while others are without spaces between the brackets and compartment, i.e. [c]
> 
> Which way do you prefer, spaces or not? (maybe a question for @mihai-sysbio ?)

Should we take all UI-related things in https://app.zenhub.com/workspace/o/metabolicatlas/private-issues/issues/116 ? It would simplify this PR.

e0:
> > > some compartments are on the form [ c ] while others are without spaces between the brackets and compartment, i.e. [c]
> > 
> > 
> > Which way do you prefer, spaces or not? (maybe a question for @mihai-sysbio ?)
> 
> Should we take all UI-related things in https://app.zenhub.com/workspace/o/metabolicatlas/private-issues/issues/116 ? It would simplify this PR.

That make sense. Just want to add that for now please feel free to go with either one, as long as it's consistent. It would be easier to change later in the new issue if necessary.

frontend/src/helpers/utils.js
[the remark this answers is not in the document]
MalinAhlberg: Solved by 42b7fdbc03c125c06e2ff1b1fd627d1c4fb2b41c

MalinAhlberg:
> > > > some compartments are on the form [ c ] while others are without spaces between the brackets and compartment, i.e. [c]
> > > 
> > > 
> > > Which way do you prefer, spaces or not? (maybe a question for @mihai-sysbio ?)
> > 
> > 
> > Should we take all UI-related things in https://app.zenhub.com/workspace/o/metabolicatlas/private-issues/issues/116 ? It would simplify this PR.
> 
> That make sense. Just want to add that for now please feel free to go with either one, as long as it's consistent. It would be easier to change later in the new issue if necessary.

I went for no spaces. Feel free to create new issues or comment on the linked one if you disagree :)

e0:
Thanks for adding the changes @MalinAhlberg, I will try to do a more thorough review now.

MalinAhlberg:
> Yeah makes sense to fix it here as well. I'm not aware of any more places, @mihai-sysbio might have some input there.
> 
> When working on the Reaction table, note this comment from the issue, [#681 (comment)](https://github.com/MetabolicAtlas/MetabolicAtlas/issues/681#issuecomment-954763413). The Reaction table and column 2 in it's exported tsv file should ideally have the same equation

Now the equation, related reaction, reaction table and the tsv are produced by the same code, so that we can better avoid ending up with different equations in different places. I had to do some refactoring to achieve this, and I think more is needed, but maybe someone with a better knowledge of js and the code base should have a say at this.

Note that the summary is printed at all these places now, just to be sure that they actually are correct.

frontend/src/helpers/utils.js
```diff
@@ -47,41 +47,58 @@ export function getSimpleEquation(reaction) {
...
-  const products = reaction.products.map(
-    x => `${x.stoichiometry !== 1 ? `${x.stoichiometry} ` : ''}${x.fullName}`
-  ).join(' + ');
-  return `${reactants} ${equationSign(reaction.reversible)} ${products}`;
-}
 
 const sortByName = metabolites => [...metabolites].sort((a, b) => ((a.name > b.name) ? 1 : -1));
 
+/** Get the compartement from a reactant or product */
+const getCompartment = component => component.fullName.match(/\[[a-z]{1,3}\]/)[0];
```
e0, line 54: Minor change suggestion about using [object destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#object_destructuring).

```suggestion
const getCompartment = ({ fullName }) => fullName.match(/\[[a-z]{1,3}\]/)[0];

```

frontend/src/helpers/utils.js
```diff
@@ -47,41 +47,58 @@ export function getSimpleEquation(reaction) {
...
+/** Create  the compartements for the summary, as used in Equation and Related Reactions */
+export const formatCompartmentStr = (reaction) => {
+  const reactants = reaction.metabolites.filter(m => m.outgoing);
+  const products = reaction.metabolites.filter(m => !m.outgoing);
+
+  /** Extract the compartements from the full names,
+    * discard duplicates and return as a string  */
+  function uniqueCompartments(xs) {
+    return Array.from(new Set(xs.map(r => getCompartment(r)))).join(' + ');
+  }
```
e0, line 66: Personally I would prefer to have this defined outside (and use the `const = uniqueCompartments = (xs) => ...` syntax) so it could be reused in other places too. I'm not sure if there is any best practice guidelines about this though.  I know we have existing examples of nested function definitions already though, maybe it's something worth discussing about as a group.

e0 (approved):
Really good job @MalinAhlberg! I left a few minor suggestions/comments but they are not needed. I think this is a great fix and could be merged.

frontend/src/helpers/utils.js
MalinAhlberg: cool :O

frontend/src/helpers/utils.js
MalinAhlberg: I initially put this outside, but then thought there might be practice not to, as it's not used anywhere else (yet), and I noticed other inner functions in the code. But if you think it's better on its own, then I'll move it!
Also, if we have any guidelines about when to use `const` and when to use `function`, that would be nice.

inghylt: @e0 @MalinAhlberg should we create an issue regarding discussing/documenting const vs function?

inghylt:
image
 @JonathanRob Does this look valid to you? I'm not really sure. I'm mainly thinking about the : [m] in the end

JonathanRob:
@inghylt I guess the [m] at the end is the summary of compartments involved in the reaction, which in this case is just the one.

I don't really see the purpose of this additional compartment information after the reaction equation and would recommend removing it altogether, unless there are some reasons supporting its inclusion?

e.g., `glucose[s] => glucose[c] : [s] => [c]` would just be `glucose[s] => glucose[c]`.

inghylt:
@JonathanRob  Thanks for responding so quickly! Good to know. I talked to @mihai-sysbio and he wasn't sure about keeping it either. So @MalinAhlberg probably it makes sense to remove the summary of components involved?

inghylt:
image

By chance I just discovered another place where we use the equations. Do we want the compartments to be shown here as well? @mihai-sysbio

mihai-sysbio:
> Do we want the compartments to be shown here as well? @mihai-sysbio

For consistency, I would say yes. However, the left sidebar is already very crowded, so for a cleaner UI my view is that it should not be added. I would even go further, to remove the coefficients as well (those numbers next to the metabolites do not provide so much value here compared to the Reaction page).

inghylt:
@mihai-sysbio Cool. I'll create another issue for that then

frontend/src/helpers/utils.js
MalinAhlberg: yes, or should it go as a subitem on the linting card?

inghylt: @MalinAhlberg good idea!

MalinAhlberg:
Another question: in reaction MAR09222 (`/explore/Human-GEM/gem-browser/reaction/MAR09222`) there are no products. I assume the equation sign should be removed as well in cases like this?

2021-11-09-103354_429x153_scrot

mihai-sysbio:
Nice that you spotted this @inghylt . These cases are to be expected, so there is an issue to address the formatting https://github.com/MetabolicAtlas/private-issues/issues/96

JonathanRob:
@MalinAhlberg I would recommend leaving the equation sign there, even though there are no products. I realize this looks strange, but it's actually informative to see whether the reactants or products are missing, since some reactions may be missing reactants rather than products (I don't think Human-GEM has any such cases, but other models do).

MalinAhlberg:
@JonathanRob ok, thanks for your quick response! Good to know :+1:

MalinAhlberg:
The summary now removed (by 8119d1c0323309ee06fcf68998fc90bf43d19098). 
2021-11-09-105311_499x134_scrot

inghylt (approved):
Nice work!

mihai-sysbio (approved):
looks good to me!

Issue #745, opened 2021-11-03
Inconsistent and sometimes nonfunctional url for Cytosol


inghylt:
**Describe the bug**
The url when moving between 2D and 3D map(s) for Cytosol, the url is not consistent, which can lead to broken map viewer functionality

**To reproduce**
1. Visit `/explore/Human-GEM/gem-browser/reaction/MAR10743`
2. Press Cytosol 3D
3. Note that the url is `/explore/Human-GEM/map-viewer/cytosol?dim=3d`
4. Press 2D
5. Note that the url is `/explore/Human-GEM/map-viewer/cytosol?dim=2d` (seems like it is Cytosol_4 showing)
6. Reload the current page (Ctrl-r)
7. Note that you can't click on 3D
8. Now visit `/explore/Human-GEM/gem-browser/reaction/MAR10743` again
9. Press 2D Part 1
10. Note that the url is `explore/Human-GEM/map-viewer/cytosol_1?dim=2d`
11. Press 3D
12. Note that the url is `explore/Human-GEM/map-viewer/cytosol_1?dim=3d`

**Expected behavior**
That users can go from either 3D to 2D and vice versa and that the url reflect the current page and does not lead to broken map virw

**Screenshots**
![image.png](https://images.zenhubusercontent.com/5fc0f648ba2981921764866b/07913056-2a6f-4cc1-90aa-a66d8fca836d)

PR #746, opened 2021-11-03
fix: non existing 2d map when going back to map viewer

### LINKED ISSUE #737: Can't access 3D map and inconsistent error message in Map Viewer
**Describe the bug**
When going from non existing 2D map information to the gem browser and then pressing backwards, you can't press 3D to access the 3D map and another type of error message than standard is shown

**To reproduce**
1. Visit /explore/Fruitfly-GEM/gem-browser/reaction/MAR06749
2. Press `Phenylalanine, tyrosine and tryptophan biosynthesis 3D`
3. Press 2D
4. Press gem browser link (subsystem or reaction does not matter)
5. Press backwards button
6. Note the different error message and try pressing 3D button

**Expected behavior**
That pressing backwards button leads to a normally functioning map viewer where you can click 3D and it is the same error message as in step 3.

**Screenshots**

inghylt:
This PR closes #737

#### Type of change
 
#### List of changes made
- If visiting `/explore/:model/map-viewer/:map_id?` where the map_id only has a associated 3D map and not a 2D map as well, the map viewer will not be broken. Instead, now you can still click "Switch to 3D".
- If there is no 2D map, there will be no summary of missing ID:s shown in the Sidebar Panel in the 2D view
- If there is no 2D, the search bar is disabled for the 2D view

### Testing
1. Visit /explore/Fruitfly-GEM/gem-browser/reaction/MAR06749
2. Press Phenylalanine, tyrosine and tryptophan biosynthesis 3D
3. Press 2D
4. Press gem browser link (subsystem or reaction does not matter)
5. Press backwards button
6. Note that the error message is same as in step 3 and that you can click on Switch to 3D

But please play around as well with 2D and 3D for other maps and going from gem browser components to the map viewer to see that nothing is broken/weird.

Note that a `map_id` that does not belong to 2D or 3D maps will still have the same error message and non clickable 3D button as before, e.g. `/explore/Fruitfly-GEM/map-viewer/phenylalanine_tyrosine_and_tryptophan_biosynthesisxxx?dim=2d`

#### Question
During step 2-3, there is a section in the Sidebar Panel for both 2D and 3D called Reaction: MAR06749. During step 6, this is gone. It will occur again if you press 3D again, and still be there if you press 2D, but will disappear if you follow 4-6 once more. I'm not sure if this needs to be fixed or not. Tried looking in to it but did not find a quick fix. If a fix is needed, should it always disappear when there is no 2D map?

#### Discovered bug
#745 was discovered when working on this issue.

### Checklist:
- [x] My code follows the [style guidelines](https://github.com/NBISweden/development-guidelines)
- [x] I have made corresponding changes to the documentation

### DIFF
--- a/frontend/src/components/explorer/MapViewer.vue
+++ b/frontend/src/components/explorer/MapViewer.vue
@@ -218,7 +218,7 @@ export default {
         for (let i = 0; i < categories.length; i += 1) {
           for (let j = 0; j < items[i].length; j += 1) {
             const item = items[i][j];
-            if (this.showing2D) {
+            if (this.showing2D && item.svgs.length > 0) {
               for (let k = 0; k < item.svgs.length; k += 1) {
                 if (item.svgs[k].id === id) {
                   this.currentMap = { ...item };
--- a/frontend/src/components/explorer/MapViewer.vue
+++ b/frontend/src/components/explorer/MapViewer.vue
@@ -235,9 +235,11 @@ export default {
               this.currentMap = item;
               this.currentMap.type = categories[i].slice(0, -1);
               this.mapNotFound = false;
-              this.currentMap.mapReactionIdSet = item.svgs;
-              this.setMapReactionList();
-              this.setMissingReactionList();
+              if (item.svgs.length > 0) {
+                this.currentMap.mapReactionIdSet = item.svgs;
+                this.setMapReactionList();
+                this.setMissingReactionList();
+              }
               return;
             }
             this.mapNotFound = true;
--- a/frontend/src/components/explorer/mapViewer/MapSearch.vue
+++ b/frontend/src/components/explorer/mapViewer/MapSearch.vue
@@ -5,7 +5,7 @@
              :class="searchInputClass"
              title="Exact search by id, name, alias. Press Enter for results" class="input"
              placeholder="Exact search by id, name, alias"
-             :disabled="loading"
+             :disabled="loading || (showing2D && !svgMap)"
              :value="searchTerm"
              type="text" @input="e => handleChange(e.target.value)"
              @keyup.enter="e => search(e.target.value)" />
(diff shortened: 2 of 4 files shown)


MalinAhlberg (approved):
Seems to work perfectly!

I noted the following weird behavior, though, maybe it's related to this bug?
- Navigate to Explore
-  select Zebrafish and Map Viewer
- try to click anywhere, choose a map or switch to 3D. Nothing works.

mihai-sysbio:
> should we also make the 3D the default view for models that don't have 2D maps (or maybe all models)?

It might be that this is the intention right now - to disable the `2D/3D` button since there are no 2D maps. But for some reason the URL gets stuck on the 2D mode. A new issue would keep things from growing out of scope.

MalinAhlberg:
> A new issue would keep things from growing out of scope.

I created issue #756 for this.

e0 (approved):
Looks great @inghylt! Really elegant changes to fix this issue.

nanjiangshu (commented):
Well done @inghylt!

There is only one problem I found. When visiting `/explore/Fruitfly-GEM/map-viewer/phenylalanine_tyrosine_and_tryptophan_biosynthesisxxx?dim=2d`, the message says "probably there is a typo in the map identifier in the URL. Use the list of maps to find other maps". However, since the 2D maps for `phenylalanine_tyrosine_and_tryptophan_biosynthesisxxx` does not exist, all entries on the map listing table are disabled. I think this is related to the newly created issue #756, but not exactly the same. I think other entries on the map listing table should still be clickable even when the current 2D map does not exist.

inghylt:
@nanjiangshu Really good catch! I think however fixing that should belong to another issue (#756 probably) since
- this behavior is not introduced by this PR
- this PR mainly focuses on keeping the error message consistent and the functionality of switching between 2D and 3D when a user have gone to another page and then back to the map viewer

Does that make sense to you as well?

nanjiangshu:
@inghylt It make sense to address it in another issue.

nanjiangshu (approved):

inghylt:
@nanjiangshu Cool! Added what you wrote as a comment to #756

Issue #689, opened 2021-09-17
fix CodeQL warnings in svgThumbnail.js


inghylt:
As a developer, I would like the CodeQl warnings to be fixed in `api/src/endpoints/svgThumbnail.js`, so that I don't have to worry or have them reported when making PR:s

PR #753, opened 2021-11-08
feat: use both w and h for thumbnails

### LINKED ISSUE #752: Add h param to API thumbnail
As an API user, I want choose between specifying a height or a width to the map thumbnail call so that I can choose what looks best in my use case.

**Describe the solution you'd like**
To be able to specify the `height` param like the currently implemented `width` (either or).

mihai-sysbio:
This PR closes #752.

#### Type of change
 
#### List of changes made
There is now support for a `height` param instead of/in addition to the `width`.

### Testing
Use this url to test
> /api/v2/svg-thumbnail/acylglycerides_metabolism?model=Human-GEM&height=200

And feel free to try invalid heights, and/or the `width` param.

### Further comments
If relevant, specify questions or related information

# Checklist:
- [ ] My code follows the [style guidelines](https://github.com/NBISweden/development-guidelines)
- [ ] I have performed a self-review of my own code


mihai-sysbio:
Regarding the CodeQL warning, it's a leftover from the previous implementation, meant to be resolved via #689.

api/src/endpoints/index.js
```diff
@@ -1,7 +1,6 @@
 import express from 'express';
 import neo4jRoutes from 'endpoints/neo4j';
 import repoRoutes from 'endpoints/repository';
-import hpaRoutes from 'endpoints/hpaRna';
```
e0, line 4: This looks like a safe removal but not relevant to the linked issue. I specifically wanted to mention it here because we have associated code in the `frontend` that can also be removed.

e0 (commented):
The `height` parameter seems to work well. I wanted to discuss the resizing logic and validation feedback before approving.

It looks like there are some other nice fixes included in this PR that are not related to the linked issue. It would be nice if we could refrain from including commits that are out of scope.

api/src/endpoints/index.js
mihai-sysbio: To my understanding, this removal was meant to be part of the epic https://github.com/MetabolicAtlas/private-issues/issues/124, specifically issue https://github.com/MetabolicAtlas/MetabolicAtlas/pull/699 or https://github.com/MetabolicAtlas/data-generation/pull/23 so this is meant just to fix the message shown in `logs api`.

api/src/endpoints/svgThumbnail.js
```diff
@@ -7,10 +7,25 @@ const VALID_MODELS = models.map((m) => m.short_name);
 
 const routes = express.Router();
 
+function validateDimension(dimensionString) {
+  const minDim = 100;
+  const maxDim = 600;
+  if(dimensionString){
+    if (!/^\d+$/.test(dimensionString)) {
+      throw new Error(`Invalid dimension provided: ${dimensionString}. The width and/or height parameters should be an integer between ${minDim} and ${maxDim}.`);
```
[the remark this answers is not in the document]
mihai-sysbio: Sure, I'm up for reconsidering.
The `and/or` is meant to indicate that providing only one of the params is also acceptable.

api/src/endpoints/svgThumbnail.js
```diff
@@ -7,10 +7,25 @@ const VALID_MODELS = models.map((m) => m.short_name);
 
 const routes = express.Router();
 
+function validateDimension(dimensionString) {
```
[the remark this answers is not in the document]
mihai-sysbio: Good idea, the renaming is implemented in ad5af79 .

mihai-sysbio:
I can see how the frontend fix is outside the scope. The other one though.. would we want to consider that out of scope too, as a reference for other issues?

api/src/endpoints/index.js
e0: It's true that we no longer need it but I wanted to mention it since the frontend has some files that should be removed too along with this, for example: https://github.com/MetabolicAtlas/MetabolicAtlas/blob/develop/frontend/src/store/modules/humanProteinAtlas.js

api/src/endpoints/svgThumbnail.js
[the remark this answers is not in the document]
e0: Sorry I should have clarified, I feel like we should go with one of these options:
- use `or` since it only uses `width` or `height` depending on which is less
- use `and/or` but change the resize logic to something like `fill` so that it uses both `width` and `height`

api/src/endpoints/svgThumbnail.js
```diff
@@ -7,10 +7,25 @@ const VALID_MODELS = models.map((m) => m.short_name);
...
+function getDimension(dimensionString) {
+  const minDim = 100;
+  const maxDim = 600;
+  if(dimensionString){
+    if (!/^\d+$/.test(dimensionString)) {
+      throw new Error(`Invalid dimension provided: ${dimensionString}. The width and/or height parameters should be an integer between ${minDim} and ${maxDim}.`);
+    } else {
+      var myDimension = parseInt(dimensionString) || null;
+      if ((myDimension < minDim || myDimension > maxDim)){
+        throw new Error(`Invalid dimension provided: ${dimensionString}. The width and/or height parameters should be an integer between ${minDim} and ${maxDim}.`);
```
inghylt, line 19: This error message is repeated twice. Would we like to create a file similar to `frontend/src/content/messages.js` and define it there?

mihai-sysbio: I'm not sure how the variable inclusion would work - those strings are static.

frontend/src/components/explorer/gemBrowser/Reaction.vue
```diff
@@ -106,7 +106,7 @@ export default {
     }
 
     const title = `${this.reaction.id}, Reaction in ${this.model.short_name}`;
-    const description = `The reaction ${this.reaction.id} in ${this.model.short_name} (version ${this.model.version}) can be found in the ${this.reaction.compartments.[0].name} compartment and the ${this.reaction.subsystems[0].name} subsystem.`;
+    const description = `The reaction ${this.reaction.id} in ${this.model.short_name} (version ${this.model.version}) can be found in the ${this.reaction.compartments[0].name} compartment and the ${this.reaction.subsystems[0].name} subsystem.`;
```
inghylt, line 109: 1. This feels like a random change to include in this PR
2. How would you suggest to test this change?

mihai-sysbio: 1. It was reported by `logs frontend` so yeah pretty random.
2. The `description` field non-visible html. To test, I inspect the page and look for `description`.

api/src/endpoints/svgThumbnail.js
inghylt: We could define a function that takes  `dimensionString`,  `minDim`, and `maxDim` as arguments, either here or in a `messages.js` similar file. But I'm fine with keeping it as is as well

mihai-sysbio: It's a good point, but I think we should have a separate issue for this. `messages.js` in under `/frontend`, whereas this is under `/api`, and once we have a way of dealing with arguments in strings then I guess a lot more strings could be migrated to `messages.js`.
I'm leaving this conversation open for others to join. Also, if there is a new issue for this, it would be good to have it linked to this conversation.

frontend/src/components/explorer/gemBrowser/Reaction.vue
inghylt: It is weird, I can't see any difference in the description when comparing between branch develop and this branch. But the recent changes do look more correct code wise.

However, I did notice though that for `/explore/Human-GEM/gem-browser/reaction/MAR12870`, the description is:

"The reaction MAR12870 in Human-GEM (version 1.8.0) can be found in the Extracellular compartment and the Drug metabolism subsystem."

I.e, Cytosol is not listed. I guess we should create an issue for this?

Would also like to add that I think that unrelated changes that require testing to be reviewed should be in a separate PR, with a description on how to test the changes

api/src/endpoints/svgThumbnail.js
inghylt: Just to clarify, the suggestion was not to use file `frontend/src/content/messages.js`, but to create a file with similar purpose in the api folder

api/src/endpoints/index.js
nanjiangshu: I agree with @shan-chalmers that the removal of this file is irrelevant to this issue. But we can keep the change this time.

api/src/endpoints/svgThumbnail.js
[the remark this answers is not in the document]
mihai-sysbio: [`fill`](https://sharp.pixelplumbing.com/api-resize#resize) cannot be used because it changes the aspect ratio. I went for `inside`, but I guess `contain` and `cover` are equally good. We could point to this documentation in the error message.

frontend/src/components/explorer/gemBrowser/Reaction.vue
mihai-sysbio: To me, I corrected a typo in accordance with the recommendation from `logs frontend`. I didn't think this would change the code, or require review. I guess the lesson learnt is to open issues for typos.
Yes please a new issue for this.

mihai-sysbio:
Thanks for the review @inghylt !

>     1. I'm not really sure what the user requirements are. I feel like it is a bit weird that if I visit for example: `/api/v2/svg-thumbnail/acylglycerides_metabolism?model=Human-GEM&height=600&width=100` the height param is ignored. But maybe the users of this functionality only cares about specifying either height or width and do not wish to control both at the same time? If they would like to control both, a suggestion would be to use `fit: 'fill'` and if only one param is specified we could calculate the dimension ratio and through that provide the non specified param to the `resizeParams` object.

This is what `inside` does. It might be easier to continue the discussion [the comment thread](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/753#discussion_r744845996). 
 
>     2. If I visit `/api/v2/svg-thumbnail/acylglycerides_metabolism?model=Human-GEM`, a rather large map is returned, but we have 600 as a limit for `height` and `width`. What is the reasoning behind the limit?

Good thinking. The limit is to stop unnecessary computation for little benefit - after a certain size, the user might as well control the image size themselves.

api/src/endpoints/svgThumbnail.js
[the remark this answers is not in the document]
mihai-sysbio: @nanjiangshu based on your comment I would guess you're advocating for `'contain'`.

api/src/endpoints/svgThumbnail.js
e0: I think it's a good idea to have a new issue for backend error messages. For now I do think it would be nice to refactor the function and use a variable for the two instances of the same error message.

mihai-sysbio: ebc272d combines the 2 error messages into 1.

frontend/src/components/explorer/gemBrowser/Reaction.vue
inghylt: Makes sense! To me a typo that does not require a separate PR would be for example spelling errors, e.g. `The top table pf the page` to  `The top table of the page`, or fixing indentation. Changes that do not affect the logic of the code. This is a very small change but still it changes how properties of an object is accessed (right?). But yeah I would also have assumed that `logs frontend` would only report non functional changes.

Created issue #754 for the problem mentioned above

mihai-sysbio: Hmm I see your point. Here's how `logs frontend` looks without the fix (eg in `develop`):

```
frontend  |  ERROR  Failed to compile with 1 error4:18:56 PM
frontend  | 
frontend  |  error  in ./src/components/explorer/gemBrowser/Reaction.vue?vue&type=script&lang=js&
frontend  | 
frontend  | Syntax Error: Unexpected token (109:167)
frontend  | 
frontend  |   107 |
frontend  |   108 |     const title = `${this.reaction.id}, Reaction in ${this.model.short_name}`;
frontend  | > 109 |     const description = `The reaction ${this.reaction.id} in ${this.model.short_name} (version ${this.model.version}) can be found in the ${this.reaction.compartments.[0].name} compartment and the ${this.reaction.subsystems[0].name} subsystem.`;
frontend  |       |                                                                                                                                                                        ^
frontend  |   110 |
frontend  |   111 |     return {
frontend  |   112 |       title,
```

I do agree that it was surprising to see this in the review. Perhaps I should have included the error above in the PR. But yeah next time it would be more straightforward to fix this via a different issue.

api/src/endpoints/svgThumbnail.js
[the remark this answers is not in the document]
mihai-sysbio: Great! That leaves us only `cover`.

api/src/endpoints/svgThumbnail.js
inghylt: Nice commit! Created issue #755 for having a common file for error messages

frontend/src/components/explorer/gemBrowser/Reaction.vue
e0: That's interesting, I don't get any problems when I run `logs frontend`. @mihai-sysbio you mentioned you did a clean install so maybe some packages are more updated than the rest of the project and it's picking up different errors?

mihai-sysbio: Could very well be. Just by looking at `this.reaction.compartments.[0].name` and `${this.reaction.subsystems[0].name` side by side still makes me think of a (code-)typo.

api/src/endpoints/svgThumbnail.js
```diff
@@ -19,20 +30,12 @@ routes.get('/:svgName', async (req, res) => {
...
-      w_thumb =  parseInt(width);
-      if (w_thumb < min_w || w_thumb > max_w){
-        throw new Error(`Invalid width provided: ${width}. It should be between ${min_w} and ${max_w}.`);
-      }
-    }
-    const svgThumbnail = await getSvgThumbnail(svgName, w_thumb, model)
+    const resizeParams = {
+      width: getDimension(width),
+      height: getDimension(height),
+      fit: 'inside',
```
mihai-sysbio, line 36:
```suggestion
      fit: 'cover',
```
@e0 @nanjiangshu @inghylt what do you think?

e0: That is exactly what I was looking for and it solves the `and/or` thing I brought up before as well.

inghylt: If you specify both width and height, the map is cropped off for heights > 330, even if width is set to 600. But I'm not really sure of how the users will use this so if you feel like this is what we need, go for it

mihai-sysbio:
This PR has been updated as per the discussions above (`cover` instead of `inside`) - please verify if they are resolved.

When combined with the `fix/font-family` branch from `data-files` (as per [this PR](https://github.com/MetabolicAtlas/data-files/pull/22)), the latest commit also resolves the `Fontconfig` problem that was not detected before.

inghylt (approved):

api/src/endpoints/svgThumbnail.js
mihai-sysbio: To me this more of a back-up option, using `width` or `height` separately are not sufficient.

MalinAhlberg:
Nice work!

Apart from the very thorough previous discussions, I think it looks good!

As usual, a related question: is there any specific reason why `/api/v2/svg-thumbnail/cytosol_2?model=Human-GEM&width=100`  returns nothing (but still 200), while `/api/v2/svg-thumbnail/katt?model=Human-GEM&width=100` returns an error? This behavior is different on the prod server.

e0:
> As usual, a related question: is there any specific reason why `/api/v2/svg-thumbnail/cytosol_2?model=Human-GEM&width=100` returns nothing (but still 200), while `/api/v2/svg-thumbnail/katt?model=Human-GEM&width=100` returns an error? This behavior is different on the prod server.

Interesting, I can confirm this is the case and also same thing for `cytosol_4`. Looking at the files, both of these are over 3MB so looks like there might be a hard limit somewhere. Also for `cytosol_3`, which is 1.6MB, it took 20+ seconds for me to process and load the image for me locally. It would be good to test the images for larger maps if we were to increase the limit.

mihai-sysbio:
> /api/v2/svg-thumbnail/cytosol_2?model=Human-GEM&width=100

Super interesting, for the maps `mitocondria`, `cytosol_2`, `cytosol_4` I get this image 
image

mihai-sysbio:
Commit 815d4c2 removes the input limit that prevented thumbnails to be generated for the large maps.

e0:
> Commit [815d4c2](https://github.com/MetabolicAtlas/MetabolicAtlas/commit/815d4c200f3be68b60a662d72af9766d61aa330c) removes the input limit that prevented thumbnails to be generated for the large maps.

This seems to work. For `/api/v2/svg-thumbnail/mitochondria?model=Human-GEM&height=300`, it took 1.3min to process and load. It would be good to test this on the dev server and see if it manages to load within the nginx timeout.

MalinAhlberg (approved):

e0 (approved):
I'm approving this for now since it's an urgent issue that needs to be merged and deployed soon. The following two problems should be addressed in the future. If everyone is in agreement I can create issues for them.

1. Text not rendering.
2. It takes a long time to process and load maps, especially larger ones.

PR #751, opened 2021-11-05
2.1

mihai-sysbio:
This minor release, as a `tock` release, is focused on improving the usability of the website. Some new features were added, while existing features were re-implemented and expanded. The changes are summarized below.

#### New featuress
- the _Identifier_ page is set up for all the cross-references and the Met Atlas identifiers #708 #750 
- model component counts will be always up-to-date #724
- API endpoint for map thumbnails #677 #752 

#### Feature re-implemenentation
- Data Overlay
    - it is now supporting different data types (metabolomics in addition to transcriptomics)
    - the same interface is shared between the Map Viewer and the Interaction partners #723

##### Map Viewer
- #658
- #659 
- #666 
- #695 
- #727
##### Interaction Partners 
- #660
- #703
##### GEM Browser
- #692
- #665
- show compartments in the reaction equations #743
##### Other
- #714
- visual improvement of the Advisors section #661 
- #716
- for Human-GEM, include previous (non-MA) ids in search #700
- #722

#### Misc
- #699
- #701
- #711
- #715
- #720

#### Non-code improvements
- #712 
- #713
- #717

> _Note: This release relies on https://github.com/MetabolicAtlas/data-generation/tree/e28d3034d46b6dd8cbfe4df487c8e00b132aff6d and https://github.com/MetabolicAtlas/data-files/commit/d577b55abd6b4c6650baec183c519aa8da6df7a7 ._

PR #757, opened 2021-11-09
chore: 2.1 news

mihai-sysbio:
This PR prepares the news for the 2.1 release, fixes broken links in the news section and reverts the Docker image for the `frontend` container so that it is compatible with the old `node-sass` in `package.json`. It's meant as a "last PR" prior to merging #751 .

### DIFF
--- a/frontend/Dockerfile
+++ b/frontend/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:lts-alpine
+FROM node:14-alpine
 
 WORKDIR /project
 COPY . .
--- a/frontend/src/content/news.json
+++ b/frontend/src/content/news.json
@@ -2,7 +2,7 @@
   "2021": [
     {
       "date": "2021-11-09",
-      "title": "Metabolic Atlas 2.1 revamps data overlay",
+      "title": "Metabolic Atlas 2.1 revamps Data overlay",
       "text": "<b>Metabolic Atlas 2.1</b> is out. In this minor release we focused on improving the usability of the website, either through feature re-implementation, as was the case for the <i>Data overlay</i>, or through independent fixes. Moreover, several small new features were added, for example the <i>Identifier</i> pages. More details can be found <a href='https://github.com/MetabolicAtlas/MetabolicAtlas/releases/tag/2.1' target='_blank'>on GitHub</a> or <a href='https://zenodo.org/badge/latestdoi/53664497' target='_blank'>on Zenodo</a>."
     },
     {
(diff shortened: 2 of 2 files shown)


MalinAhlberg (approved):
Looks good to me! :v:

frontend/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:lts-alpine
+FROM node:14-alpine
```
nanjiangshu, line 1: I'm curious about the reason to make this change.

frontend/src/content/news.json
```diff
@@ -1,9 +1,14 @@
 {
   "2021": [
+    {
+      "date": "2021-11-09",
+      "title": "Metabolic Atlas 2.1 revamps data overlay",
+      "text": "<b>Metabolic Atlas 2.1</b> is out. In this minor release we focused on improving the usability of the website, either through feature re-implementation, as was the case for the <i>Data overlay</i>, or through independent fixes. Moreover, several small new features were added, for example the <i>Identifier</i> pages. More details can be found <a href='https://github.com/MetabolicAtlas/MetabolicAtlas/releases/tag/2.1' target='_blank'>on GitHub</a> or <a href='https://zenodo.org/badge/latestdoi/53664497' target='_blank'>on Zenodo</a>."
```
nanjiangshu, line 6: `Data overlay` has been used at multiple places in this file but with different format, might be good to make it consistent.

nanjiangshu (commented):
Looks great @mihai-sysbio! Just a few minor comments.

frontend/Dockerfile
mihai-sysbio: The version of `node-sass` specified in the `package.json` in the `devDependencies` requires Node 14, according to the [documentation](https://www.npmjs.com/package/node-sass). Most likely `node:lts` was at version 14 until recently, but the now `node:lts-alpine` points to version version 16, according to [Dockerhub](https://hub.docker.com/_/node/).

frontend/src/content/news.json
mihai-sysbio: Thanks, should be resolved in 0986877

frontend/src/content/news.json
```diff
@@ -1,9 +1,14 @@
...
     },
```
nanjiangshu, line 12: For links to `zenodo`, would it be possible to link to the specific release? I realized that both version 2.0 and 2.1 link to the same place.

mihai-sysbio: A specific `doi` is obtained only after the release on GitHub, so the way to fix this would be to create an issue and update the previous links. The current link points to the latest release that was archived by Zenodo.

frontend/Dockerfile
nanjiangshu: I see. Thanks for the explanation.

mihai-sysbio:
@nanjiangshu please continue the review, even though I have to merge and deploy. We can create new issues from what you discover here.

mihai-sysbio:
I already had the fix, and the news was easy enough to write, so a `1`.

PR #758, opened 2021-11-09
feat: improve reaction formatting

MalinAhlberg:
This PR closes [private issue #116]( https://app.zenhub.com/workspaces/metatlas-sprint-607745622d49260019ce115a/issues/metabolicatlas/private-issues/116) and relates to #743.

#### Type of change

 
#### List of changes made
- Line break between related reactions
- Superscripting `+` and `-` in names

#### Screenshot of the fix
superscript
Also `-` is superscripted, as can be seen here (O2-):
minus

### Testing
See eg `/explore/Human-GEM/gem-browser/reaction/MAR02041`. 
This fix affects reactions in multiple places (reaction equation, related reactions, reactions in the search results...), but should not change the `Reaction table` tsv.
To see the change in the search result, search for eg `MAR01303` in the quick search.
It would be very good if someone familiar with the api could dobule check that everything looks good (`reaction.js` and `search.js` ~are affected by this~ uses this code, but not with the parameter `html`, so they are not affected).

### Further comments
If you approve of this change, maybe it should be applied to even more places, eg in metabolites like `/explore/Human-GEM/gem-browser/metabolite/MAM02039m` ?

# Checklist:

### DIFF
--- a/frontend/src/components/explorer/gemBrowser/Reaction.vue
+++ b/frontend/src/components/explorer/gemBrowser/Reaction.vue
@@ -56,7 +56,7 @@
                 {{ rr.id }}
               </router-link>:&nbsp;
               <span v-html="reformatChemicalReactionHTML(
-                {reaction: rr, noLink : true, model : model.short_name, comp : true})"/>
+                {reaction: rr, noLink : true, model : model.short_name, comp : true})" />
             </p>
           </td>
         </tr>
--- a/frontend/src/components/explorer/gemBrowser/Reaction.vue
+++ b/frontend/src/components/explorer/gemBrowser/Reaction.vue
@@ -51,13 +51,13 @@
         <tr v-if="relatedReactions.length !== 0">
           <td class="td-key has-background-primary has-text-white-bis">Related reaction(s)</td>
           <td>
-            <span v-for="rr in relatedReactions" :key="rr.id">
+            <p v-for="rr in relatedReactions" :key="rr.id">
               <router-link :to="{ name: 'reaction', params: { model: model.short_name, id: rr.id } }">
                 {{ rr.id }}
               </router-link>:&nbsp;
               <span v-html="reformatChemicalReactionHTML(
-                {reaction: rr, noLink : true, model : model.short_name, comp : true})"></span>
-            </span>
+                {reaction: rr, noLink : true, model : model.short_name, comp : true})"/>
+            </p>
           </td>
         </tr>
       </table>
--- a/frontend/src/helpers/utils.js
+++ b/frontend/src/helpers/utils.js
@@ -86,7 +86,7 @@ export function reformatChemicalReactionHTML({ reaction, model, noLink = false, 
   const type = 'metabolite';
   const stoichiometry = x => (Math.abs(x.stoichiometry) !== 1 ? `${x.stoichiometry} ` : '');
   // use super- and subscript for charge and number of atoms, if in html mode
-  const chemName = ({ name, charge }) => (html ? chemicalFormula(name.replace(/([-+])$/, ''), charge) : name);
+  const chemName = ({ name }) => (html ? chemicalFormula(name, null) : name);
   const getComponentName = x => (noLink ? chemName(x) : buildCustomLink({ model, type, id: x.id, cssClass: x.id === sourceMet ? 'cms' : undefined, title: chemName(x) }));
 
   function formatReactionElement(x) {
--- a/frontend/src/helpers/utils.js
+++ b/frontend/src/helpers/utils.js
@@ -83,7 +83,9 @@ export function reformatChemicalReactionHTML({ reaction, model, noLink = false, 
   const addComp = comp || reaction.compartment_str.includes('=>');
   const type = 'metabolite';
   const stoichiometry = x => (Math.abs(x.stoichiometry) !== 1 ? `${x.stoichiometry} ` : '');
-  const getComponentName = x => (noLink ? x.name : buildCustomLink({ model, type, id: x.id, cssClass: x.id === sourceMet ? 'cms' : undefined, title: x.name }));
+  // superscript + and - at the end of names
+  const supName = ({ name }) => (html ? name.replace(/([-+])$/, '<sup>$1</sup>') : name);
+  const getComponentName = x => (noLink ? supName(x) : buildCustomLink({ model, type, id: x.id, cssClass: x.id === sourceMet ? 'cms' : undefined, title: supName(x) }));
 
   function formatReactionElement(x) {
     if (!addComp) {
--- a/frontend/src/helpers/utils.js
+++ b/frontend/src/helpers/utils.js
@@ -53,6 +53,7 @@ const sortByName = metabolites => [...metabolites].sort((a, b) => ((a.name > b.n
 /** Get the compartement from a reactant or product */
 const getCompartment = ({ fullName }) => fullName.match(/\[[a-z]{1,3}\]/)[0];
 
+
 /** Extract the compartements from the full names,
   * discard duplicates and return as a string  */
 const uniqueCompartments = xs => Array.from(new Set(xs.map(r => getCompartment(r)))).join(' + ');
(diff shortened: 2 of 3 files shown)


frontend/src/helpers/utils.js
```diff
@@ -82,14 +83,18 @@ export function reformatChemicalReactionHTML({ reaction, model, noLink = false,
   const addComp = comp || reaction.compartment_str.includes('=>');
   const type = 'metabolite';
   const stoichiometry = x => (Math.abs(x.stoichiometry) !== 1 ? `${x.stoichiometry} ` : '');
+  // superscript + and - at the end of names
```
nanjiangshu, line 86: Great that you added superscript for the charge of ions. But I think you may reuse the [chemicalFormula](https://github.com/MetabolicAtlas/MetabolicAtlas/blob/develop/frontend/src/helpers/chemical-formatters.js#L1). With this function, chemicals such as `O2` will also be formatted as O<sub>2</sub>

frontend/src/helpers/utils.js
```diff
@@ -82,14 +83,18 @@ export function reformatChemicalReactionHTML({ reaction, model, noLink = false,
...
+  const supName = ({ name }) => (html ? name.replace(/([-+])$/, '<sup>$1</sup>') : name);
+  const getComponentName = x => (noLink ? supName(x) : buildCustomLink({ model, type, id: x.id, cssClass: x.id === sourceMet ? 'cms' : undefined, title: supName(x) }));
```
nanjiangshu, line 88: Great code refactoring to remove the redundancy!

nanjiangshu (commented):
Great work @MalinAhlberg. I just point out that the code to convert the expression of the charge of ions can reuse the already existing function `chemicalFormula`.

inghylt:
Really nice job! Will approve once Nanjiangs comment is solved

If someone wants to look at a reaction with multiple related reaction after the update to Human-GEM 1.10.0, visit for example `/explore/Human-GEM/gem-browser/reaction/MAR03936`

I can't give good input regarding applying the change to other places such as metabolites. Probably @mihai-sysbio or any of the scientific advisors might have an idea

frontend/src/helpers/utils.js
mihai-sysbio: ➕

frontend/src/components/explorer/gemBrowser/Reaction.vue
```diff
@@ -51,13 +51,13 @@
...
           <td>
-            <span v-for="rr in relatedReactions" :key="rr.id">
+            <p v-for="rr in relatedReactions" :key="rr.id">
               <router-link :to="{ name: 'reaction', params: { model: model.short_name, id: rr.id } }">
                 {{ rr.id }}
               </router-link>:&nbsp;
               <span v-html="reformatChemicalReactionHTML(
-                {reaction: rr, noLink : true, model : model.short_name, comp : true})"></span>
-            </span>
+                {reaction: rr, noLink : true, model : model.short_name, comp : true})"/>
```
mihai-sysbio, line 59: As suggested by `logs frontend`
```suggestion
                {reaction: rr, noLink : true, model : model.short_name, comp : true})" />
```

frontend/src/helpers/utils.js
mihai-sysbio: This is an awesome fix 🌟  It resolves my previous comment that it can be weird with `H+ [c] +`.

mihai-sysbio (changes requested):
Great job!
It feels odd to request a change for a typo-level warning, otherwise I agree with the previous reviews.

frontend/src/helpers/utils.js
MalinAhlberg: Thanks for the tip, @nanjiangshu ! :+1:  Fixed in f3d70fdf0ab015d2d602b20ff0756a992deb6f51.

frontend/src/components/explorer/gemBrowser/Reaction.vue
MalinAhlberg: Fixed in b73a2fcb05cf1d7e9c6a51a885c700fbc68e3ad2

MalinAhlberg:
`chemicalFormula` is now used, as suggested by @nanjiangshu . 
chemformula

MalinAhlberg:
Should we do anything with formatting of metabolites (address in this pr, open new pr..)? The `name` and `related metabolites` in the picture would be nicer with some formatting, no?
metabolite-formatting

nanjiangshu:
Hi @MalinAhlberg , great that you have implemented with the function `ChemicalFormula`. However, it also cased bad formatting with e.g. `5-oxo-12-HETE`. I would suggest adding a regular expression checking of such names (mostly metabolite names) to avoid wrong formatting.

e0:
> Hi @MalinAhlberg , great that you have implemented with the function `ChemicalFormula`. However, it also cased bad formatting with e.g. `5-oxo-12-HETE`. I would suggest adding a regular expression checking of such names (mostly metabolite names) to avoid wrong formatting.
> 
> My suggestion of change:
> 
> ```js
> export default function chemicalFormula(formula, charge) {
>   if (formula === null || formula === undefined) {
>     return '';
>   }
>   if (/-\w/g.test(formula)) {
>     return formula;
>   } 
>   let form = formula.replace(/([0-9])/g, '<sub>$1</sub>');
>   if (charge) {
>     form = `${form}<sup>${Math.abs(charge) !== 1 ? Math.abs(charge) : ''}${charge > 0 ? '+' : '-'}</sup>`;
>   } 
>   return form;
> } 
> ```
> 
> What do you think @e0 and @inghylt?

Thanks for including the suggestion @nanjiangshu. Can you please help me pinpoint how the formatting for this example has changed? For example when visiting `/explore/Human-GEM/gem-browser/metabolite/MAM01123m`, I didn't noticed any differences between the two functions.

inghylt:
@nanjiangshu I like that suggestion, fixes what currently looks off in `/explore/Human-GEM/gem-browser/reaction/MAR03936
`
I have a related question. When comparing `localhost/explore/Human-GEM/gem-browser/reaction/MAR04335` with `https://metabolicatlas.org/explore/Human-GEM/gem-browser/reaction/MAR04335`, a lot of the items in the equation has a number next to their charge in localhost. Is that legit?

MalinAhlberg:
> And @MalinAhlberg regarding formatting of metabolites, I would suggest to open a new issue.

I created #773 for this.

nanjiangshu:
> @nanjiangshu I like that suggestion, fixes what currently looks off in `/explore/Human-GEM/gem-browser/reaction/MAR03936 `
> I have a related question. When comparing `localhost/explore/Human-GEM/gem-browser/reaction/MAR04335` with `https://metabolicatlas.org/explore/Human-GEM/gem-browser/reaction/MAR04335`, a lot of the items in the equation has a number next to their charge in localhost. Is that legit?

Hi @inghylt, I think the current formatting is better than on `https://metabolicatlas.org/`. For the example you mentioned `/explore/Human-GEM/gem-browser/reaction/MAR04335`, the charge for `NADP` is `-3`, in the new formatting, it shows 
as NPDP<sup>3-</sup>, while in the production server, it shows `NADP+`. I guess in the previous formatting, charges are not used unless it is included in the name.

MalinAhlberg:
Thanks for the code, @nanjiangshu :star: !

And also thanks for the testing, @inghylt and @e0 !
I don't have the proper knowledge to answer any questions about how the charge should be shown, but I agree with @nanjiangshu that the current output looks closer to the data, where the `charge` of `NPDP` is `-3`.

MalinAhlberg:
@JonathanRob it would be very valuable to hear your opinion on this. Does the formatting of the reaction equations in the pictures below look ok to you?

formatting1
formatting2
(apart from the line breaks second pic)

JonathanRob:
@MalinAhlberg this looks really nice, though I'm undecided whether it is good to show the formal charges (i.e., the values in the `metCharges` field) of all metabolites. I like that it presents more information in the equation, but it may become confusing in cases such as NADP<sup>+</sup> mentioned above. Although chemists may appreciate the actual formal charge being shown, biochemists/biologists are more familiar with the notation NADPH and NADP<sup>+</sup>, rather than NADPH<sup>4-</sup> and NADP<sup>3-</sup>.

I'm leaning toward the option of only showing the charge if it's included in the name (i.e., NADP<sup>+</sup>), but am open to the version that you have shown which presents all charges.

nanjiangshu:
> @MalinAhlberg this looks really nice, though I'm undecided whether it is good to show the formal charges (i.e., the values in the `metCharges` field) of all metabolites. I like that it presents more information in the equation, but it may become confusing in cases such as NADP+ mentioned above. Although chemists may appreciate the actual formal charge being shown, biochemists/biologists are more familiar with the notation NADPH and NADP+, rather than NADPH4- and NADP3-.
> 
> I'm leaning toward the option of only showing the charge if it's included in the name (i.e., NADP+), but am open to the version that you have shown which presents all charges.

Thanks @JonathanRob for your comment. I think I'm now more inclined to the notation NADPH or NADP+ than with the actual charge, since I would guess the majority of users are biochemists/biologists instead of chemists. 

@MalinAhlberg, to implement it, I think a quick fix would be just set `charge` as null when calling the function `chemicalFormula`.

MalinAhlberg:
> Thanks @JonathanRob for your comment. I think I'm now more inclined to the notation NADPH or NADP+ than with the actual charge, since I would guess the majority of users are biochemists/biologists instead of chemists.
> 
> @MalinAhlberg, to implement it, I think a quick fix would be just set `charge` as null when calling the function `chemicalFormula`.

Yes, thanks for your input @JonathanRob!

@nanjiangshu: to achieve having NADP<sup>+</sup> (rather than NADP+), some more modifications to the code is needed. I could add this `else` clause
```
  if (charge) {
    form = `${form}<sup>${Math.abs(charge) !== 1 ? Math.abs(charge) : ''}${charge > 0 ? '+' : '-'}</sup>`;
  } else {
    form = form.replace(/([-+])$/, '<sup>$1</sup>');
  }
```
to `chemicalFormula`. Do you think that makes sense also for other usage of this function? Otherwise I can add that part to the the caller function (`reformatChemicalReactionHTML`).

nanjiangshu:
Nice fix! I think it will be useful at other places as well.

MalinAhlberg:
Ok, the current (final?) version is this:
formatting-final
finalformatting2

nanjiangshu (approved):
Great!

mihai-sysbio (approved):

PR #763, opened 2021-11-12
fix: update model in the store after model change

### LINKED ISSUE #734: Reactions table not being populated for genes from quick search
**Describe the bug**
The Reactions table is sometimes empty when clicking on a matching gene in quick search. When doing a hard refresh for the url, the reaction table is populated

**To reproduce**
1. Search for `EEF2` in Human-GEM the quick search, choose matching gene
2. Press the matching result
3. Look at the Reactions table
4. Hard refresh

**Expected behavior**
The reaction table always being populated

**Screenshots**
When the match first clicked, the reaction table is not populated

nanjiangshu:
This PR closes #734 

**Type of change**  

**List of changes made**  

model in the vue store is updated after model change in the quick search

**Fixed buggy behavior**  

Two buggy behaviors described in the issue #734 are fixed.

**Further comments**  

**Checklist**  

### DIFF
--- a/frontend/src/components/explorer/gemBrowser/GemSearch.vue
+++ b/frontend/src/components/explorer/gemBrowser/GemSearch.vue
@@ -43,10 +43,10 @@
             <hr v-if="i2 !== 0" class="m-0">
             <div class="is-clickable">
               <span v-if="type === 'metabolite' || type === 'gene'" class="pr-1">
-                <span class="color-gb is-clickable" @mousedown="handleClickResult(type, r)">
+                <span class="has-text-primary is-clickable" @mousedown="handleClickResult(type, r)">
                   <span class="icon is-medium is-left" title="Gem Browser"><i class="fa fa-table" /></span>
                 </span>
-                <span class="color-ip is-clickable" @mousedown="handleClickResult('interaction', r)">
+                <span class="has-text-icon-interaction-partner is-clickable" @mousedown="handleClickResult('interaction', r)">
                   <span class="icon is-medium is-left" title="Interaction Partners"><i class="fa fa-connectdevelop" /></span>
                 </span>
               </span>
@@ -302,14 +302,6 @@ export default {
         > span:last-child {
           flex-grow: 1;
         }
-      }
-
-      .color-gb {
-        color: $primary;
-      }
-
-      .color-ip {
-        color: $icon-interaction-partner;
       }
     }
   }
--- a/frontend/src/components/explorer/gemBrowser/GemSearch.vue
+++ b/frontend/src/components/explorer/gemBrowser/GemSearch.vue
@@ -157,6 +157,7 @@ export default {
         this.searchModel = this.models[modelKey];
         await this.searchDebounce(this.searchTermString);
       }
+      this.$store.dispatch('models/selectModel', modelKey);
     },
     async searchDebounce(searchTerm) {
       this.noResult = false;
--- a/frontend/src/components/explorer/gemBrowser/GemSearch.vue
+++ b/frontend/src/components/explorer/gemBrowser/GemSearch.vue
@@ -41,7 +41,7 @@
         <template v-for="type in resultsOrder">
           <div v-for="(r, i2) in searchResults[type]" :key="`${r.id}-${i2}`" class="searchResultSection px-1 py-0">
             <hr v-if="i2 !== 0" class="m-0">
-            <div class="is-clickable">
+            <div>
               <span v-if="type === 'metabolite' || type === 'gene'" class="pr-1">
                 <span class="has-text-primary is-clickable" @mousedown="handleClickResult(type, r)">
                   <span class="icon is-medium is-left" title="Gem Browser"><i class="fa fa-table" /></span>
@@ -50,7 +50,7 @@
                   <span class="icon is-medium is-left" title="Interaction Partners"><i class="fa fa-connectdevelop" /></span>
                 </span>
               </span>
-              <span class="has-text-link" @mousedown="handleClickResult(type, r)">
+              <span class="has-text-link is-clickable" @mousedown="handleClickResult(type, r)">
                 <b class="is-capitalized">{{ type }}: </b>
                 <label class="is-clickable" v-html="formatSearchResultLabel(type, r, searchTermString)"></label>
               </span>
@@ -293,7 +293,6 @@ export default {
     .searchResultSection {
       > div {
         display: flex;
-        padding: 7px 0px;
         align-items: center;
         white-space: nowrap;
         overflow: hidden;
@@ -301,6 +300,7 @@ export default {
 
         > span:last-child {
           flex-grow: 1;
+          padding: 11px 0px;
         }
       }
     }
--- a/frontend/src/components/explorer/gemBrowser/GemSearch.vue
+++ b/frontend/src/components/explorer/gemBrowser/GemSearch.vue
@@ -304,7 +304,7 @@ export default {
       }
 
       .color-ip {
-        color: #A15786;
+        color: $icon-interaction-partner;
       }
     }
   }
--- a/frontend/src/components/explorer/gemBrowser/GemSearch.vue
+++ b/frontend/src/components/explorer/gemBrowser/GemSearch.vue
@@ -41,7 +41,7 @@
         <template v-for="type in resultsOrder">
           <div v-for="(r, i2) in searchResults[type]" :key="`${r.id}-${i2}`" class="searchResultSection px-1 py-0">
             <hr v-if="i2 !== 0" class="m-0">
-            <div py-2>
+            <div class="is-clickable">
               <span v-if="type === 'metabolite' || type === 'gene'" class="pr-1">
                 <span class="color-gb is-clickable" @mousedown="handleClickResult(type, r)">
                   <span class="icon is-medium is-left" title="Gem Browser"><i class="fa fa-table" /></span>
@@ -292,11 +292,16 @@ export default {
 
     .searchResultSection {
       > div {
-        display: block;
+        display: flex;
         padding: 7px 0px;
+        align-items: center;
         white-space: nowrap;
         overflow: hidden;
         text-overflow: ellipsis;
+
+        > span:last-child {
+          flex-grow: 1;
+        }
       }
 
       .color-gb {
--- a/frontend/src/components/explorer/gemBrowser/GemSearch.vue
+++ b/frontend/src/components/explorer/gemBrowser/GemSearch.vue
@@ -41,24 +41,20 @@
         <template v-for="type in resultsOrder">
           <div v-for="(r, i2) in searchResults[type]" :key="`${r.id}-${i2}`" class="searchResultSection px-1 py-0">
             <hr v-if="i2 !== 0" class="m-0">
-            <router-link class="is-clickable"
-                         :to="{ name: type, params: { model: searchModel.short_name, id: r.id } }"
-                         @click.native="handleClickResult">
-              <span v-if="type === 'metabolite' || type === 'gene'" class="search-result-icons pr-1">
-                <router-link class="is-clickable"
-                             :to="{ name: type, params: { model: searchModel.short_name, id: r.id } }"
-                             @click.native="handleClickResult">
+            <div py-2>
+              <span v-if="type === 'metabolite' || type === 'gene'" class="pr-1">
+                <span class="color-gb is-clickable" @mousedown="handleClickResult(type, r)">
                   <span class="icon is-medium is-left" title="Gem Browser"><i class="fa fa-table" /></span>
-                </router-link>
-                <router-link class="is-clickable"
-                             :to="{ name: 'interaction', params: { model: searchModel.short_name, id: r.id } }"
-                             @click.native="handleClickResult">
+                </span>
+                <span class="color-ip is-clickable" @mousedown="handleClickResult('interaction', r)">
                   <span class="icon is-medium is-left" title="Interaction Partners"><i class="fa fa-connectdevelop" /></span>
-                </router-link>
+                </span>
               </span>
-              <b class="is-capitalized">{{ type }}: </b>
-              <label class="is-clickable" v-html="formatSearchResultLabel(type, r, searchTermString)"></label>
-            </router-link>
+              <span class="has-text-link" @mousedown="handleClickResult(type, r)">
+                <b class="is-capitalized">{{ type }}: </b>
+                <label class="is-clickable" v-html="formatSearchResultLabel(type, r, searchTermString)"></label>
+              </span>
+            </div>
           </div>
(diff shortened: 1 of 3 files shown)


inghylt:
Nice fix!

I'm curious, why do we have the model in both store (line 118) and as a data property in the GemBrowser component (line 113)? Why do we not only use the value in store? @e0 do you know perhaps?

e0:
> Nice work!
> 
> ~I'm curious, why do we have the model in both store (line 118) and as a data property in the GemBrowser component (line 113)? Why do we not only use the value in store? @e0 do you know perhaps?~ Ok so it seems that we separated the model value between store and the GemSearch data property in commit [79dbc2b](https://github.com/MetabolicAtlas/MetabolicAtlas/commit/79dbc2bf588e2f1243b49bfdc5d1a5ca2f7e1e00), which closed an issue with description "Move GemSearch to the menu bar". With this fix, they are not separated as clearly as before. So if you change the model in the GemSearch, the model used for the rest of the page is also updated. This was not the case before, try out for example:
> 
> 1. Visit root
> 2. Click on GemSearch
> 3. Change to Zebrafish-GEM
> 4. Press Explore
> 
> On both localhost and metabolicatlas.org. On localhost, the ticked box will be Zebrafish-GEM but on metabolicatlas.org it will be Fruitfly-GEM.
> 
> Note that if you on localhost visit `/explore` as step 4 instead of pressing the Explore button or do a page refresh as step 5, the model is still Fruitfly-GEM which I feel is quite inconsistent.
> 
> Not sure how we want this to work. Probably wise to discuss at standup today

Thanks for looking into this @inghylt. I think the behavior on metabolicatlas.org is correct. Imagine that you are on a gem browser components page (for example `/explore/Human-GEM/gem-browser/reaction/MAR09329`). Now, if you go to the GemSearch and change to `Yeast-GEM`, the model in the store should not immediately change to `Yeast-GEM`. We only want the store model to change to `Yeast-GEM` after the user has searched AND clicked on one of the search results.

e0 (changes requested):
Based on the [earlier discussion](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/763#issuecomment-968717807), I don't think we can implement this change. The model in the store should not update immediately after selecting an option in the models dropdown. It could lead to the example in the screenshot, where the model in the store (`Yeast-GEM`) and the model being displayed  (on the page and in the route) are out of sync.

image

nanjiangshu:
Thanks @inghylt and @e0 for pointing out this problem. I agree with you that updating of the model in the store should happen after clicking a hit in the GemSearch.

nanjiangshu:
Hi @inghylt and @e0, I've made a new commit to fix the problem you previously mentioned. 
For some reason, with `router-link`, function after page re-directing was run before the function linked to the click event. 
As suggested by Shan, now I'm using <span> instead of router-link and make the re-directing happening inside the function `handleClickResult`.

mihai-sysbio:
> For some reason, with `router-link`, function after page re-directing was run before the function linked to the click event. As suggested by Shan, now I'm using instead of router-link and make the re-directing happening inside the function `handleClickResult`.

Could you dig deeper into what is the root of the problem here? We are using `router-link` all throughout the app, so we cannot accept "random" misbehavior. And if it's really a problem, we should then make sure the fix is available everywhere (even if problems haven't been reported yet).

frontend/src/components/explorer/gemBrowser/GemSearch.vue
```diff
@@ -292,23 +290,21 @@ export default {
...
         text-overflow: ellipsis;
       }
 
-      .search-result-icons a {
-        &:first-child {
-          color: $primary;
-        }
+      .color-gb {
+        color: $primary;
+      }
```
mihai-sysbio, line 304: It looks like this could be replaced by `is-primary` or something of the sort.

frontend/src/components/explorer/gemBrowser/GemSearch.vue
```diff
@@ -292,23 +290,21 @@ export default {
...
 
-        &:nth-child(2) {
-          color: #A15786;
-        }
+      .color-ip {
+        color: #A15786;
```
mihai-sysbio, line 307: Since this color is shared with `GemSearch.vue`, maybe it would make sense to define it in the CSS vars.

nanjiangshu:
> > For some reason, with `router-link`, function after page re-directing was run before the function linked to the click event. As suggested by Shan, now I'm using instead of router-link and make the re-directing happening inside the function `handleClickResult`.
> 
> Could you dig deeper into what is the root of the problem here? We are using `router-link` all throughout the app, so we cannot accept "random" misbehavior. And if it's really a problem, we should then make sure the fix is available everywhere (even if problems haven't been reported yet).

I think we'd better to create a new issue to investigate it.

frontend/src/components/explorer/gemBrowser/GemSearch.vue
nanjiangshu: I was trying to keep the styling the same as before. The color of the Gem-Browser icon is not the same by using `is-primary`.

frontend/src/components/explorer/gemBrowser/GemSearch.vue
nanjiangshu: > Since this color is shared with `GemSearch.vue`, maybe it would make sense to define it in the CSS vars.

Do you mean define it under the root of `style` in this file? It seems this coloring, i.e. `$primary` for the icon Gem-Browser and `#A15786` for the icon IP is set only at this place and at `InteractionPartners.vue`. All the rest places, such as the Map Viewer and Documentation, another set of colors are used. Should we make them consistent? As I can see, current coloring at this search window looks quite nice and I guess there was some effort to define these colors when this feature was introduced.

frontend/src/components/explorer/gemBrowser/GemSearch.vue
mihai-sysbio: This sounds very confusing to me - do you mean `is-primary` shows a different color than `color: $primary`?

frontend/src/components/explorer/gemBrowser/GemSearch.vue
mihai-sysbio: Most (all?) colors used in the codebase are defined in this file https://github.com/MetabolicAtlas/MetabolicAtlas/blob/develop/frontend/src/style/vars.scss

Searching for `#A15786` yields 2 matches in the codebase, in `GemSearch.vue` and `InteractionPartners.vue`. So instead of defining the same value, it would make sense to me to make this color available in the entire codebase by adding it to `vars.scss`.

frontend/src/components/explorer/gemBrowser/GemSearch.vue
nanjiangshu: yes

frontend/src/components/explorer/gemBrowser/GemSearch.vue
nanjiangshu: Make sense to me. I will also leave a note in the private issue [#84](https://app.zenhub.com/workspaces/metatlas-roadmap-60d96b68da0fb10019d26deb/issues/metabolicatlas/private-issues/84) for refactoring of the color definition for IP.

nanjiangshu: refactored with the commit 92dda32

inghylt:
@nanjiangshu The new changes seems to fix what I noted previously :+1:  However, currently on metabolicatlas.org the entire row for a result is clickable. With these changes you have to click on for example `ENSG00000167658 - EEF2` specifically to visit the result when searching for gene EEF2

e0:
Hi @nanjiangshu, sorry that I didn't give this more time before. When I responded to your questions before I didn't take into consideration of the context and I probably gave you some bad advice.

I looked into the problem in more depth now and tested the behavior a few times. I took at look at the `ReactionTable` logic to see if there was anything that can be done there since that's where the error happens. Here is an alternative solution as you see in the screenshot.

In `beforeMount`, we only call `setup` (fetch data from the API) if the model in the store and the model in the route are the same. This is the case when you switch from one `Human-GEM` component to another for example. The cases where this `if` condition doesn't satisfy is what triggered our problem before, like searching for a `Human-GEM` gene and clicking on it when the current model in the store is something else. In this case, the `beforeMount` will now just ignore it and we have a new function inside `watch` that monitors the changes. Here we compare the new model and the old one, for example when switching from `Yeast-GEM` to `Human-GEM`, and we call `setup` only if they are different.

**TLDR: When model is same, call `setup` from `beforeMount`. When model changes, call `setup` from `watch`.**

image

While testing this I noticed another bug. Not sure if we already have an issue for this. But you can do the following to trigger it:

1. Go to homepage, click gem search, select `Human-GEM` and type `eef-2`.
2. Click on the first gene. If you followed above changes this page should now work ok including the related reactions.
3. Go to gem search, change the model to `Worm-GEM`.
4. Click on the first gene. This now throws an error because the model has not been updated.

I took a quick look at this problem as well and it looks like the following change (diff and code) does the trick. The only problem is that at step 4, it seems that the relate reactions endpoint is being called twice. The first one is the gene id from the previous page, and the second one is the gene id from the new page. This doesn't actually seem to be a problem for now for the end user but of course it would be nice to fix it too. I don't have more time to look into this now but hopefully someone else can try to take a look at it if you like this alternative solution.

image

image

nanjiangshu:
Thanks very much @e0 for looks so deep into this issue and figured out an alternative solution. I've tested your solution in the branch https://github.com/MetabolicAtlas/MetabolicAtlas/tree/test/gemsearch-bug-solution-2 and it seems the bug has been solved 🌟 . As to the second bug you mentioned, it has been mentioned in the latest comment for [this issue](https://app.zenhub.com/workspaces/metatlas-sprint-607745622d49260019ce115a/issues/metabolicatlas/metabolicatlas/734). However, the problem with the router-link as I posted in the issue [#152](https://app.zenhub.com/workspaces/metatlas-sprint-607745622d49260019ce115a/issues/metabolicatlas/metabolicatlas/734) still remains. If you try out the branch `gemsearch-bug-solution-2`, in the console you will see the following output, the setup function in the `componentLayout` runs before the `handleClickResult`.
Screenshot 2021-11-16 at 19 31 29

Your alternative solution is to check in every child objects whether the model has been updated and update it if not. However, I think it is more logical that model change happens directly after mouse clicking.

Anyway, for the solution I implemented in the branch `fix/empty-reaction-table-quicksearch`, I do stuck at the CSS problem pointed out by @inghylt  (https://github.com/MetabolicAtlas/MetabolicAtlas/pull/763#issuecomment-970278638). I want the  <span> with the text in the search result window expand to fill the rest of the empty space but failed. If I use 
```css
{
  display:block;
  width: 100%;
}
```
the text do fill the full width but will start from a new line and thus break from the icons as shown by the screenshot. 
Screenshot 2021-11-16 at 19 57 07

with your solution, there is no such problem though.

e0:
Sorry again @nanjiangshu, I wanted to look for another solution to see if we can bypass some of the problems being discussed now but failed to give you proper review over your existing solution again.

I think the downside of your solution (moving away from `router-link`) can be overlooked by the fact that it eliminates all of the other problems. You make a good point that the model should be changed here instead of at several places in the child components.

For the CSS problem you mentioned I think the following might give you what you want. I gave it a quick test and it looked ok but please feel free to try it out more thoroughly yourself. Besides making it to take the remaining width (`flex-grow: 1`) of the flex layout, it also moves the padding there so that the click area is larger.

```scss
    .searchResultSection {
      > div {
        display: flex;
        align-items: center;
        white-space: nowrap;
        overflow: hidden;
        text-overflow: ellipsis;

        > span:last-child {
          flex-grow: 1;
          cursor: pointer;
          padding: 7px 0px;
        }
      }
```

nanjiangshu:
Thanks so much @e0 for your tips. I've struggle a while for the CSS but did not know it could be solved with `flex-grow: 1`. I've pushed the commit cff1d6c with small modifications from the code you posted. As to the place for `padding: 7px 0px;`, I kept it at the original place since I think the table looks cleaner with larger line spacing. Maybe we can discuss about the line spacing issue in the standup tomorrow.

nanjiangshu:
> @nanjiangshu The new changes seems to fix what I noted previously 👍 However, currently on metabolicatlas.org the entire row for a result is clickable. With these changes you have to click on for example `ENSG00000167658 - EEF2` specifically to visit the result when searching for gene EEF2

Hi @inghylt, with the help from Shan @e0,  the clickable region problem has been solved in the commit cff1d6c

e0:
> Thanks so much @e0 for your tips. I've struggled a while for the CSS but did not know it could be solved with `flex-grow: 1`. I've pushed the commit [cff1d6c](https://github.com/MetabolicAtlas/MetabolicAtlas/commit/cff1d6c78ce1cf8a27859bf23b6411d4674c0a82) with small modifications from the code you posted. As to the place for `padding: 7px 0px;`, I kept it at the original place since I think the table looks cleaner with larger line spacing. Maybe we can discuss about the line spacing issue in the standup tomorrow.

Looks good! The minor concern I have of not moving the padding from the `div` to `span:last-child` is that the actual clickable area of the `span` has less height now. If a user clicks a few pixels above the `span` nothing would happen. When I tested it didn't make a visual difference to the table, it was just about placing the padding in the `span` instead of the `div`. But if this was not the case for you we could try to take a quick look during the standup like you mentioned.

frontend/src/components/explorer/gemBrowser/GemSearch.vue
mihai-sysbio: Right, the correct Bulma class for this context is `has-text-primary`, which renders the expect `primary` color.

nanjiangshu: Great to learn the syntax of Bulma for setting the text color. While `has-text-primary` does the trick, `has-text-icon-interaction-partner` does not seem to work, although `icon-interaction-partner` is also defined in `frontend/src/style/vars.scss`. Nevertheless, I do think the code is more understandable with the current setting, i.e. using `color-gb` and `color-ip`. In addition, with this setting, if the color of icons needs to be changed, we only need to change the css without touching the template.

mihai-sysbio: [`has-text-primary`](https://bulma.io/documentation/helpers/color-helpers/ is a Bulma class. By default, we should use Bulma classes as much as possible.

It's no surprise that there is no `has-text-icon-interaction-partner`, since these classes are not dynamically defined. The usage of `vars.scss` is just to share or overload the Bulma-provided CSS variables. Feel free to define a global class in `App.vue` for this class.

nanjiangshu: Good to know. I've implemented the usage of Bulma syntax in the commit 94e4fe22

inghylt (approved):
Tried out the changes and I think it looks good! Good job

e0 (approved):

PR #776, opened 2021-11-15
fix: map layout lock

### LINKED ISSUE #762: Transport reactions not loading correctly in 3D
**Describe the bug**
Transport reactions in 3D is not correctly

**To reproduce**
Visit https://metabolicatlas.org/explore/Human-GEM/map-viewer/transport_reactions?dim=3d

**Expected behavior**
That users are able to view the transport reactions network.
Note that when trying out the fix, make sure that it is also functional with search values, e.g https://metabolicatlas.org/explore/Human-GEM/map-viewer/transport_reactions?dim=3d&panel=0&sel=MAR05094&search=MAR05094&coords=-4046.68,-23369.77,0.15,0,0,500&g1=None&g2=None

**Screenshots**

**others**

The server might not be functioning as normal after seeing this error. It will be back to normal after a while (ca 10 min)

e0:
This PR closes #762.

**Type of change**  

**List of changes made**  
Two main changes are included:
1. Use worker process (new thread) to handle 3d network layout generation so that it doesn't lock up the main thread. This was a serious problem since loading any large 3d network would lock up the API for all users. The number of iterations has also been decreased for larger networks. About 3min -> 45s for Human transport reactions on my local machine.
2. Use `JOIN` statements to improve the neo4j query time of fetching data needed for 3d networks. This is now ~4-5 times faster compared to before (~7s -> 1.5s on a cold start for example). Note that even before this change the time taken was still little compared to the time it needed for layout generation.

**Testing**  

- A rebuild is needed due to new dependencies. As mentioned [here](https://github.com/MetabolicAtlas/MetabolicAtlas/issues/762#issuecomment-969193554), the neo4j docker image needed to be upgraded.
1. Visit: `/explore/Human-GEM/map-viewer/transport_reactions?dim=3d`.
2. The 3d network should load within 60s.
3. At any time that the 3d network is loading, open a new tab and go to `/explore/Human-GEM/gem-browser` and note that the random components should load as usual. Try refreshing multiple times as the 3d network is loading to make sure the API is still responsive to other requests throughout the whole process.

**Checklist**  


e0:
Please note that since the number of iterations have been decreased, the older comments in the code which are now obsolete are removed. I haven't tested extensively yet but we may want to create a new issue to further fine tune these numbers and document the new findings.

api/src/workers/3d-network.js
```diff
@@ -0,0 +1,18 @@
+import { Worker } from 'worker_threads'
+
+export default (workerData) => new Promise((resolve, reject) => {
```
MalinAhlberg, line 3: I think the code looks all good, but some comments/documentation would be nice!

MalinAhlberg (approved):
Very good job :star2: !
Works very fast and also well, as far as I can tell.

neo4j/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM neo4j:4.2.5
+FROM neo4j:4.2.11
```
mihai-sysbio, line 1: Since some work has gone into choosing a newer Docker imagine, why not go all the way to the latest `4.3.7`? It would same some time to not go through all the versions incrementally, but perhaps you want a dedicated issue for it?

mihai-sysbio:
Really nice fixes @e0 ! The worker is such a good addition. Do you think the resource allocation (cpu, ram) in the production Dockerfile should be revised?

Also, I want to make a note that the previously high iterations of the layout algorithm were needed in order to obtain a usable layout. There is a separate issue dedicated just to optimizing this: better layout with less processing https://github.com/MetabolicAtlas/private-issues/issues/62 .

e0:
> Really nice fixes @e0 ! The worker is such a good addition. Do you think the resource allocation (cpu, ram) in the production Dockerfile should be revised?
> 
> 
> 
> Also, I want to make a note that the previously high iterations of the layout algorithm were needed in order to obtain a usable layout. There is a separate issue dedicated just to optimizing this: better layout with less processing https://github.com/MetabolicAtlas/private-issues/issues/62 .

Good suggestion to reconsider the resource allocation. I haven't stress tested this too much yet (try to load many large 3D networks at the same time) but it would be good to try that on the dev server and adjust resources if needed. Perhaps it could go into the issue you mentioned. Even if some large networks get layouts that don't look as good now, I think the immediate key is that they load within the timeout. I think it's good if we try to merge and deploy this as soon as possible considering the current problems.

neo4j/Dockerfile
e0: I should have made a comment about this but I didn't get the 4.3.x versions to work locally. I didn't try that much too look into it to be fair. For now this means hopefully less surprises as well since we are staying on 4.2.x.

inghylt (approved):
Great job @e0, I'm impressed that you were able to solve this as quickly/neatly as you did.

api/src/workers/3d-network.js
e0: Thank you for the review and feedback! I have added some comments now.

e0:
> Great job @e0, I'm impressed that you were able to solve this as quickly/neatly as you did.

Thank you!

There are two reviews now but I will try to deploy this to the dev server and try it out a bit there before merging this PR.

e0:
Looks like it went well after testing on the dev server. The larger networks take about 30s to load. On a second tab it doesn't look like new requests are being held up.

e0:
> > On a second tab it doesn't look like new requests are being held up.
> 
> This might be the case if in `dev` nginx caches the request - it certainly does this in `prod`.

The problem [mentioned here](https://github.com/MetabolicAtlas/MetabolicAtlas/issues/762#issuecomment-969193554) was reproducible on dev server and should still be reproducible on prod. I can see nginx solving many of the cases but for the random components it is never cached. I should have clarified here that I was specifically testing the gem browser page with the random components so it could not have been nginx.

PR #777, opened 2021-11-17
fix: use new docker compose commands

mihai-sysbio:
This PR updates the Docker compose syntax to the latest version. Deploying to `dev` went smooth.

### DIFF
--- a/proj.sh
+++ b/proj.sh
@@ -16,17 +16,17 @@ function generate-data {
 
 function build-stack {
   generate-data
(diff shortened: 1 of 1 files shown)


nanjiangshu (approved):
It looks good when testing on Mac. But for my Linux box with Ubuntu 18, I still need to use docker-compose since I'm not able to install the latest Docker.

e0 (approved):
This looks fine but I believe it only affects the local development. The deploy function was already using `docker compose` with `--context`.

PR #786, opened 2021-11-29
fix: switch to 3d

### LINKED ISSUE #756: Not possible to switch to 3D maps
**Describe the bug**
When visiting the map viewer, it's not possible to switch to 3D if there are no 2D maps.

**To reproduce**
- Navigate to Explore
- Select Zebrafish and Map Viewer
- Try to switch to 3D

**Expected behavior**
3D should be the default if there are no 2D maps. The 2D map button should be disabled.

**Additional context**
See comments [here](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/746#issuecomment-963329101).

MalinAhlberg:
This PR closes #756 

Solves the issue of not being able to switch to 3D maps when there are no 2D maps, or when you end up on an invalid map page (eg `/explore/Fruitfly-GEM/map-viewer/phenylalanine_tyrosine_and_tryptophan_biosynthesisxxx?dim=2d`).

**Type of change**  

**List of changes made**  
1. adds a state that keeps track of whether there are available 2D maps
2. defaults to 3D when there are no 2D maps
3. disables the switch buttons when there are no 2D maps
4. allows switching to 3D when visiting a non-existing 2D map
5. allows switching to 3D before choosing map.

**Testing**  
- To test 5:
    - go to `https://metabolicatlas.org/explore/Human-GEM/map-viewer?dim=2d`. Clicking `Switch to 3D` will not do anything
    - on this branch `/explore/Human-GEM/map-viewer?dim=2d`, clicking the same button will refresh the list of available maps
- To test 4, visit `/explore/Fruitfly-GEM/map-viewer/phenylalanine_tyrosine_and_tryptophan_biosynthesisxxx?dim=2d` and change to 3D.
- To test 2&3: visit the Map Viewer for Fruitfly.

**Further comments**  
I'm not very sure about the general design of the state and related functions. 

**Checklist**  

### DIFF
--- a/frontend/src/components/explorer/MapViewer.vue
+++ b/frontend/src/components/explorer/MapViewer.vue
@@ -162,7 +162,7 @@ export default {
       if (this.avail2D) {
         return `Switch to ${this.dimensionalState(!this.showing2D)}`;
       }
-      return 'This model only has 3D maps availabe';
+      return 'This model only has 3D maps available';
     },
   },
   watch: {
--- a/frontend/src/components/explorer/MapViewer.vue
+++ b/frontend/src/components/explorer/MapViewer.vue
@@ -19,7 +19,7 @@
              v-on="sidebarLayoutReset ? { scroll: () => handleSidebarScroll() } : {}">
           <div id="mapSidebar__header" class="has-background-lightgray pt-3">
             <div class="buttons has-addons is-centered padding-mobile m-0"
-                 :title="`Switch to ${dimensionalState(!showing2D) }`"
+                 :title="switchTitle"
                  @click="(!currentMap || (currentMap && currentMap.type !== 'custom'))
                    && $store.dispatch('maps/toggleShowing2D')">
               <button v-for="dim in [true, false]" :key="dim"
@@ -157,6 +157,12 @@ export default {
     }),
     queryParams() {
       return { ...this.mapQueryParams, ...this.dataOverlayQueryParams };
+    },
+    switchTitle() {
+      if (this.avail2D) {
+        return `Switch to ${this.dimensionalState(!this.showing2D)}`;
+      }
+      return 'This model only has 3D maps availabe';
     },
   },
   watch: {
--- a/frontend/src/components/explorer/MapViewer.vue
+++ b/frontend/src/components/explorer/MapViewer.vue
@@ -20,7 +20,8 @@
           <div id="mapSidebar__header" class="has-background-lightgray pt-3">
             <div class="buttons has-addons is-centered padding-mobile m-0"
                  :title="`Switch to ${dimensionalState(!showing2D) }`"
-                 @click="currentMap && currentMap.type !== 'custom' && $store.dispatch('maps/toggleShowing2D')">
+                 @click="(!currentMap || (currentMap && currentMap.type !== 'custom'))
+                   && $store.dispatch('maps/toggleShowing2D')">
               <button v-for="dim in [true, false]" :key="dim"
                       class="button m-0"
                       :class="dim === showing2D ? 'is-selected is-primary has-text-weight-bold' : 'is-light'"
--- a/frontend/src/components/explorer/MapViewer.vue
+++ b/frontend/src/components/explorer/MapViewer.vue
@@ -149,6 +149,7 @@ export default {
       showing2D: state => state.maps.showing2D,
       dataOverlayPanelVisible: state => state.maps.dataOverlayPanelVisible,
       mapsListing: state => state.maps.mapsListing,
+      avail2D: state => state.maps.avail2D,
     }),
     ...mapGetters({
       mapQueryParams: 'maps/queryParams',
--- a/frontend/src/components/explorer/MapViewer.vue
+++ b/frontend/src/components/explorer/MapViewer.vue
@@ -25,7 +25,7 @@
               <button v-for="dim in [true, false]" :key="dim"
                       class="button m-0"
                       :class="dim === showing2D ? 'is-selected is-primary has-text-weight-bold' : 'is-light'"
-                      :disabled="currentMap && currentMap.type === 'custom'">
+                      :disabled=" !avail2D || (currentMap && currentMap.type === 'custom')">
                 <span v-if="dim === showing2D" class="icon">
                   <i class="fa fa-check-square-o"></i>
                 </span>
--- a/frontend/src/store/modules/maps.js
+++ b/frontend/src/store/modules/maps.js
@@ -21,6 +21,7 @@ const data = {
     nodes: [],
     links: [],
   },
+  avail2D: true,
   showing2D: true,
   dataOverlayPanelVisible: false,
   coords: {
@@ -52,9 +53,15 @@ const getters = {
 };
 
 const actions = {
+
   async getMapsListing({ commit }, model) {
     const payload = { model: model.apiName, version: model.apiVersion };
     const mapsListing = await mapsApi.fetchMapsListing(payload);
+    let avail2D = false;
+    Object.keys(mapsListing).forEach((cat) => {
+      avail2D = mapsListing[cat].some(x => x.svgs.length === 1) || avail2D;
+    });
+    commit('setAvail2D', avail2D);
     commit('setMapsListing', mapsListing);
   },
 
@@ -203,6 +210,13 @@ const actions = {
 };
 
 const mutations = {
+  setAvail2D: (state, available) => {
+    if (!available) {
+      state.showing2D = false;
+    }
+    state.avail2D = available;
+  },
+
   setAvailableMaps: (state, maps) => {
     state.availableMaps = maps;
   },
@@ -228,7 +242,7 @@ const mutations = {
   },
 
   setShowing2D: (state, showing2D) => {
-    state.showing2D = showing2D;
+    state.showing2D = showing2D && state.avail2D;
   },
 
   setDataOverlayPanelVisible: (state, dataOverlayPanelVisible) => {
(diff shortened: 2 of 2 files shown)


frontend/src/store/modules/maps.js
```diff
@@ -52,9 +53,15 @@ const getters = {
 };
 
 const actions = {
+
   async getMapsListing({ commit }, model) {
     const payload = { model: model.apiName, version: model.apiVersion };
     const mapsListing = await mapsApi.fetchMapsListing(payload);
+    let avail2D = false;
+    Object.keys(mapsListing).forEach((cat) => {
+      avail2D = mapsListing[cat].some(x => x.svgs.length === 1) || avail2D;
```
inghylt, line 62: I wonder if it makes sense to have `x.svgs.length > 0` since Cytosol has multiple svg:s. But I don't think any model has only Cytosol as 2d map so this should also be fine

frontend/src/store/modules/maps.js
```diff
@@ -52,9 +53,15 @@ const getters = {
 };
 
 const actions = {
+
   async getMapsListing({ commit }, model) {
     const payload = { model: model.apiName, version: model.apiVersion };
     const mapsListing = await mapsApi.fetchMapsListing(payload);
+    let avail2D = false;
+    Object.keys(mapsListing).forEach((cat) => {
```
inghylt, line 61: Out of curiosity, why `cat`?

inghylt:
Nice work! :clap:

I think it makes sense to disable the button switch between 2D and 3D for models without any 2D maps. However, I think it might be nice to make the text displayed when hovering the button explain why it is disabled, maybe something like "This model only have 3D maps available" or something

image

frontend/src/store/modules/maps.js
e0: I think this is a good suggestion. I'm also wondering if we can short-circuit the iteration so we don't have to go through every item. Maybe something like the following. 

```javascript
    // eslint-disable-next-line
    for (const { svgs } of Object.values(mapsListing).flat()) {
      if (svgs.length > 0) {
        avail2D = true;
        break;
      }
    }
```

Please note that I'm not actually sure if this is faster.. both ways are probably fast enough to have any noticeable effect but this makes more sense to me logically.

e0 (approved):
This looks great! Really nice way of setting it to 3D when svgs aren't available.

frontend/src/store/modules/maps.js
MalinAhlberg: I was thinking `category`, the `mapListing` has the keys `compartment maps`, `subsystem maps` etc (but now I had to think for some minutes before remembering why I called it `cat`, so maybe it's not a very good name after all... If you have better suggestions, please let me know :))

frontend/src/store/modules/maps.js
MalinAhlberg: Thanks for the suggestions :)

I more or less copied the `svgs.length === 1` from `MapListings.vue` and didn't dare to change it. But yes, `> 0` sounds better!

I also think that your idea makes much sense @e0. Do you prefer the `for` loop, even though `eslint` gets upset about it?

nanjiangshu (commented):
Really nice fix @MalinAhlberg. I agree with @inghylt that it will be good to add a notification to explain why it is not able to switch to 2D, either in the hover text, e.g. "2D map not available, switch to 2D disabled", or similar pop up text when the users click on the map switching region. However, if you think this is out of scope of this issue. I can also accept it as it is.

MalinAhlberg:
> I think it makes sense to disable the button switch between 2D and 3D for models without any 2D maps. However, I think it might be nice to make the text displayed when hovering the button explain why it is disabled, maybe something like "This model only have 3D maps available" or something
> 
> image

Good idea! Added in a5c6719e24c7639dafed0d5c7b4a7304b9ec56c7

frontend/src/store/modules/maps.js
MalinAhlberg: Updated to the suggestion above in 26a2acc62d9759498d19a333708c528a05ed8baa

e0: Sorry for the slow response @MalinAhlberg. Good that you thought about the `eslint` warning. It's not ideal but I looked up how to [short-circuit a `forEach`](https://stackoverflow.com/a/2641374) and thought this was the better alternative.

There is also another variant using `every` (it immediately returns when a falsy value is returned in the iterator but I still think the `for` loop is more readable.

```javascript
    Object.values(mapsListing).flat().every(({ svgs }) => {
      if (svgs.length > 0) {
        avail2D = true;
        return false;
      }
      return true;
    });
```

frontend/src/components/explorer/MapViewer.vue
```diff
@@ -156,6 +158,12 @@ export default {
     queryParams() {
       return { ...this.mapQueryParams, ...this.dataOverlayQueryParams };
     },
+    switchTitle() {
+      if (this.avail2D) {
+        return `Switch to ${this.dimensionalState(!this.showing2D)}`;
+      }
+      return 'This model only has 3D maps availabe';
```
inghylt, line 165:
```suggestion
      return 'This model only has 3D maps available';
```

Another alternative could be "There are only 3D maps available for this model"

MalinAhlberg: :grimacing: oops.

frontend/src/store/modules/maps.js
MalinAhlberg: Thanks for the link and examples!! Using `every` is tempting, I had to try out the one-line-version. But it got very unreadable, especially with the negated truth value checking :smile:
```js
    avail2D = ! Object.values(mapsListing).flat().every(function(map) {return map.svgs.length === 0});
```

e0: That looks pretty good I think! I guess you can make this slightly shorter with using an arrow function and knowing that `0` is a falsy value in JavaScript.

```javascript
    avail2D = !Object.values(mapsListing).flat().every((m) => m.svgs.length);
```

nanjiangshu (approved):
Very nice with added information when switching is disabled

frontend/src/store/modules/maps.js
MalinAhlberg: Aha, nice :star: !!! Maybe not very readable for random humans, but indeed nice.

> Edit: On second thought, this might not work as intended. If the first compartment map does not have any SVGs but the other maps have, then this gives the wrong result right?

Since it flattened, that shouldn't be a problem?

e0: I meant my follow up suggestion, since it only looks at the length and doesn't check if it is equal to 0, it would be the inverse or your original one-liner.

MalinAhlberg: Ah, I see, good point. So `m.svgs` would be null(-ish) and hence falsy?
But then it would have to be like this
```js
    avail2D = !Object.values(mapsListing).flat().every((m) => m.svgs.length === 0);

```
Still nice!

frontend/src/components/explorer/MapViewer.vue
```diff
@@ -156,6 +158,12 @@ export default {
     queryParams() {
       return { ...this.mapQueryParams, ...this.dataOverlayQueryParams };
     },
+    switchTitle() {
```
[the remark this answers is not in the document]
MalinAhlberg: That's the actual title on the switch button, which is always "Switch to 2D / 3D". Do you think that should be changed to? In that case  'This model only has 3D maps available' might be a bit too long.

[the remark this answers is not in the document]
e0: > That's the actual title on the switch button, which is always "Switch to 2D / 3D". Do you think that should be changed to? In that case  'This model only has 3D maps available' might be a bit too long.

Want to chime in and say the button itself should not be modified imo. The tooltip change is great though. So I support the way things are now.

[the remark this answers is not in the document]
inghylt: Aah sorry my bad, yes makes sense to keep things as they are :+1:

frontend/src/store/modules/maps.js
mihai-sysbio: Random idea: instead of passing on this computation to the client browser, would it make sense to have it in the API response?

frontend/src/store/modules/maps.js
mihai-sysbio: `componentWithMap`?

inghylt (approved):
Nice work!

frontend/src/components/explorer/MapViewer.vue
```diff
@@ -156,6 +158,12 @@ export default {
...
+      if (this.avail2D) {
+        return `Switch to ${this.dimensionalState(!this.showing2D)}`;
+      }
+      return 'This model only has 3D maps available';
```
mihai-sysbio, line 165: Maybe this sounds clearer?
```suggestion
      return 'This model has only 3D maps available';
```

frontend/src/store/modules/maps.js
MalinAhlberg: I think it's a very reasonable idea, but I'm not sure it's worth the refactoring? It's not a heavy computation. But I'm up for changing if there is an agreement that we should do so :)

e0: > Random idea: instead of passing on this computation to the client browser, would it make sense to have it in the API response?

Interesting suggestion but I would not do this for the following reasons.

1. The performance difference is near-zero considering all the other intensive things that the frontend must handle such as rendering the maps.
2. This serves a very frontend specific purpose. IMO the data returned from the backend should be as flexible as possible so that the data is cleaner for API users.

frontend/src/components/explorer/MapViewer.vue
mihai-sysbio: Fixed in another PR #794 .

PR #787, opened 2021-11-29
Fix/interaction partner overlay error message

### LINKED ISSUE #770: Interaction Partners throws error message for majority of the models
**Describe the bug**

- Interaction Partners throws error message for all models except Human-GEM and Yeast-GEM. Happens for both gene and metabolites.
- This is also the case when loading Data Overlay in Map viewer for all models except Human-GEM and Yeast-GEM
- Yeast-GEM have the mock data files in prod, these should be removed as well
- Suggested to only support uploading custom data for maps that don't have relevant data to overlay on the server (i.e. all models except Human-GEM currently?)

**Steps to reproduce**

1. Visit `/explore`
2. Click Worm-GEM, Mouse-GEM, Rat-GEM or Fruitfly-GEM
3. Press Interaction Partners
4. Press a random gene or metabolite
5. Note that an error message is thrown (if the network is not too large)

**Expected behavior**

That step 1-5 does not lead to an error message for any model
The data overlay component should not show the default data section if there are no data sources. User uploaded data should still be allowed.

**Screenshots**

inghylt:
This PR closes #770

**Type of change**  

**List of changes made**  

- Backend does not throw 404 when models don't have a index for data overlay since this is expected for a majority of the models
- Frontend only shows custom data overlay functionality if there is no server data for this model
- Data overlay store values are reset when a data overlay component is destroyed

**Testing**  

Relative urls that can be reused both for production and local testing:

Error message:
    1. Visit /explore
    2. Click Worm-GEM, Mouse-GEM, Rat-GEM or Fruitfly-GEM
    3. Press Interaction Partners
    4. Press a random gene or metabolite
    5. Note that no error message is thrown and that only custom data upload functionality is provided
    
Reset values:
   1. Visit explore/Human-GEM/interaction-partners/ENSG00000168748
   2. Choose adipose tissue
   3. Visit /explore/Human-GEM/map-viewer?dim=2d
   4. Choose a map
   5. Note that the url contains `datatype=None&datasource=None&dataSet=None` and not `datatype=transcriptomics&datasource=hpaRna.tsv&dataSet=adipose tissue`
   6. Open data overlay sidebar
   7. Note that the url values for data overlay are the same as before the sidebar was opened

Please also play around with the Interaction Partner and Map Viewer for different models to make sure I haven't missed anything

**Further comments**  

- The fix regarding resetting the values was not part of issue #770, but the bug was discovered during the development and required such a small fix that I though it make sense to include it in this PR. I can remove it if you feel it is bad to include in this PR

- Yeast-GEM has mock data deployed, these are removed in https://github.com/MetabolicAtlas/data-files/pull/25

**Checklist**  

### DIFF
--- a/api/src/endpoints/dataOverlay.js
+++ b/api/src/endpoints/dataOverlay.js
@@ -19,8 +19,12 @@ routes.get('/:model', async (req, res) => {
     );
     res.json(JSON.parse(indexJson));
   } catch (e) {
-    console.error(e.message);
-    res.sendStatus(404);
+    if (e.code === 'ENOENT'){
+      res.json({});
+    } else {
+      console.error(e.message);
+      res.sendStatus(404);
+    }
   }
 });
 
--- a/frontend/src/components/explorer/mapViewer/DataOverlay.vue
+++ b/frontend/src/components/explorer/mapViewer/DataOverlay.vue
@@ -34,37 +34,41 @@
     <div class="card my-3">
       <div class="card-content py-2 p-3">
         <div class="has-text-centered title is-size-6">Data</div>
-        <div v-if="Object.keys(dataSourcesIndex).length > 0" class="control">
-          <p>Select data type</p>
-          <div v-if="dataType" class="select is-fullwidth">
-            <select @change="handleDataTypeSelect">
-              <option v-for="type in Object.keys(dataSourcesIndex)" :key="type"
-                      :selected="type === dataType.name"
-                      :value="type"
-                      class="is-clickable is-capitalized">{{ type }}</option>
-            </select>
+        <div v-if="modelHasOverlayData()">
+          <div class="control">
+            <p>Select data type</p>
+            <div v-if="dataType" class="select is-fullwidth">
+              <select @change="handleDataTypeSelect">
+                <option v-for="type in Object.keys(dataSourcesIndex)" :key="type"
+                        :selected="type === dataType.name"
+                        :value="type"
+                        class="is-clickable is-capitalized">{{ type }}</option>
+              </select>
+            </div>
           </div>
-        </div>
-        <div v-if="Object.keys(dataSourcesIndex).length > 0" class="control">
-          <p>Select data source</p>
-          <div v-if="dataType" class="select is-fullwidth">
-            <select @change="handleDataSourceSelect">
-              <option v-for="s in dataSourcesIndex[dataType.name]" :key="s.filename"
-                      :selected="dataSource && s.filename === dataSource.filename"
-                      :value="s.filename"
-                      class="is-clickable is-capitalized">{{ s.name }}</option>
-            </select>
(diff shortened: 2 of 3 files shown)


e0 (approved):
Great PR fix and description!

A minor thing I noticed was that the "Data" `div` still shows even when no custom data has been added. Do you think it would be better to hide that by default too? I'm approving the PR already because I think this is a debatable point and I'm ok with either direction.

image

inghylt:
@e0 Yes I noted that too! I left it since that is what we currently have in prod and we don't really have a title for the colored bar. But I'm fine with having the `Data` title being conditional as well

MalinAhlberg (approved):
Good job! This looks very good, works for all cases I've tested.

Related to the resetting overlays values, if you
- go to `/explore/Human-GEM/map-viewer/protein_modification?dim=2d&panel=1&sel=&search=&coords=-830.11,-1766.88,0.32,0,0,500&datatype=transcriptomics&datasource=hpaRna.tsv&dataSet=cerebral%20cortex` (protein modification with overlay for cerebral cortex chosen)
- change maps to eg `purine metabolism` and then change back to `protein modification` without closing the overlay sidebar

then cerebral cortex will still be selected in the menu, but not applied to the map.

Obviously a bit unrelated to the original issue, but I mention it anyway since you've been working with the overlay store values and it might be easy to fix.

nanjiangshu:
> Good job! This looks very good, works for all cases I've tested.
> 
> Related to the resetting overlays values, if you
> 
>     * go to `/explore/Human-GEM/map-viewer/protein_modification?dim=2d&panel=1&sel=&search=&coords=-830.11,-1766.88,0.32,0,0,500&datatype=transcriptomics&datasource=hpaRna.tsv&dataSet=cerebral%20cortex` (protein modification with overlay for cerebral cortex chosen)
> 
>     * change maps to eg `purine metabolism` and then change back to `protein modification` without closing the overlay sidebar
> 
> 
> then cerebral cortex will still be selected in the menu, but not applied to the map.
> 
> Obviously a bit unrelated to the original issue, but I mention it anyway since you've been working with the overlay store values and it might be easy to fix.

I think this is a bug. Actually, when you change the 2D map while keep the data overlay side bar open, the selected tissue is still there but the transcriptomics data will not be overlayed on the 2D map. @MalinAhlberg, could you please create an issue for that bug?

nanjiangshu (approved):
Great fix! I randomly tested various situations and all works as expected.

MalinAhlberg:
> I think this is a bug. Actually, when you change the 2D map while keep the data overlay side bar open, the selected tissue is still there but the transcriptomics data will not be overlayed on the 2D map. @MalinAhlberg, could you please create an issue for that bug?

#789

inghylt:
@MalinAhlberg Very good catch! But yes makes sense to fix that in a separate issue and PR

PR #793, opened 2021-12-02
chore: add linter

### LINKED ISSUE #604: Add lint integration

To be linted:
- [ ] `.js`
- [ ] `.vue`
- [ ] `.scss`

MalinAhlberg:
This is very much work in progress, don't hesitate to have ideas on how to do things differently!

---------------

This PR closes #604.

Add prettier, remove eslint, set up automatic formatting of files on commit.

**Type of change**  
- [X] Other
 
**List of changes made**  

- [X] Remove old linters (would we prefer keeping eslint?)
- [X] Add prettier
- [X] Set up hooks to format all staged files on commit
- [ ] Describe how to use with editor (vim)
- [ ] Set up GH action to run prettier on every pull request
- [ ] Configure prettier
- [ ] Let prettier go through all existing code

**Testing**  
Check out this branch and
- `cd frontend`
- `npm i`
- open a file, eg `frontend/src/main.js`, and add something which you think should cause an syntactic error. For example, join the first two lines
- `git add frontend/src/main.js` and then commit it. Open the file again and make sure that your error has been fixed (or that you got an error message if it couldn't be fixed automatically).

You can also set up your editor to highlight errors for you, by using eg ALE, see https://prettier.io/docs/en/vim.html.


e0:
Really nice work on this so far @MalinAhlberg!

Regarding eslint, I think I mentioned previously that we could remove the `frontend/.eslintrc.js` file but that was misleading. What I had in mind was that we could remove all of the custom eslint rules but still have eslint in the project. More about prettier vs lint here: https://prettier.io/docs/en/comparison.html

Having said that, I hadn't really explored the idea of having only prettier and not eslint. I'm open to the idea but I need to think a bit more about it. Sorry for not clarifying initially.

MalinAhlberg:
Thanks for the feedback :) !
I had to restart this work, so I will close this PR and create a new one.

PR #794, opened 2021-12-02
feat: deploy improvements

### LINKED ISSUE #618: Improve deployment workflow regarding environment variables for different environments
As a developer, I would like to improve the deployment workflow regarding environment variables, so that I can avoid editing the `.env` file before deploying to a different environment

This is to avoid potential problems with using wrong environment variables when deploying.

For example, different environment files could be used:

- `.env_local` -> localhost
- `.env_dev` -> dev server
- `.env_prod` -> prod server

mihai-sysbio:
This PR closes #618 by implementing a way to use a user-provided short text to decide which environment file to use, relying on that environment file to also select the appropriate Docker context.

**Type of change**  

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
 
**List of changes made**  

Instead of having to manually switch the `.env`, there are now multiple environment files that can be used, with the expected pattern `env-[NAME].env`. Before, the deployment command was expecting only a Docker context. Now, it is expecting the desired `[NAME]`, using it to identify the desired `.env` file, which now also specified the Docker context. I think this way of specifying the context might require a certain Docker version.

**Screenshot of the fix**  

**Testing**  

- Instructions on how to test

**Further comments**  
After fetching the `.env` files, try building locally and also
```bash
deploy-stack dev
```

**Checklist**  

- [x] I have made corresponding changes to the documentation
- [ ] ~I have checked so that the same data as before is returned by the API~ n/a

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,16 @@ start-stack
 
 Given successful deployment, the frontend should be accessible at: `http://localhost/`. If you encounter any problems try looking at the logs `logs api` / `logs frontend`.
 
+To deploy the stack to a remote server, create another `.env` file and modify it accordingly:
+```bash
+cp env-local.env.sample env-dev.env
+```
+
+Use the name of the file (`dev` in the previous example) as an argument to the `deploy-stack` command to pass the `.env` file to the Docker command and the Docker compose file:
+```bash
+deploy-stack dev
+```
+
 ## Description of helper commands
 
 * To bootstrap the project: `build-stack`
--- a/README.md
+++ b/README.md
@@ -55,12 +55,12 @@ start-stack
 
 Given successful deployment, the frontend should be accessible at: `http://localhost/`. If you encounter any problems try looking at the logs `logs api` / `logs frontend`.
 
-To deploy the stack to a remote server, create another `.env` file and modify it accordingly:
+To deploy the stack to a remote server, create another `ENV` file, e.g. `env-dev.env`, and modify it accordingly:
 ```bash
 cp env-local.env.sample env-dev.env
 ```
 
-Use the name of the file (`dev` in the previous example) as an argument to the `deploy-stack` command to pass the `.env` file to the Docker command and the Docker compose file:
+Use the tag of the file (`dev` in the previous example) as an argument to the `deploy-stack` command to pass the `ENV` file to the Docker command and the Docker compose file:
 ```bash
 deploy-stack dev
 ```
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ In the folder `MetabolicAtlas` that has been cloned, add a `env-local.env` file 
 ```bash
 cp env-local.env.sample env-local.env
 ```
-and modify this `.env` file.
+and modify this `env-local.env` file.
 
 Make sure the paths for `DATA_FILES_PATH` and `DATA_GENERATOR_PATH` are correct for your setup, eg. the paths to where you have downloaded the repositories `data-files` and `data-generation`.
 
--- a/README.md
+++ b/README.md
@@ -28,38 +28,15 @@ Clone the three required repositories by
     git clone https://github.com/MetabolicAtlas/data-files && pushd data-files; git lfs pull; popd
 
 
-Go to the repository `data-generation` and follow the
-[instructions](https://github.com/MetabolicAtlas/data-generation#readme) on how to generate the data files required by MetabolicAtlas.
+Go to the repository `data-generation` and follow the [instructions](https://github.com/MetabolicAtlas/data-generation#readme) on how to generate the data files required by Metabolic Atlas.
 
-In the folder `MetabolicAtlas` that has been cloned, add a `.env` file based on the `.env.sample` file:
+In the folder `MetabolicAtlas` that has been cloned, add a `env-local.env` file based on the `env-local.env.sample` file:
 ```bash
-cp .env.sample .env
+cp env-local.env.sample env-local.env
 ```
 and modify this `.env` file.
 
-The content of the file `.env` that has just been copied from `.env.sample` is shown below. Make sure the paths for `DATA_FILES_PATH` and `DATA_GENERATOR_PATH` are correct for your setup,
-eg. the paths to where you have downloaded the repositories `data-files` and `data-generation`.
-
-
-```
-CERTBOT_EMAIL=
-SERVER_NAME=localhost
-DATA_FILES_PATH=../data-files
-DATA_GENERATOR_PATH=../data-generation
-
-NEO4J_USERNAME=neo4j
-NEO4J_PASSWORD=password-unhackable
-
-FTP_MIN_PORT=30000
-FTP_MAX_PORT=31000
-
-PORT_NGINX=80
-PORT_FRONTEND=81
-PORT_NEO4J_1=7474
-PORT_NEO4J_2=7687
-
-IP_FILTER=""
-```
+Make sure the paths for `DATA_FILES_PATH` and `DATA_GENERATOR_PATH` are correct for your setup, eg. the paths to where you have downloaded the repositories `data-files` and `data-generation`.
 
 To load the list of helper commands run:
 ```bash
@@ -94,9 +71,7 @@ When rebuilding the stack, you might have to change the ownership of the directo
 ```
 sudo chown -R <user> neo4j
 ```
-Replace `<user>` with your user name.
-The ownership will be automatically reset when running the project, so you will
-have to repeat this step for every rebuild.
+Replace `<user>` with your user name. The ownership will be automatically reset when running the project, so you will have to repeat this step for every rebuild.
 
 
 ## Licenses
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -11,7 +11,6 @@ services:
   api:
     container_name: api
     build: api
-    env_file: .env
     depends_on: 
       - neo4j
     networks:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -11,6 +11,7 @@ services:
   api:
     container_name: api
     build: api
+    env_file: ${CHOSEN_ENV}
     depends_on: 
       - neo4j
     networks:
--- a/env-local.env.sample
+++ b/env-local.env.sample
@@ -1,6 +1,9 @@
 ## The variables below are required for the production enviroment.
 ## Some might be also required for the local environment.
 
+# Assign this to the adequate Docker context
+export DOCKER_CONTEXT=default
+
 # These paths are needed to use code and files from other
 # repositories (dependencies).
 DATA_FILES_PATH=../data-files
--- a/env-local.env.sample
+++ b/env-local.env.sample
@@ -1,8 +1,11 @@
 ## The variables below are required for the production enviroment.
 ## Some might be also required for the local environment.
 
-# Assign this to the adequate Docker context
+# Assign this to the adequate Docker context.
 export DOCKER_CONTEXT=default
+
+# Set if Node.js should run in develop or production mode.
+NODE_ENV=develop
 
 # These paths are needed to use code and files from other
 # repositories (dependencies).
--- a/env-local.env.sample
+++ b/env-local.env.sample
@@ -2,9 +2,16 @@
 ## Some might be also required for the local environment.
 
 # Assign this to the adequate Docker context.
+# This will be used in the `docker context` command.
+# To read more about how to create and use a Docker context, visit
+# https://docs.docker.com/engine/reference/commandline/context/
+# The default value for a Docker setup is that the local context
+# is called `default`.
+# When deploying to a remote server, first create a Docker context
+# manually, and then reference it here.
 export DOCKER_CONTEXT=default
 
-# Set if Node.js should run in develop or production mode.
+# Set if Node.js should run in `develop` or `production` mode.
 NODE_ENV=develop
 
 # These paths are needed to use code and files from other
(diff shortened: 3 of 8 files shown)


MalinAhlberg (approved):
Nice! Works well for me, and I enjoy not having to move around the env files.

README.md
```diff
@@ -28,38 +28,15 @@ Clone the three required repositories by
...
-[instructions](https://github.com/MetabolicAtlas/data-generation#readme) on how to generate the data files required by MetabolicAtlas.
+Go to the repository `data-generation` and follow the [instructions](https://github.com/MetabolicAtlas/data-generation#readme) on how to generate the data files required by Metabolic Atlas.
 
-In the folder `MetabolicAtlas` that has been cloned, add a `.env` file based on the `.env.sample` file:
+In the folder `MetabolicAtlas` that has been cloned, add a `env-local.env` file based on the `env-local.env.sample` file:
 ```bash
-cp .env.sample .env
+cp env-local.env.sample env-local.env
 ```
 and modify this `.env` file.
```
nanjiangshu, line 37:
```suggestion
and modify this `env-local.env` file.
```

frontend/src/components/explorer/MapViewer.vue
```diff
@@ -163,7 +163,7 @@ export default {
       if (this.avail2D) {
         return `Switch to ${this.dimensionalState(!this.showing2D)}`;
       }
-      return 'This model only has 3D maps available';
+      return 'This model has only 3D maps available';
```
nanjiangshu, line 166: Why changing the MapViewer in this PR?

README.md
```diff
@@ -78,6 +55,16 @@ start-stack
 
 Given successful deployment, the frontend should be accessible at: `http://localhost/`. If you encounter any problems try looking at the logs `logs api` / `logs frontend`.
 
+To deploy the stack to a remote server, create another `.env` file and modify it accordingly:
+```bash
+cp env-local.env.sample env-dev.env
+```
+
+Use the name of the file (`dev` in the previous example) as an argument to the `deploy-stack` command to pass the `.env` file to the Docker command and the Docker compose file:
```
nanjiangshu, line 63:
```suggestion
Use the tag of the file (`dev` in the previous example) as an argument to the `deploy-stack` command to pass the `env-dev.env` file to the Docker command and the Docker compose file:
```

nanjiangshu: I think it might be good to mention that the  env variable `DOCKER_CONTEXT` in the file `env-dev.env` needs to be changed accordingly.

nanjiangshu (commented):
Great job! I have tested deploying both locally and on the dev server and it works well on both cases. However, I think it might be good to add example ENV files for the deployment to the remote server as well. Maybe add two files `env-dev.env.example`and `env-prod.env.example` or just one of them depending on how they should be distinguished. 

Apart from that, I've added a few minor comments.

frontend/src/components/explorer/MapViewer.vue
mihai-sysbio: It was an unresolved comment from [a previous PR](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/786#discussion_r766683723).

README.md
mihai-sysbio: Like all other variables, the use of `DOCKER_CONTEXT` is documented. I'm thinking we either document in the Readme all "important" variables, or none.

README.md
```diff
@@ -78,6 +55,16 @@ start-stack
 
 Given successful deployment, the frontend should be accessible at: `http://localhost/`. If you encounter any problems try looking at the logs `logs api` / `logs frontend`.
 
+To deploy the stack to a remote server, create another `.env` file and modify it accordingly:
```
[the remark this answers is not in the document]
mihai-sysbio: Resolved in 6160976.

mihai-sysbio:
> I think it might be good to add example ENV files for the deployment to the remote server as well. Maybe add two files `env-dev.env.example`and `env-prod.env.example` or just one of them depending on how they should be distinguished.

These could be added but I wouldn't be able to provide sample values that are different from the ones in the `local` sample, so I'm not sure what value they would bring. How do you imagine they would come of use?

frontend/src/components/explorer/MapViewer.vue
inghylt: Also think this is unrelated to this PR, Would create another PR for this

README.md
nanjiangshu: I could not find the documentation of DOCKER_CONTEXT in MetAtlas, or do you mean the general documentation of DOCKER_CONTEXT for docker?

frontend/src/components/explorer/MapViewer.vue
mihai-sysbio: A PR without an issue then? Is that acceptable?

nanjiangshu:
> > I think it might be good to add example ENV files for the deployment to the remote server as well. Maybe add two files `env-dev.env.example`and `env-prod.env.example` or just one of them depending on how they should be distinguished.
> 
> These could be added but I wouldn't be able to provide sample values that are different from the ones in the `local` sample, so I'm not sure what value they would bring. How do you imagine they would be of use?

It is not obvious from the README that how the ENV file for deploying to the dev server should be changed. I think that is the value to add and example `env-dev.env.example` with 

`DOCKER_CONTEXT`, `ERTBOT_EMAIL`, `SERVER_NAME` and `IP_FILTER` set to the correct value. 

Or maybe it is better to add in the README how these variables should be set when deploying to the `dev` server if only one example file is provided.

README.md
mihai-sysbio: Sorry, I meant that the variable is explained on the line prior it's use in `sample.env`.

frontend/src/components/explorer/MapViewer.vue
inghylt: For typo kind of things, i.e. changing one or two words in one file, I believe so. Things that does not require an extensive review. If a change affects several files/actual code functionality, an new issue should be created IMO. But apparently something that is not established so we should discuss it further.

What we have here is a PR related to an issue with changes unrelated to that issue, I'm not fan of that

README.md
nanjiangshu: I would say a new member will encounter problem by just reading `To deploy the stack to a remote server, create another ENV file, e.g. env-dev.env, and modify it accordingly:` without assistant from others. 
However, I agree that we don't need over documenting, since only the developers of MetAtlas will deploy to the dev server.

frontend/src/components/explorer/MapViewer.vue
mihai-sysbio: I'm also not a fan of the amount of clicks and typing required in a PR template for what could essentially be a typo vs a single, individual commit with clear explanation.

This commit has been moved to a separate PR https://github.com/MetabolicAtlas/MetabolicAtlas/pull/802

README.md
mihai-sysbio: I think we have continued this discussion in the main thread, so I am closing it here.

frontend/src/components/explorer/MapViewer.vue
e0: From what I remember, what we have established before was to check with the team if an un-established action is intended.

In this case, @mihai-sysbio did ask about this in another PR but no one responded so I feel like the rest of the team could have responded better there. Personally I feel like it was not a big difference either way but I could have written that instead of ignoring it. I'm also ok with including tiny changes like this in any unrelated PR (such as this one), but this is not what we have previously agreed on and it would have been nice to discuss with the team. Yes it is tedious to have a discussion about such a tiny fix, but if we can establish a new norm from that discussion it would help us avoid these situations in the future.

nanjiangshu (approved):
Looks good to me now.

inghylt (approved):

README.md
```diff
@@ -28,38 +28,15 @@ Clone the three required repositories by
     git clone https://github.com/MetabolicAtlas/data-files && pushd data-files; git lfs pull; popd
 
 
-Go to the repository `data-generation` and follow the
-[instructions](https://github.com/MetabolicAtlas/data-generation#readme) on how to generate the data files required by MetabolicAtlas.
+Go to the repository `data-generation` and follow the [instructions](https://github.com/MetabolicAtlas/data-generation#readme) on how to generate the data files required by Metabolic Atlas.
 
-In the folder `MetabolicAtlas` that has been cloned, add a `.env` file based on the `.env.sample` file:
+In the folder `MetabolicAtlas` that has been cloned, add a `env-local.env` file based on the `env-local.env.sample` file:
```
e0, line 33: This question applies for the rest of the PR too. Could we do without the `env-` prefix? So we would have `local.env` and `prod.env` for example.

docker-compose.yml
```diff
@@ -11,7 +11,7 @@ services:
   api:
     container_name: api
     build: api
-    env_file: .env
+    env_file: ${CHOSEN_ENV}
```
e0, line 14: Just a minor thing but I personally would prefer `TARGET_ENV` as opposed to `CHOSEN_ENV`. Not a strong opinion so feel free to ignore this.

env-local.env.sample
```diff
@@ -1,12 +1,28 @@
...
 ## Some might be also required for the local environment.
 
+# Assign this to the adequate Docker context.
+# This will be used in the `docker context` command.
+# To read more about how to create and use a Docker context, visit
+# https://docs.docker.com/engine/reference/commandline/context/
+# The default value for a Docker setup is that the local context
+# is called `default`.
+# When deploying to a remote server, first create a Docker context
+# manually, and then reference it here.
```
e0, line 11: Convenient with this inline doc here. There is a bit of overlap with what we have in the wiki. Do we want to make that page public, perhaps under `.github` and link to that instead?

e0 (commented):
Looks great @mihai-sysbio! Really nice that you added comments to the env files. There is one issue about naming change (about the env files) that I wanted to discuss before approving. Other issues are minor and feel free to ignore them.

proj.sh
```diff
@@ -36,12 +37,13 @@ function clean-stack {
 }
 
 function logs {
-  docker compose -f docker-compose.yml -f docker-compose-local.yml logs -f $@
+  docker compose --env-file $CHOSEN_ENV -f docker-compose.yml -f docker-compose-local.yml logs -f $@
 }
 
 function deploy-stack {
+  CHOSEN_ENV="env-${1:=local}.env"
```
e0, line 44: The `$CHOSEN_ENV` var should be reset to "local" after deploying (regardless if the deploy is successful or not). Or else the other commands such as `clean-stack` or `import-db` will run on the run context. Not sure what's the best way to go about this.

e0: Also we should really update the `clean-stack` command so it doesn't remove all docker containers. @mihai-sysbio do you know if we have an existing issue for this? I can create one otherwise.

e0: Fixed in https://github.com/MetabolicAtlas/MetabolicAtlas/pull/794/commits/ca9ac041c9351f46f1825e3676243dfbc26cf1a7

mihai-sysbio: Thanks a lot for the fix. I see that we are referring to `local` in a few places in that file, which isn't so nice. I'm okay with leaving it be though.

env-local.env.sample
mihai-sysbio: We do have some stuff in the wiki, but if we would move it to a public location I'd rather we do it in a planned way, by having a dedicated issue and a discussion about it. That's why it feels easier to me right now to reference some external documentation here, like a guide on Medium or something.

README.md
e0: Ah ok, hadn't thought about that at all. In my file browser I usually display things grouped by "Kind"anyways. Personally I would value less redundant file name parts more than this detail but it's not a big deal. I think it's overkill to have a folder for this now too.

image

proj.sh
e0: Definitely room for improvement. I'm fine with leaving it like this now as well.

@mihai-sysbio just wanted to ping you about [this comment](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/794#discussion_r768790064) in case you missed it. Would be nice to get your input.

mihai-sysbio: Ah thanks for the reminder, I missed that.
If I understand correctly, you're proposing to keep the `clean-stack` command, but to improve its behavior so that it removes only project-related containers, instead of all containers?

README.md
mihai-sysbio: I didn't even think of the option of grouping files like that with Finder, I was just thinking of my own use-case in VS Code and imagined it would be quite common for all alphabetically-sorted terminal-based scenarios as well
image
I feel we should go with the popular vote for this one instead of basing this on my own use-case.

docker-compose.yml
mihai-sysbio: I thought `TARGET_ENV` can be misinterpreted as "the env _of the_ target". So I purposefully opted for the admittedly less-common wording `CHOSEN_ENV` hoping that it would be clearer that as a developer I am deploying the stack for this chosen environment.
How does this explanation resonate with you?

README.md
e0: Feels like this is not too big of a deal. Let's abandon the change suggestion for now.

docker-compose.yml
e0: That sounds fine! I was really just nitpicking on the past perfect tense of the verb. Let's leave it as is for now.

proj.sh
e0: > If I understand correctly, you're proposing to keep the `clean-stack` command, but to improve its behavior so that it removes only project-related containers, instead of all containers?

Yes, that's right.

e0 (approved):
Sorry that I hadn't responded to these earlier @mihai-sysbio. Thanks to @inghylt and @nanjiangshu for reminding about this during the standup. Good job again with the PR!

proj.sh
e0: @mihai-sysbio To follow up on this. Did we agree that there is no current issue for this and a new one should be created? Also as mentioned yesterday, please feel free to merge this PR even before the new issue is created (if necessary).

mihai-sysbio: #816 was create to address this, as the behavior of this command was neither introduced nor altered with this PR.

Issue #719, opened 2021-10-19
resolve vulnerabilities


mihai-sysbio:
Aim for no warnings https://github.com/MetabolicAtlas/MetabolicAtlas/security/dependabot

e0:
If the dependabot issue is not directly related to the versions, it might be good to not change any versions for now. For example it seems like node 14 is needed for now and the latest Docker version is causing some problems for deployment.

mihai-sysbio:
The dependabot warnings are afaik only for `package.json`, not related with Docker.

e0:
Thanks for the reply @mihai-sysbio . Looking at https://github.com/MetabolicAtlas/MetabolicAtlas/security/code-scanning, I'm not sure exactly how `package.json` is related to these issues. Can you please provide more details?

mihai-sysbio:
There is a different issue for the CodeQL warning. This issue is different - the OP links to dependabot.

e0:
Ah that's right, sorry. That link is not working though.

mihai-sysbio:
It does, as long as one is signed in.

e0:
I have double checked that I am signed in. Is it an issue with user permissions?

mihai-sysbio:
You're right, it was a permission issue. Somewhere in the org settings under `Security & analysis`. It should be fixed now.

nanjiangshu:
Now I see the same page as yours @mihai-sysbio . Thanks for fixing the permission.

e0:
Relevant issue for `ansi-html`: https://github.com/Tjatse/ansi-html/issues/21

inghylt:
https://docs.github.com/en/code-security/supply-chain-security/managing-vulnerabilities-in-your-projects-dependencies/about-dependabot-security-updates

I think it would make sense to try this out for updating the dependencies. And maybe keep it so that it is easier to update in the future since it will automatically create a PR that we can review an see that everything is working accordingly. That would avoid us having to do this type of update where a lot of dependencies are changed at the same time

https://docs.github.com/en/code-security/supply-chain-security/managing-vulnerabilities-in-your-projects-dependencies/configuring-dependabot-security-updates#enabling-or-disabling-dependabot-security-updates-for-an-individual-repository
This is how we add it in that case, however I do not have permissions to do so

mihai-sysbio:
[...]

PR #797, opened 2021-12-07
feat: prettier

### LINKED ISSUE #604: Add lint integration

MalinAhlberg:
This PR closes #604.

It adds prettier, sets up automatic formatting of files on commit, checks file formats on pull requests.

All code in `/frontend` is reformatted by prettier. That makes this a very large PR. It could be split in two, one or all the configuration, and one for the code?

**Type of change**  
- [X] Other
 
**List of changes made**  

- [X] Add prettier
- [X] Set up hooks to format all staged files on commit
- [X] Set up GH action to run prettier on every pull request
- [X] Configuations of eslint and prettier
- [X] Let prettier go through all existing code
- [ ] Set up for backend (`/api`)

**Testing**  
***Autoformatting and running eslint on commit***
Check out this branch and
-  rebuild the stack
- `git config core.hooksPath git_hooks`
- open a file, eg `frontend/src/main.js`, and add something which you think should cause an syntactic error. For example, join the first two lines
- `git add frontend/src/main.js` and then commit it. Open the file again and make sure that your error has been fixed (or that you got an error message if it couldn't be fixed automatically).

***Running eslint and prettier manually***
- `source proj.sh` (or replace `ma-exec` with `docker compose -f docker-compose.yml -f docker-compose-local.yml` in the following commands)
- `ma-exec frontend yarn lint` to lint all `js` and `vue` files
- `ma-exec frontend yarn check-format` to print all files that should be reformatted
- `ma-exec frontend yarn format` to format all files

***In your editor***
You can set up your editor to highlight errors (eslint) and autoformat your code on save (prettier) by using eg ALE, see https://github.com/dense-analysis/ale and https://prettier.io/docs/en/vim.html.

**Further comments**
- To make eslint and prettier agree, the following eslint rules have been disabled:
     -   `vue/max-attributes-per-line`
     -    `comma-dangle`
     -    `max-len` prettier will take care of the line length, setting it to ~80, see https://prettier.io/docs/en/options.html#print-width
    
- The following vue rules have been disabled, as they are introduced by newer versions. They should probably be enabled, but in a separate issue/pr.
     -   `vue/multi-word-component-names`
     -   `vue/no-useless-template-attributes`
      -  `vue/no-lone-template`,
      -  `vue/order-in-components`

- The browser error overlay has been disabled, as I could not get it to apply the configuration of disabled rules above. Could be re-enabled when this is solved. (This is the only change done by me to `frontend/vue.config.js`. The rest is done by prettier.)

- Regarding the reformatted code, I had to do some manual editing. The main problem is that prettier often puts comments on a 
new line. If the comment is of the type `eslint-disable-line`, then eslint will be upset next time it runs. I have changed to `eslint-disable-nextline` and `/* eslint-disable rule*/ ... /* eslint-enable rule*/` for this. See https://github.com/eslint/eslint/issues/11247.

- The only other thing I had to change manually was this [p tag](https://github.com/MetabolicAtlas/MetabolicAtlas/blob/11ddfc55c25874b14b25e39ce62ef552b3830335/frontend/src/components/explorer/gemBrowser/Tile.vue#L51) which was previously closed after the `ul` tag. That was not acceptable by prettier. It would be great if someone could verify that this fix is correct.

- I have added a workflow for an github action on pull requests. It checks that prettier and eslint are happy.


frontend/.prettierrc.json
```diff
@@ -0,0 +1,6 @@
+{
+  "trailingComma": "es5",
+  "tabWidth": 4,
```
e0, line 3: Is this a recommended setting from somewhere? I think the default in prettier is 2 and that's what we had in the project as well. The PR should be a lot smaller as well if we keep it at 2.

MalinAhlberg: aha, good that you noticed this! I took it from [here](https://prettier.io/docs/en/configuration.html#basic-configuration) (also saw it at other places), but if 2 is our standard, I agree we should keep that.

MalinAhlberg: fixed by 22701255906e2abeac8aa41496d8e5997f1b334f

mihai-sysbio:
Would it be possible to perform the linting in GH actions via our own Dockerfile(s)? I'm thinking it would be a lot easier to maintain consistency this way by the same setup between what we have running locally and what the workflow does. For example, the node version is now defined in both places (dockerfile and workflow).

MalinAhlberg:
> Would it be possible to perform the linting in GH actions via our own Dockerfile(s)? I'm thinking it would be a lot easier to maintain consistency this way by the same setup between what we have running locally and what the workflow does. For example, the node version is now defined in both places (dockerfile and workflow).

Yes, @e0 also suggested this and I think he is looking at it atm :)

e0:
> > Would it be possible to perform the linting in GH actions via our own Dockerfile(s)? I'm thinking it would be a lot easier to maintain consistency this way by the same setup between what we have running locally and what the workflow does. For example, the node version is now defined in both places (dockerfile and workflow).
> 
> Yes, @e0 also suggested this and I think he is looking at it atm :)

I was mostly focusing on getting things to run inside docker locally now. I'm wrapping up the work for today (this week) but for anyone continuing to work on using docker for GH actions, this seems like a relevant article: https://docs.github.com/en/actions/creating-actions/creating-a-docker-container-action

MalinAhlberg:
Thanks so much for the fixes you did, @e0! It looks very good!
I've made two more commits:
- eb2068699743c228b269f543217433cfb97ec18c Only eslint staged files (and only files in `frontend/src`). I think it's nice if you can have messy, but unstaged, files around while you commit something else.
- 4f2c146993a83775c7df6438c645eb66240822ad On Ubuntu, I need to use `/bin/bash` to get the script running

I also had troubles with `.eslintrc.js` and `.prettierrc.json`, they were not updated in the container, even after a rebuild, and hence the linters did not work at first. I had to copy over them manually, but I'm sure someone who is more fond of docker knows how to solve this.

e0:
> Thanks so much for the fixes you did, @e0! It looks very good! I've made two more commits:
> 
> * [eb20686](https://github.com/MetabolicAtlas/MetabolicAtlas/commit/eb2068699743c228b269f543217433cfb97ec18c) Only eslint staged files (and only files in `frontend/src`). I think it's nice if you can have messy, but unstaged, files around while you commit something else.
> * [4f2c146](https://github.com/MetabolicAtlas/MetabolicAtlas/commit/4f2c146993a83775c7df6438c645eb66240822ad) On Ubuntu, I need to use `/bin/bash` to get the script running
> 
> I also had troubles with `.eslintrc.js` and `.prettierrc.json`, they were not updated in the container, even after a rebuild, and hence the linters did not work at first. I had to copy over them manually, but I'm sure someone who is more fond of docker knows how to solve this.

The commits you added look great! I didn't know about the ubuntu/linux needing `/bin/bash`, TIL.

It's weird that it didn't work after a rebuild, to mount (or get two-way sync of) files or folders, we use the `docker-compose*` files. For example [these lines](https://github.com/MetabolicAtlas/MetabolicAtlas/blob/develop/docker-compose-local.yml#L18-L20) indicate the volumes that are currently mounted.

api/.dockerignore
```diff
@@ -1,3 +1,2 @@
 README
 Dockerfile
-node_modules
```
mihai-sysbio, line 3: This line is needed to keep the Docker image small - relevant for faster transfer times when deploying the stack. Is it necessary to remove it?

frontend/src/App.vue
```diff
@@ -116,14 +154,21 @@
...
-          <p class="button is-small is-rounded has-background-danger has-text-white has-text-weight-bold"
-             @click="showCookieMsg=false; acceptCookiePolicy()">
+          We use cookies to enhance the usability of our website. By continuing you are agreeing to
+          our
+          <router-link
+            class="has-text-white has-text-weight-bold"
+            :to="{ name: 'about', hash: '#privacy' }"
+          >
+            Privacy Notice and Terms of Use </router-link
+          >&emsp;
```
mihai-sysbio, line 164: This looks weird, I was expecting:
```suggestion
            Privacy Notice and Terms of Use
          </router-link>
          &emsp;
```

api/.dockerignore
e0: This is definitely work in progress and if we move forward with it we should have different files for local and remote. Good that you spotted this.

frontend/src/App.vue
```diff
@@ -116,14 +154,21 @@
...
+          <p
+            class="button is-small is-rounded has-background-danger has-text-white has-text-weight-bold"
+            @click="
+              showCookieMsg = false
```
mihai-sysbio, line 168: Can somebody confirm that no semicolon is needed here?

nanjiangshu: I can confirm that ';' is not needed here.

frontend/src/App.vue
nanjiangshu: This does look weird and it happens at multiple places. 
I did a quick search with the command 
```$ grep -irn "[a-z].*</router-link" .``` in the folder `frontend/src` and found quite a few examples. 

```
./components/SearchTable.vue:53:              <span class="suggestions">{{ v }}</span> </router-link
./components/SearchTable.vue:187:                        >{{ sub.name }}</router-link
./components/SearchTable.vue:209:                          >{{ comp.name }}</router-link
./components/About.vue:44:            >Human-GEM</router-link
./components/About.vue:48:            >Yeast-GEM</router-link
./components/About.vue:82:          <router-link :to="{ name: 'gems' }">GEM Repository</router-link> for links to those.<br /><br />
./components/explorer/gemBrowser/Reaction.vue:41:                  {{ v.name }}</router-link
./components/explorer/gemBrowser/Reaction.vue:55:                      >{{ c.name }}</router-link
./components/explorer/gemBrowser/Reaction.vue:86:                {{ rr.id }} </router-link
./components/explorer/gemBrowser/Subsystem.vue:37:                    >{{ c.name }}</router-link
./components/explorer/gemBrowser/Metabolite.vue:43:                >{{ metabolite[el.name].id }}</router-link
./components/Repository.vue:67:          <router-link :to="{ name: 'documentation', hash: '#FTP-access' }">FTP server</router-link
./components/Repository.vue:179:                    >FTP server</router-link
./components/Documentation.vue:93:              >GEM Repository</router-link
./components/Documentation.vue:510:              >GEM Repository</router-link
./components/NotFound.vue:31:            <b>new version of Metabolic Atlas</b> </router-link
./components/NotFound.vue:34:          <router-link :to="{ name: 'gems' }"><b>GEM > Repository</b></router-link>
./App.vue:163:            Privacy Notice and Terms of Use </router-link
```

File: frontend/src/components/About.vue [hunk 5]
```diff
@@ -6,48 +6,139 @@
...
+            <a href="https://proteinatlas.org" target="_blank">The Human Protein Atlas</a>,
+            <a href="https://www.ensembl.org/" target="_blank">Ensembl</a>,
+            <a href="https://www.uniprot.org" target="_blank">UniProt</a>,
+            <a href="https://europepmc.org" target="_blank">Europe PMC</a>,
+            <a href="https://identifiers.org" target="_blank">Identifiers.org</a>,
+            <a href="https://www.ebi.ac.uk/chebi" target="_blank">ChEBi</a> and other essential
+            sources of information:
+            <a href="http://www.genome.jp/kegg/" target="_blank"
+              >KEGG: Kyoto Encyclopedia of Genes and Genomes</a
+            >, <a href="http://www.hmdb.ca/" target="_blank">The Human Metabolome Database</a>,
```
nanjiangshu, line 95: It looks weird here as well for `<a>`.

frontend/src/components/About.vue
```diff
@@ -141,34 +262,41 @@ export default {
...
+        },
+        {
           name: 'Dr. Rui Benfeitas',
           img: require('../assets/pic-rui.jpg'),
           orcid: 'https://orcid.org/0000-0001-7972-0083',
           linkedin: 'https://www.linkedin.com/in/ruibenfeitas/',
           github: 'https://github.com/Benfeitas',
-          content: '<a href="https://www.benfeitas.net" target="_blank">Senior Bioinformatician at NBIS, SciLifeLab</a>',
+          content:
+            '<a href="https://www.benfeitas.net" target="_blank">Senior Bioinformatician at NBIS, SciLifeLab</a>',
```
nanjiangshu, line 293: This kind of line breaking does not make the code prettier, since the line after breaking is still longer than 80 characters. I would think the original one looks better. Maybe it is not so easy for Prettier to handle such cases.

MalinAhlberg: Thanks for spotting all the weird formatting, @nanjiangshu and @mihai-sysbio . I think most cases can be fixed by improving our prettier configuration (eg the whitespace sensitivity). It will be updated!

frontend/src/components/explorer/gemBrowser/Gene.vue
```diff
@@ -1,40 +1,43 @@
...
-    :interaction-partner="true" :viewer-selected-i-d="gene.id"
+    component-type="gene"
+    :component-name="gene.geneName"
+    :external-dbs="gene.externalDbs"
+    query-component-action="genes/getGeneData"
+    :interaction-partner="true"
+    :viewer-selected-i-d="gene.id"
   >
-    <template v-slot:table>
+    <template #table>
```
nanjiangshu, line 10: Is that fix done manually (and why) or by the Prettier?

frontend/src/components/explorer/mapViewer/Svgmap.vue
```diff
@@ -108,207 +114,208 @@ export default {
...
   },
   created() {
-    this.updateURLCoord = debounce(this.updateURLCoord, 150);
+    this.updateURLCoord = debounce(this.updateURLCoord, 150)
   },
   async mounted() {
-    const self = this;
-    ['.met', '.enz', '.rea', '.subsystem'].forEach((aClass) => {
+    const self = this
+    ;['.met', '.enz', '.rea', '.subsystem'].forEach((aClass) => {
```
nanjiangshu, line 132: I guess this ';' can be removed?

nanjiangshu (commented):
Hi @MalinAhlberg and @e0 , the linter integration works great when I tested, both with the automated linting and manual linting. I think it would be great if the instructions about how to use linting can be documented at some better place then this PR description. 

I have also tested the overall functions of the web-site and did not find any problem so I guess Prettier does break any functions despite big changes of the code formatting. 

For the editor integration, in my case `vim`, I'm not able to make it work. But as I already mentioned last Friday, this is just optional. 

Except for the minor comments I posted. I think this PR looks great now.

MalinAhlberg:
I created #801, for going through the configurations of prettier and eslint. So this PR will only add the integrations, not enforce any new standard.

frontend/src/components/explorer/gemBrowser/Gene.vue
MalinAhlberg: This was done to make eslint (`--fix`), but is [now restored](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/797/commits/0cef748835c1044dc72db8e9b0a045017f426234#diff-e885f43c224f5da447be0fb7dc905906bdc760dcdb5b1bc2d3c64268de83bb93R10).
The [corresponding rule is disabled](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/797/commits/d02abb96e4be15e989eacb13de0d23d401b15ea9) for now.

e0:
> I created #801, for going through the configurations of prettier and eslint. So this PR will only add the integrations, not enforce any new standard.

Thanks for creating this new issue! Sometimes we have issues grow into something bigger and it really helps to create new issues when appropriate given our current point system.

@MalinAhlberg I think you had a list of things that should be done before completing this PR. Just wanted to note that we also mentioned that we should add some instructions about how to enable text-editor integration in the documentation. If you want to share the todo list here please add this documentation update as a todo as well. I'm happy to help out with any of the remaining todos, just let me know!

frontend/src/App.vue
e0: Fixed here: https://github.com/MetabolicAtlas/MetabolicAtlas/pull/797/commits/a21f080b80e8b59f661b064710407d0b5b947a34#diff-347c85641aa3243ab418a0dfededf202763a1d5a28dc34eacb00b912b1a97fe4R5

frontend/src/components/About.vue
e0: Fixed in: https://github.com/MetabolicAtlas/MetabolicAtlas/pull/797/commits/a21f080b80e8b59f661b064710407d0b5b947a34#diff-347c85641aa3243ab418a0dfededf202763a1d5a28dc34eacb00b912b1a97fe4R5

MalinAhlberg:
> @MalinAhlberg I think you had a list of things that should be done before completing this PR.

Most things on the list are already fixed (:balloon:) but here's what's left:

- [ ] add to `/api` (maybe in a separate, later pr?)
- [ ] document about how to use in editor
- [ ] re-enable error browser overlay (I have this commit ready, just haven't pushed yet)
- [ ] test using the website

Feel free to update this list if you come to think about more things.

e0:
Thanks for adding the list @MalinAhlberg!

I have glanced through the PR and think everything looks good in terms of formatting. If there is no major problems for anyone else (please keep in mind that we now have a separate issue to go over the changes), I think we can start adding the `/api` integration as well, where we should be able to use a lot of the things we did in the frontend.

Tomorrow I will try to start with the documentation for editor integration.

MalinAhlberg:
> Would it be possible to perform the linting in GH actions via our own Dockerfile(s)? I'm thinking it would be a lot easier to maintain consistency this way by the same setup between what we have running locally and what the workflow does. For example, the node version is now defined in both places (dockerfile and workflow).

I've spent some time trying out different ways of doing this. I can't say that of my attempts are better than what we have now. If anyone is interested in seeing them, let me know. If anyone wants to give it a go on their own, that would be nice. Otherwise, I wonder if we shouldn't stick to running the actions without Docker.

e0:
> I've spent some time trying out different ways of doing this. I can't say that of my attempts are better than what we have now. If anyone is interested in seeing them, let me know. If anyone wants to give it a go on their own, that would be nice. Otherwise, I wonder if we shouldn't stick to running the actions without Docker.

I haven't had much experience with this and to avoid this issue from growing bigger than it already has, I think it's ok to leave things the way they are for now. I actually thought that this initial issue was only supposed to have local integration so having the GitHub actions is already a bonus imo.

e0 (approved):
Looks great @MalinAhlberg! Thanks for taking the initiative to create more issues as well!

inghylt:
I really enjoy the new functionality introduced here, good job @MalinAhlberg @e0 

Just a thought, autoformatting and running eslint on commit is only working when there are containers running. No surprise since we are dependent on docker but something to be aware of since developers might not always have docker running on their local setup when doing commits. Good thing to have this in combination with GitHub actions so no errors slip :+1:

inghylt (approved):

e0:
> Just a thought, autoformatting and running eslint on commit is only working when there are containers running. No surprise since we are dependent on docker but something to be aware of since developers might not always have docker running on their local setup when doing commits. Good thing to have this in combination with GitHub actions so no errors slip 👍

Good that you had this in mind @inghylt. It is briefly mentioned [here](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/797/commits/5c7d44a0d2e3722547c88148fd8d99f924c97b04#diff-98e64bc1cd2db9333c6effe87bbe0d6dfe8714aba4c6bde45aa037fe0796e44cR39), but it's specifically about the commit hooks. The idea is that once the commit hooks are setup the user wouldn't need to run the commands manually. Do you think this is enough though?

inghylt:
One could perhaps add a check to see that containers are running and if not display a message or something, but I think things are fine as they are now :+1:

.github/CONTRIBUTING.md
```diff
@@ -28,6 +28,39 @@ Once you have the project running locally, create a new branch based on the `dev
...
+
+Many text editors have support for integrating `eslint` and `prettier`. For `eslint`, the editor could show inline warnings and errors live. For `prettier`, the editor can auto-format the code whenever a file is saved.
+
+##### prettier
+
+Prettier relies on the `.prettierrc.json` file, which is provided by default in the project. Please refer to the plugin documentation for more information on how to set them up. For example: [ale](https://github.com/dense-analysis/ale) and [coc.nvim](https://github.com/neoclide/coc.nvim) for Vim or [the Prettier extension](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode) for Visual Studio Code.
+
+##### eslint
+
+To enable the `eslint` integrations, the text editor needs to have access to the `node_modules` folder. Since the project runs in docker, the `node_modules` are not mounted to the local file system by default for performance reasons. Therefore, to enable text editor integrations, the `node_modules` folder needs to be mounted, which can be done by making the following changes (this is for `frontend`, but it can also be added to `api` for the same effect) in `docker-compose-local.yml`.
```
mihai-sysbio, line 51:
```suggestion
To enable the `eslint` integrations, the text editor needs to have access to the `node_modules` folder. Since the project runs in Docker, the `node_modules` are not mounted to the local file system by default for performance reasons. Therefore, to enable text editor integrations, the `node_modules` folder needs to be mounted, which can be done by making the following changes (this is for `/frontend`, but it can also be added to `/api` for the same effect) in `docker-compose-local.yml`.
```

.github/CONTRIBUTING.md
```diff
@@ -28,6 +28,39 @@ Once you have the project running locally, create a new branch based on the `dev
...
+
+Under `frontend:`:
+
+- change `command:` from `yarn serve` to `yarn && yarn serve`.
+- under `volumes:`, add a new line: `- ./frontend/node_modules:/project/node_modules`
```
mihai-sysbio, line 56: It feels easier to have a dangling branch with these changes, so that one can both easily compare and also cherry-pick or merge.

.github/workflows/prettier.yml
```diff
@@ -0,0 +1,30 @@
+name: "Prettier"
+
+on:
+  pull_request:
+    branches: [develop]
```
mihai-sysbio, line 5: I'd be tempted to allow this to run on all non-`main` branches, so it can run also on:
- direct commits in `develop`
- commits on other branches prior to creating a PR, so that one has a chance to fix things if necessary prior to a PR creation

mihai-sysbio:
> I actually thought that this initial issue was only supposed to have local integration so having the GitHub actions is already a bonus imo.

This PR is meant to close #604 which says:
> In a PR: As a developer, I would like to have lint checks run when pushing to GitHub, so that lint warnings are discovered and fixed asap.

@e0 from this I understand that having a GH Action should be expected from this PR, unless maybe there have been some changes (other issues created) that I'm not aware of.

mihai-sysbio:
Very nice work everyone, both for the implementation and the testing. This (and the previous) PR have had good discussions, which was nice to see, and bring some highly-appreciated comforts to the development workflow. 

I'd like a chance to review this, but I'm not sure I'll be able to go through the process before the sprint ends, so it looks like I'll have to review retrospectively. I'll provide the more important feedback now and create a new issue if needed after the hands-on.

1. I am still confused about the changes in the configurations for eslint and prettier. It looks like we are still using custom rules, which was the opposite of the desired goal? And if so, it feels like our previous configuration was mostly based on the Airbnb, which is also quite popular, so then why are dropping that standard? There should be easy ways to have that standard working with prettier as well I think https://www.npmjs.com/package/prettier-airbnb-config
Especially that #801 exists, I would imagine that we aim to keep the linter changes minimal in this PR, rather than changing now and maybe reverting some of the changes later.

2. There was a separate issue to update the dependencies in the `.lock` files. I understand why these files needed to be updated because of new dependencies, but I think it should be possible to make changes more restrictive. Last time when I updated all dependencies en-masse we ended up having problems with the 3D viewer, so doing such updates require more hands-on testing of the web app, which is why #719.

.github/CONTRIBUTING.md
e0: That would be more convenient but I'm worried that it would make it more likely for those changes to be checked into version control in other branches by mistake. Happy to hear more input on this from others though.

e0:
> > I actually thought that this initial issue was only supposed to have local integration so having the GitHub actions is already a bonus imo.
> 
> This PR is meant to close #604 which says:
> 
> > In a PR: As a developer, I would like to have lint checks run when pushing to GitHub, so that lint warnings are discovered and fixed asap.
> 
> @e0 from this I understand that having a GH Action should be expected from this PR, unless maybe there have been some changes (other issues created) that I'm not aware of.

That's true. I remember that we discussed about this during issue planning and someone (maybe @inghylt?) suggested that we could just focus on local setup initially. Maybe I missed when we decided against it or we forgot to make a note of that? I should/would probably not have voted a `3` if GitHub actions are included though.

e0:
> 1. I am still confused about the changes in the configurations for eslint and prettier. It looks like we are still using custom rules, which was the opposite of the desired goal? And if so, it feels like our previous configuration was mostly based on the Airbnb, which is also quite popular, so then why are dropping that standard? There should be easy ways to have that standard working with prettier as well I think https://www.npmjs.com/package/prettier-airbnb-config
>    Especially that [Update configs for prettier and eslint #801](https://github.com/MetabolicAtlas/MetabolicAtlas/issues/801) exists, I would imagine that we aim to keep the linter changes minimal in this PR, rather than changing now and maybe reverting some of the changes later.

I agree that this was the original intent. When working on this issue, I think it was discovered that the amount of work was more than expected. For situations like this, I think it makes sense for the issue to be broken down into smaller ones so we can keep story points sensible and make incremental progress. @MalinAhlberg made a great job at carving out new issues while making the scope of this PR focused and still useful to be merged.

Previously we had a lot of code that did not follow any coding standards, since the formatter was not as powerful. We tried to keep a lot of things the same for now, but there are certain things that are difficult to configure relying only on prettier. So while there are a lot of file changes, it would have been a lot more if we completely removed all the custom rules. It might also require a lengthy discussion to go through all of the rules, so I think it makes a lot of sense that we now have a separate issue for it. @mihai-sysbio there has also been some less informal discussion about this on Slack, I will try to ping you there about them.

---

> 2. There was a separate issue to update the dependencies in the `.lock` files. I understand why these files needed to be updated because of new dependencies, but I think it should be possible to make changes more restrictive. Last time when I updated all dependencies en-masse we ended up having problems with the 3D viewer, so doing such updates require more hands-on testing of the web app, which is why [resolve vulnerabilities #719](https://github.com/MetabolicAtlas/MetabolicAtlas/issues/719).

From what I can see, there are quite few change in the frontend `yarn.lock` file and they seem to be related to the new dependencies. There is a few I didn't recognize but I think they might be related nested dependencies. I agree that the `api/yarn.lock` file should not be included in this PR since we have a new issue for it.

nanjiangshu:
Hi @MalinAhlberg I'm a bit confused with the commit 0cef748, is that just to fix the ugly linting for e.g. router-link or want to undo all code fixes by linting, as you mentioned that this PR is just for lint integration, but for applying linting. However, the code has still been changed in many files.

e0:
> Hi @MalinAhlberg I'm a bit confused with the commit 0cef748, is that just to fix the ugly linting for e.g. router-link or want to undo all code fixes by linting, as you mentioned that this PR is just for lint integration, but for applying linting. However, the code has still been changed in many files. 

Please read the comment above yours.

mihai-sysbio:
I think in a way maybe @nanjiangshu and I are wondering about the same thing. Let me be more specific: `.eslintrc.js` references `'@vue/airbnb'`, but I don't see the [Airbnb style guide](https://github.com/airbnb/javascript) being referenced in by `prettier` configuration.

e0:
> I think in a way maybe @nanjiangshu and I are wondering about the same thing. Let me be more specific: `.eslintrc.js` references `'@vue/airbnb'`, but I don't see the [Airbnb style guide](https://github.com/airbnb/javascript) being referenced in by `prettier` configuration.

For this specific case, I would want to argue for removing `@vue/airbnb` in #801. Nothing is done about it for now because this is a part of keeping things as similar as possible to the state before this issue. Does that make sense?

nanjiangshu:
> Please read the comment above yours.

Thanks for the explanation @e0 and sorry that I have missed your comment before I posted. I think it makes sense to split the task before it grows too big.

mihai-sysbio:
> For this specific case, I would want to argue for removing `@vue/airbnb` in #801. Nothing is done about it for now because this is a part of keeping things as similar as possible to the state before this issue. Does that make sense?

I see, but wouldn't the PR be smaller if that standard would be still adopted for now? And then talk about removing it (which not yet fully decided) in #801 rather than reviewing changes now, and reviewing again if it's decided to keep it? I know it's yet more work in an already large PR.

e0:
> > For this specific case, I would want to argue for removing `@vue/airbnb` in #801. Nothing is done about it for now because this is a part of keeping things as similar as possible to the state before this issue. Does that make sense?
> 
> I see, but wouldn't the PR be smaller if that standard would be still adopted for now? And then talk about removing it (which not yet fully decided) in #801 rather than reviewing changes now, and reviewing again if it's decided to keep it? I know it's yet more work in an already large PR.

We have two tools now, `eslint` and `prettier`, all of the code formatting is done using `prettier`. Some of the things we have in `.eslintrc.js` are formatting rules, but some are for code quality rules, which `prettier` does not care about. Since all code is being formatted by `prettier` in this PR, some of the rules from `.eslintrc.js` has no effect. We have removed some custom rules from there, which when removed, do not produce any errors or warnings after formatting with `prettier`. At the same time we don't want to make the `prettier` config over complicated since it can't cover everything that is in `@vue/airbnb`. See [this article](https://prettier.io/docs/en/comparison.html) for more information on the differences.

e0:
>  I agree that the `api/yarn.lock` file should not be included in this PR since we have a new issue for it.

Addressed in: https://github.com/MetabolicAtlas/MetabolicAtlas/pull/797/commits/021cf167db5f8748797679e80278720f0c707ef3

nanjiangshu (approved):
Looks good to me now.

inghylt:
I think this PR is ready to be merged no? And regarding the discussions about using MetabolicAtlas Docker files or images in GitHub Actions, #804 and #805 have been created

mihai-sysbio:
I guess this question is mainly addressed to me? Given the approving reviews, I too think it should be merged. I'd just want to mention the 2 unresolved minor discussions.
And thanks for creating those issues @inghylt .

inghylt:
More intended as a general questions since the sprint is ending soon!

Regarding the minor discussions, personally I don't really have anything to add to either of them. I'm fine with keeping things as is, but if anyone feels strongly about making the suggested changes I'm fine with that as well. Maybe wise to create issues for them in that case to be able to close this PR asap

.github/workflows/prettier.yml
MalinAhlberg: I think this is a good suggestion, I added issue #809.

.github/CONTRIBUTING.md
MalinAhlberg: Good suggenstion. I added #810 so where we can discuss more about this.

Issue #801, opened 2021-12-13
Update configs for prettier and eslint


MalinAhlberg:
**Please describe the feature**

As a developer, I want the rules for the linter and formatter to be updated so that our code base follows the current standards, looks good and is easy to work with.

**Describe the solution you would like**

When adding prettier (#797), the rules for prettier and eslint were set to leave the code base as intact as possible.

The list of things that could be updated / added includes:
- vue/multi-word-component-names
- vue/no-useless-template-attributes
- vue/no-lone-template,
- vue/order-in-components
- vue/v-slot-style

PR #803, opened 2021-12-14
feat: lint setup for api

### LINKED ISSUE #604: Add lint integration

MalinAhlberg:
This PR closes #604  and relates to #797 .

**Type of change**  

- [X] Other
 
**List of changes made**  

This pr sets up prettier and eslint for `/api`, and formats the code accordingly.
Note that many eslint rules are disabled.  This should be addressed in #801.

The `git_hooks` file is slightly modified so that it can easily be merged with the version in #797 .

**Testing**  

*NB*: Remember to rebuild and to run `git config core.hooksPath git_hooks`, even if you have done this earlier.

***Autoformatting and running eslint on commit***
Check out this branch and

- rebuild the stack
- `git config core.hooksPath git_hooks`
- open a file, eg `api/src/index.js`, and add something which you think should cause an syntactic error. For example, join the first two lines
- `git add api/src/index.js` and then commit it. Open the file again and make sure that your error has been fixed (or that you got an error message if it couldn't be fixed automatically).

***Running eslint and prettier manually***
- `source proj.sh` (or replace `ma-exec` with `docker compose -f docker-compose.yml -f docker-compose-local.yml` in the following commands)
- `ma-exec api yarn lint` to lint all `js` and `vue` files
- `ma-exec api yarn check-format` to print all files that should be reformatted
- `ma-exec api yarn format` to format all files

**Further comments**  

Note: if you need to make a commit without running the prehook, you can use `git commit --no-verify`. Useful if you for example run into problems with docker but still want to commit your code.

**Checklist**  


MalinAhlberg:
Merging this with #797 will require a bit of manual editing for  `.github/workflows/prettier.yml` and `git_hooks/pre-commit`.

e0:
Thanks for preparing this PR @MalinAhlberg. I tried it now locally and it seems to work great! Since some additional work needs to be done to prevent merge conflicts with #797 as you mentioned, I will wait with the approval for now until #797 is merged.

It's really nice to see how much you could derive from the `frontend` changes in this PR. Great work with the GitHub actions and local git hooks. Feels like this setup should be easy to adopt into our other repos such as https://github.com/MetabolicAtlas/data-generation and https://github.com/MetabolicAtlas/3d-network-viewer

inghylt:
Seems to be working!

Perhaps silly question, but why no "Running eslint and prettier manually" as in #797?

e0:
Just wanted to link to the nice suggestion from @inghylt: https://github.com/MetabolicAtlas/MetabolicAtlas/pull/797#issuecomment-993686339

Since the `git_hooks/pre-commit` needs to be manually before merging this PR, maybe this change be added as well.

MalinAhlberg:
> Perhaps silly question, but why no "Running eslint and prettier manually" as in #797?

Not a silly question! But the answer is a bit silly (I forgot). It's in the PR description now. But perhaps it should also be added to the contribution guidelines (as [here](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/797/commits/5c7d44a0d2e3722547c88148fd8d99f924c97b04?short_path=98e64bc#diff-98e64bc1cd2db9333c6effe87bbe0d6dfe8714aba4c6bde45aa037fe0796e44c)) or README?

inghylt:
@MalinAhlberg `ma-exec` does not exist in `proj.sh` in this branch, I guess it should be added there as well?

And yes makes sense to add a line in the contribution guidelines

e0:
> @MalinAhlberg `ma-exec` does not exist in `proj.sh` in this branch, I guess it should be added there as well?

I should have added a comment about this earlier. @MalinAhlberg and I were discussing whether to create a new PR or to include these in #797. Since reviewing the diff in GitHub seems to be slow when there are too many files in a single PR, and also that it might be easier cognitively for reviewers, we decided that it would be easier to have a new PR instead. So this PR might make more sense to be reviewed after #797 is merged, then we would already have the `ma-exex` in the project for example.

e0:
Setting this PR to draft mode for now until the conflicts are resolved.

api/.prettierignore
```diff
@@ -0,0 +1,6 @@
+public
+dataOverlay
+Dockerfile
+package*
+repository
+svg
```
nanjiangshu, line 6: I'm not quite sure if json files at `api/src/data` should be ignored by linter or not. These files are copied from the repo `data-files`. If we want to run prettier for these files as well, then the changes should be made in the `data-files`.

nanjiangshu (commented):
Great work @MalinAhlberg and @e0! I worked well when I tested for both auto linting and manual linting. And I did not find any weird formatting when I browsed all modified files. I just left a minor comment regarding linting of json files in the folder `api/src/data`. 

This PR looks good to me when the merging conflict is solved (if there is any).

api/.prettierignore
MalinAhlberg: Thanks for pointing that out! I think it sounds reasonable to add `src/data` to the ignore list. I can push a commit for that, if @ea agrees?

But speaking of which, I don't think the ignore list will be used when running prettier from the precommit hook. If we are bothered by that, we could add it as a separate issue.

e0: I just realized that we have `api/src/data/*.json` in the `.gitignore` file so I don't think this would cause any problems with the precommit hook since those files would never be picked up by git.

api/package.json
```diff
@@ -7,7 +7,9 @@
   "scripts": {
     "start": "webpack && node ./dist/bundle.js",
     "develop": "webpack --watch",
-    "lint": "eslint ./src --fix"
+    "lint": "eslint src --ext .js,.vue",
```
e0, line 10: I don't think we need `,.vue` here since there shouldn't be any vue files in the `api`. Not a big deal either way though.

e0 (approved):
Looks excellent @MalinAhlberg! I tried all of the commands you listed as well as making a test commit to trigger formatting. Everything worked great.

api/.prettierignore
MalinAhlberg: Ah, good point@e0. Added in cf5058592f20f9fc84a3521481ace270ce0a545f

api/package.json
MalinAhlberg: True! It's removed in 4aac31465742e3d0053baa1e5d6304d653e508fb

PR #825, opened 2022-01-05
fix: broken proj.sh

### LINKED ISSUE #824: Broken proj.sh
**Describe the bug**

Many of the commands in `proj.sh` are not working.

**Steps to reproduce**

Try to run normal commands such as `build-stack`, note that it will not work due to `$CHOSEN_ENV` not being set.

**Expected behavior**

A functioning `proj.sh`, where all commands work irrespective of order.

inghylt:
This PR closes #824

**Type of change**  

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
 
**List of changes made**  

Added CHOSEN_ENV as exported global variable to `proj.sh` since this is required both by `proj.sh` and `docker-compose.yml`

**Testing**  

Try to run all commands in `proj.sh`to make sure everything is functional

**Further comments**  

**Checklist**  

### DIFF
--- a/proj.sh
+++ b/proj.sh
@@ -2,6 +2,7 @@
 export PATH=$PATH:/usr/local/bin
 LOCALENV="local"
 METATLAS_DEFAULT_ENV="env-${LOCALENV}.env"
+export CHOSEN_ENV=$METATLAS_DEFAULT_ENV
 
 function generate-data {
   # enable flag "-q" to force overwritting existing data files
--- a/proj.sh
+++ b/proj.sh
@@ -2,7 +2,7 @@
 export PATH=$PATH:/usr/local/bin
 LOCALENV="local"
 METATLAS_DEFAULT_ENV="env-${LOCALENV}.env"
-export CHOSEN_ENV=$METATLAS_DEFAULT_ENV
+CHOSEN_ENV=$METATLAS_DEFAULT_ENV
 
 function generate-data {
   # enable flag "-q" to force overwritting existing data files
--- a/proj.sh
+++ b/proj.sh
@@ -43,7 +43,7 @@ function logs {
 }
 
 function ma-exec {
-  docker compose -f docker-compose.yml -f docker-compose-local.yml exec $@
+  docker compose --env-file $CHOSEN_ENV  -f docker-compose.yml -f docker-compose-local.yml exec $@
 }
 
 function deploy-stack {
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -11,7 +11,10 @@ services:
   api:
     container_name: api
     build: api
-    env_file: ${CHOSEN_ENV}
+    environment:
+      - NODE_ENV=${NODE_ENV}
+      - NEO4J_USERNAME=${NEO4J_USERNAME}
+      - NEO4J_PASSWORD=${NEO4J_PASSWORD}
     depends_on: 
       - neo4j
     networks:


proj.sh
```diff
@@ -2,6 +2,7 @@
 export PATH=$PATH:/usr/local/bin
 LOCALENV="local"
 METATLAS_DEFAULT_ENV="env-${LOCALENV}.env"
+export CHOSEN_ENV=$METATLAS_DEFAULT_ENV
```
MalinAhlberg, line 5: If `CHOSEN_ENV` is set before `proj.sh` is read, the variable will be overwritten by this line.
```sh
$ export CHOSEN_ENV=env-dev.env
$ source proj.sh
$ echo $CHOSEN_ENV 
env-local.env
```

Something like this would avoid that:
```sh
if [ -z "$CHOSEN_ENV" ];
then
  export CHOSEN_ENV=$METATLAS_DEFAULT_ENV
fi
```

MalinAhlberg: Maybe this is a bit overkill, but it could be nice to have function to set the `CHOSEN_ENV`, so that you can easily change it:
```sh
$ set-env dev
Using env-dev.env
```
The code:
```sh
function set-env {
  export CHOSEN_ENV="env-${1}.env"
  echo "Using $CHOSEN_ENV"
}
```

MalinAhlberg (commented):
Good fix! I have not tested to do a full dev deploy, though.

I added two comments, but I think the PR is fine as it is.

nanjiangshu:
I'm a bit confused with the settings of ENVs in `docker-compose.yml`. Why `env_file: ${CHOSEN_ENV}` is set only for the container `api`? 

In addition, I feel it a bit weird that we save sensitive info such as`NEO4J_PASSWORD` as plain text even for the production deployment.

mihai-sysbio:
> I'm a bit confused with the settings of ENVs in `docker-compose.yml`. Why `env_file: ${CHOSEN_ENV}` is set only for the container `api`?

When building, Docker needs to know the value of some variables. The option was to store all build/deployment variables in a single environment file.

> In addition, I feel it a bit weird that we save sensitive info such as`NEO4J_PASSWORD` as plain text even for the production deployment.

The production env file is not committed, would you think that is alright?

proj.sh
e0: This is a good catch @MalinAhlberg! Just to double check, I think the problem you are illustrating could be problematic for when deploying right? Since the current first line in the `deploy-stack` function only sets the `CHOSEN_ENV` but does not export it, it would mean that the docker compose files could end up with the wrong `CHOSEN_ENV`?

I like your `set-env` function and I think if we use that whenever we want to set the `CHOSEN_ENV`, on line 5 and inside the `deploy-stack` function, then we should be safe and would not need to have the `if - then` statement?

e0:
> The production env file is not committed, would you think that is alright?

I think it's ok for now but it was a good point @nanjiangshu brought up. We could explore using `docker secret` if everyone is onboard.

proj.sh
MalinAhlberg: I think it would still overwrite the old env variable when reading `proj.sh`, no? The code below would keep the old value (but is weird..?):
```sh
function set-env {
  export CHOSEN_ENV="env-${1-LOCALENV}.env"
  echo "Using $CHOSEN_ENV"
}

# if not set, set chosen_env to the default (localenv)
set-env=$CHOSEN_ENV
```
But the more I look at `proj.sh`, the more unsure I get of how we really want it to work.  Since the `CHOSEN_ENV` is exported, I think it should be possible to set it from outside of this script, and still keep that variable set when reading/loading `proj.sh`. So you should be able to go
```sh
export CHOSEN_ENV=another_env
source proj.sh
build-stack
````
and expect docker compose to work with `another_env`? Or maybe `deploy-stack` is the only place where the env should be able to differ from `local`?

Comment (e0) [reply proj.sh:5]: I think it would still overwrite the old en variable but I think that's ok.

The env is set globally on line 5 and this should be the default when working locally.

When deploying, the env should be set to the target environment (`dev` or `prod`) at the beginning and reset back to the default (`local`) at the end.

In other words, the only time when the env should be something other than `local` is during deploy. So I think it's ok to overwrite the env globally whenever we source the `proj.sh` file. Does that answer your question @MalinAhlberg ?

**Comment by nanjiangshu** [comment]
> > I'm a bit confused with the settings of ENVs in `docker-compose.yml`. Why `env_file: ${CHOSEN_ENV}` is set only for the container `api`?
> 
> When building, Docker needs to know the value of some variables. The option was to store all build/deployment variables in a single environment file.
> 
I understand that the Docker container of the API needs to know some variables and that's why the env_file is set here. What I'm asking is the variables such as https://github.com/MetabolicAtlas/MetabolicAtlas/blob/develop/docker-compose.yml#L24-L30 will be obtained from the default env file if `evn_file` is not set for the container `neo4j`. There will not be any problem if we always have the same value for these variables for different ENV files, but it will cause a problem if they are different.  


> > In addition, I feel it a bit weird that we save sensitive info such as`NEO4J_PASSWORD` as plain text even for the production deployment.
> 
> The production env file is not committed, would you think that is alright?

Sure the ENV file is not committed. But this is just like you will not save your email password as plain text on your computer, or do you? :smirk:

File: proj.sh (hunk 1 above)
Comment (MalinAhlberg) [reply proj.sh:5]: Yes, thanks for the explanation! I guess it will work well for all our intentions, then :+1:

**Review by e0 — APPROVED** [review]
This seems to work well locally. I haven't tested deploying but that was working fine since before so hopefully no problems are introduced there.

The points mentioned by @nanjiangshu are great and I think we could discuss and create some new issues based on them.

**Comment by mihai-sysbio** [comment]
There seems to be a lot of discussion around `CHOSEN_ENV`. Based on @nanjiangshu's [comment above](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/825#issuecomment-1010067616), how about we treat the need for variables in `api` the same way as in `neo4j`? My shallow testing shows this can work. Have a look at this commit https://github.com/MetabolicAtlas/MetabolicAtlas/commit/0e071eca1325264abf6b504e2234d66eaba4e33a . With this, hopefully the issue of having to export `CHOSEN_ENV` goes away.

File: proj.sh (hunk 1 above)
Comment (mihai-sysbio) [reply proj.sh:5]: I've added my two cents [below](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/825#issuecomment-1010146492).

**Comment by mihai-sysbio** [comment]
> Sure the ENV file is not committed. But this is just like you will not save your email password as plain text on your computer, or do you? 😏

@nanjiangshu I'm not sure we should aim to build up fences to prevent people from doing this. I'm up for any suggestions to improve the sharing of ENV files, but I'm also satisfied enough with the current setup that I don't feel the need to create an issue for it.

**Comment by nanjiangshu** [comment]
My concern about `Why env_file: ${CHOSEN_ENV} is set only for the container api` is invalid since the env_file is set already by the flag `--env-file $CHOSEN_ENV` in the command defined in `proj.sh`. Sorry for the confusion that I brought.  

I almost never use `proj.sh` to run stack, instead I use 
```docker compose --env-file env-local.env -f docker-compose.yml -f docker-compose-nj.yml up -d ``` 
since otherwise I will get several problems such as 
```  modified:   api/yarn.lock
```
and not able to access the web-server remotely. 

I think the fix by @inghylt is great and necessary.

As to my second concern regarding the saved password in plain text file, I agree with @mihai-sysbio that it is so far not that critical since login to the neo4j database from the web-browser is not available on the production server.

nanjiangshu:
By the way, the definition of `ma-exec`
https://github.com/MetabolicAtlas/MetabolicAtlas/blob/develop/proj.sh#L45
might be better to changed to 

```docker-compose
docker compose --env-file $CHOSEN_ENV  -f docker-compose.yml -f docker-compose-local.yml exec $@
```

But it might be unrelated to this issue.

inghylt:
41a3f25f9173a0b4fbf2e3d422dc85bddf672124 should cover what we decided yesterday. Please verify locally. Someone should also try to deploy

nanjiangshu:
Looks great with the commit 41a3f25 @inghylt. I can deploy locally without any problem. However I was not able to deploy to the dev server with the following error 
```
error An unexpected error occurred: "https://registry.yarnpkg.com/three/-/three-0.126.1.tgz: getad│····
drinfo EAI_AGAIN registry.yarnpkg.com".  
```
I don't  think the problem is related to this PR since it got the same problem on the `develop` branch.
If more people can confirm the problem we should create an issue for that. 

In addition, I would prefer the tiny problem regarding `ma-exec` https://github.com/MetabolicAtlas/MetabolicAtlas/pull/825#issuecomment-1010826385 can be fixed any well. That command is needed when we run test.

inghylt:
Thanks for raising that @nanjiangshu, it has been added in commit 53fe845d0444f7381e40786cbeb5e486c969c411

Just to make sure, did you resource proj.sh before you moved to develop branch and tried to deploy?

nanjiangshu:
> Just to make sure, did you resource proj.sh before you moved to develop branch and tried to deploy?

Yes. And I have also exported the CHOSEN_ENV 

`export CHOSEN_ENV=env-dev.env `

e0:
> Looks great with the commit [41a3f25](https://github.com/MetabolicAtlas/MetabolicAtlas/commit/41a3f25f9173a0b4fbf2e3d422dc85bddf672124) @inghylt. I can deploy locally without any problem. However I was not able to deploy to the dev server with the following error
> 
> ```
> error An unexpected error occurred: "https://registry.yarnpkg.com/three/-/three-0.126.1.tgz: getad│····
> drinfo EAI_AGAIN registry.yarnpkg.com".  
> ```
> 
> I don't think the problem is related to this PR since it got the same problem on the `develop` branch. If more people can confirm the problem we should create an issue for that.
> 
> In addition, I would prefer the tiny problem regarding `ma-exec` [#825 (comment)](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/825#issuecomment-1010826385) can be fixed as well. That command is needed when we run test.

Thanks for reporting this and testing on the develop branch too @nanjiangshu ! I got some similar problems a few days ago and got this exact error message again now (tried deploying twice). If I visit https://registry.yarnpkg.com/three/-/three-0.126.1.tgz directly in the browser it looks like the file is there. We have had problems with yarnpkg before but they would usually go away after a bit but now it has been days. I still think this branch itself could be merged without testing the deploy though and we should create a new issue to investigate the yarnpkg issue. Alternatively we could investigate this in the update dependencies issue and even consider changing to `npm` or `pnpm`.

nanjiangshu (approved):

proj.sh
mihai-sysbio: I think this thread has been resolved now, please `unresolve` if otherwise.

PR #830, opened 2022-01-11
feat: verify api data

### LINKED ISSUE #780: verify API data
As a developer, I would like to run a test to validate a list of endpoints, so that I am confident that nothing important is broken when making changes to the codebase.

It is a part of this issue to figure out the important endpoints.
For model specific endpoints, only focus on the Human-GEM for now.

e0:
This PR closes #780.

**Type of change**  

- [x] Other
 
**List of changes made**  

This PR adds jest integration for the `api` as well as a series of tests focusing on verifying data from various endpoints.

**Screenshot of the fix**  
image

**Testing**  

- Instructions on how to test
  - Run the following after sourcing 'proj.sh', rebuilding, and starting the project: `ma-exec api yarn test`

**Further comments**  

Currently the tests are relatively basic but ensures some basic checks for most of the endpoints. Please let me know if there is any test that you think we should add.

**Checklist**  

### DIFF
--- a/api/test/compartments.test.js
+++ b/api/test/compartments.test.js
@@ -28,6 +28,6 @@ describe('compartments', () => {
     );
 
     const data = await res.json();
-    expect(data.length).toBe(238);
+    expect(data.length).toBe(NUCLEUS_INFO.reactionsCount);
   });
 });
--- a/api/test/compartments.test.js
+++ b/api/test/compartments.test.js
@@ -1,0 +1,22 @@
+import fetch from "node-fetch";
+
+describe("compartments", () => {
+  test("a compartment should have a list of subsystems", async () => {
+    const res = await fetch(
+      `${API_BASE}/compartments/nucleus?model=HumanGem&version=${HUMAN_GEM_VERSION}`
+    );
+
+    const { info, subsystems } = await res.json();
+    expect(info.subsystemCount).toBeGreaterThan(0);
+    expect(info.subsystemCount).toBe(subsystems.length);
+  });
+
+  test("a compartment should have related reactions", async () => {
+    const res = await fetch(
+      `${API_BASE}/compartments/nucleus/related-reactions?model=HumanGem&version=${HUMAN_GEM_VERSION}`
+    );
+
+    const data = await res.json();
+    expect(data.length).toBeGreaterThan(0);
+  });
+});
--- a/api/test/compartments.test.js
+++ b/api/test/compartments.test.js
@@ -1,22 +1,33 @@
-import fetch from "node-fetch";
+import fetch from 'node-fetch';
+import { validateComponent } from './util';
 
-describe("compartments", () => {
-  test("a compartment should have a list of subsystems", async () => {
+const NUCLEUS_INFO = {
+  id: 'nucleus',
+  letterCode: 'n',
+  name: 'Nucleus',
+  genesCount: 334,
+  metabolitesCount: 181,
+  reactionsCount: 238,
+  subsystemCount: 26,
+};
+
+describe('compartments', () => {
+  test('a compartment should have correct data', async () => {
     const res = await fetch(
       `${API_BASE}/compartments/nucleus?model=HumanGem&version=${HUMAN_GEM_VERSION}`
     );
 
     const { info, subsystems } = await res.json();
-    expect(info.subsystemCount).toBeGreaterThan(0);
+    validateComponent(info, NUCLEUS_INFO);
     expect(info.subsystemCount).toBe(subsystems.length);
   });
 
-  test("a compartment should have related reactions", async () => {
+  test('a compartment should have related reactions', async () => {
     const res = await fetch(
       `${API_BASE}/compartments/nucleus/related-reactions?model=HumanGem&version=${HUMAN_GEM_VERSION}`
     );
 
     const data = await res.json();
-    expect(data.length).toBeGreaterThan(0);
+    expect(data.length).toBe(238);
   });
 });
--- a/api/test/crossReference.test.js
+++ b/api/test/crossReference.test.js
@@ -1,0 +1,19 @@
+import fetch from "node-fetch";
+
+describe("cross reference", () => {
+  test("should return more than one component", async () => {
+    const res = await fetch(`${API_BASE}/external-db/HMR%202.0/HMR_1157`);
+
+    const { components } = await res.json();
+    expect(components.length).toBeGreaterThan(1);
+  });
+
+  test("should return a TSV file for a given data source", async () => {
+    const res = await fetch(`${API_BASE}/external-db/Recon3D/RE3335C`);
+
+    const {
+      externalDb: { url },
+    } = await res.json();
+    expect(url.length).toBeGreaterThan(1);
+  });
+});
--- a/api/test/crossReference.test.js
+++ b/api/test/crossReference.test.js
@@ -8,7 +8,7 @@ describe('cross reference', () => {
     expect(components.length).toBeGreaterThan(1);
   });
 
-  test('should return a TSV file for a given data source', async () => {
+  test('should contain an URL for an external reference', async () => {
     const res = await fetch(`${API_BASE}/external-db/Recon3D/RE3335C`);
 
     const {
--- a/api/test/crossReference.test.js
+++ b/api/test/crossReference.test.js
@@ -1,14 +1,14 @@
-import fetch from "node-fetch";
+import fetch from 'node-fetch';
 
-describe("cross reference", () => {
-  test("should return more than one component", async () => {
+describe('cross reference', () => {
+  test('should return more than one component', async () => {
     const res = await fetch(`${API_BASE}/external-db/HMR%202.0/HMR_1157`);
 
     const { components } = await res.json();
     expect(components.length).toBeGreaterThan(1);
   });
 
-  test("should return a TSV file for a given data source", async () => {
+  test('should return a TSV file for a given data source', async () => {
     const res = await fetch(`${API_BASE}/external-db/Recon3D/RE3335C`);
 
     const {
--- a/api/test/gemComparison.test.js
+++ b/api/test/gemComparison.test.js
@@ -1,7 +1,7 @@
-import fetch from "node-fetch";
+import fetch from 'node-fetch';
 
-describe("GEM comparison", () => {
-  test("the comparison should have reactions data", async () => {
+describe('GEM comparison', () => {
+  test('the comparison should have reactions data', async () => {
     const res = await fetch(
       `${API_BASE}/compare?models=[%7B"model":"FruitflyGem","version":"1_1_0"%7D,%7B"model":"HumanGem","version":"1_10_0"%7D]`
     );
@@ -13,7 +13,7 @@ describe("GEM comparison", () => {
     expect(Reaction[1].HumanGem).toBeGreaterThanOrEqual(Reaction[2].HumanGem);
   });
 
-  test("the comparison details should have lists of reactions and metabolites", async () => {
+  test('the comparison details should have lists of reactions and metabolites', async () => {
     const res = await fetch(
       `${API_BASE}/comparison-details?model=%7B"model":"FruitflyGem","version":"1_1_0"%7D&models=[%7B"model":"HumanGem","version":"1_10_0"%7D]`
     );
--- a/api/test/gemComparison.test.js
+++ b/api/test/gemComparison.test.js
@@ -13,7 +13,7 @@ describe("GEM comparison", () => {
     expect(Reaction[1].HumanGem).toBeGreaterThanOrEqual(Reaction[2].HumanGem);
   });
 
-  test("the comparison details should have lists of unique reactions and metabolites", async () => {
+  test("the comparison details should have lists of reactions and metabolites", async () => {
     const res = await fetch(
       `${API_BASE}/comparison-details?model=%7B"model":"FruitflyGem","version":"1_1_0"%7D&models=[%7B"model":"HumanGem","version":"1_10_0"%7D]`
     );
--- a/api/test/gemComparison.test.js
+++ b/api/test/gemComparison.test.js
@@ -1,0 +1,28 @@
+import fetch from "node-fetch";
+
+describe("GEM comparison", () => {
+  test("the comparison should have reactions data", async () => {
+    const res = await fetch(
+      `${API_BASE}/compare?models=[%7B"model":"FruitflyGem","version":"1_1_0"%7D,%7B"model":"HumanGem","version":"1_10_0"%7D]`
+    );
+
+    const { Reaction } = await res.json();
+    expect(Reaction[0].FruitflyGem).toBeGreaterThanOrEqual(
+      Reaction[2].FruitflyGem
+    );
+    expect(Reaction[1].HumanGem).toBeGreaterThanOrEqual(Reaction[2].HumanGem);
+  });
+
+  test("the comparison details should have lists of unique reactions and metabolites", async () => {
+    const res = await fetch(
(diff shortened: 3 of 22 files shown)


MalinAhlberg (commented):
This is so nice :star: ! The tests look simple and clean and easy enough to write.

I'm not very familiar with the api itself, so I can't say anything about the coverage. If you think it's a good idea, I could spend some time tomorrow looking at it to see if I can think of more tests to add. Otherwise, I think the PR could be accepted as is.

e0:
> This is so nice ⭐ ! The tests look simple and clean and easy enough to write.
> 
> I'm not very familiar with the api itself, so I can't say anything about the coverage. If you think it's a good idea, I could spend some time tomorrow looking at it to see if I can think of more tests to add. Otherwise, I think the PR could be accepted as is.

Thanks for the nice comment! We can always add more tests to this in the future as well so I don't think you need to burden yourself with exploring the api too much atm.

api/test/gemComparison.test.js
```diff
@@ -0,0 +1,28 @@
...
+    const res = await fetch(
+      `${API_BASE}/comparison-details?model=%7B"model":"FruitflyGem","version":"1_1_0"%7D&models=[%7B"model":"HumanGem","version":"1_10_0"%7D]`
+    );
+
+    const {
+      details: { CompartmentalizedMetabolite, Reaction },
+    } = await res.json();
+
+    expect(CompartmentalizedMetabolite.unique.length).toBeGreaterThan(0);
+    expect(Reaction.unique.length).toBeGreaterThan(0);
```
MalinAhlberg, line 26: maybe I'm missing something here, but I assume `Reaction.unique.length` won't actually check if all elements are unique? the description of the test ("...should have lists of unique reactions and metabolites") seems to indicate that there should be no duplicated elements.

e0: Ah, I could have probably worded the description better. I was looking at this part of the website as I was writing it. Do you think it makes sense in this context?

image

This is the response data:

image

MalinAhlberg: ah, yes, now I get it :D 
Maybe a rephrase would prevent future confusion?
`"the comparison details should have lists of reactions and metabolites"`

MalinAhlberg (approved):
As mentioned before, I think this looks very good. If you come up with some nice rephrasing for the comment I got confused about, that's nice. If you don't, that's also fine with me :)

api/test/gemComparison.test.js
e0: Addressed in https://github.com/MetabolicAtlas/MetabolicAtlas/pull/830/commits/26a6f3345c95a9162837579b02a0d04ce81fdfe3

Thanks again!

nanjiangshu (commented):
Thanks for the elegantly added test cases @e0, well done!
I think they all look very nice. 
We could probably add more tests to include 
```
/svg-thumbnail 
/hpa 
/dataOverlay
/external-db
```
as well. 

One of the most important tests we are lacking here is to verify if the returned data from the API are the same when we make changes to the API. I think this is one of the major reasons we prioritized testing. However, I don't exactly how to implement it with jest here. Maybe we need to write a testing script for every case or do you have better suggestions?

e0:
> Thanks for the elegantly added test cases @e0, well done! I think they all look very nice. We could probably add more tests to include
> 
> ```
> /svg-thumbnail 
> /hpa 
> /dataOverlay
> /external-db
> ```
> 
> as well.
> 
> One of the most important tests we are lacking here is to verify if the returned data from the API are the same when we make changes to the API. I think this is one of the major reasons we prioritized testing. However, I don't exactly how to implement it with jest here. Maybe we need to write a testing script for every case or do you have better suggestions?

Thanks @nanjiangshu ! Those are endpoints good suggestions. Also good point about more specific verification of the data. I think at one point @inghylt noticed the total number of components were wrong. We could add some hardcoded numbers to verify this but they would need to be updated each time if any version changes. Maybe that is ok though, or do we have any suggestions about if we can retrieve these numbers more dynamically?

nanjiangshu:
For total number of components, one way to verify is to check with the number of reactions or metabolites in the file `reactions.tsv` and `metabolites.tsv` for certain GEMS, such as Human-GEM or Mouse-GEM.

For other neo4j queries, such as searching, I don't think there is any reference value we can compare to.

e0:
> For total number of components, one way to verify is to check with the number of reactions or metabolites in the file `reactions.tsv` and `metabolites.tsv` for certain GEMS, such as Human-GEM or Mouse-GEM.
> 
That’s a really good idea, thanks!

e0:
@nanjiangshu I have added some new tests for the endpoints you mentioned now. Most are still quite basic but still nice to have a sanity check for them.

For the returned data, I focused on the integrated models endpoint and added the numbers (component counts and version numbers) manually for now. I initially wanted to do as you suggested but realized that files like `reactions.tsv` are not directly under the `./api` folder. I think it would be nice to have this as self-contained as possible. The component counts returned from the integrated models endpoint are calculated dynamically now (as opposed to reading direction from the json file before). I also included the version numbers here to make it easier to notice an error for whenever we have newer versions of any model.

Please give this another look when you have time. Thanks again for the great suggestions!

api/test/search.test.js
```diff
@@ -0,0 +1,21 @@
...
+    const data = await search({
+      searchTerm: "H2O",
+      model: "HumanGem",
+      version: HUMAN_GEM_VERSION,
+    });
+
+    expect(Object.keys(data)).toContain("Human-GEM");
+
+    const { metabolite } = data["Human-GEM"];
+    expect(metabolite.length).toBe(50);
```
nanjiangshu, line 19: I don't quite understand why this test is failed. The limit is set to `50` in the `search` function. In addition, would it be better to change `toBe(50)` to `toBeLessThanOrEqual(50)`?

api/test/integratedModels.test.js
```diff
@@ -0,0 +1,81 @@
+import fetch from "node-fetch";
+
+// This constant should be updated whenever new data (new versions
+// of existing models, or new models) is added to the project.
+const CORRECT_DATA = {
```
nanjiangshu, line 5: You are right. Files such as `reactions.tsv` does not exists in the `api` container. Sorry that I was not thinking well. I think it is fair enough to verify the data like that.

nanjiangshu (commented):
Hi @e0, the news tests you have added looks great! I have just a small comment regarding the testing for search result and otherwise I'm ready to approve this PR.

api/test/search.test.js
e0: Sorry, I should have pointed this out in the PR description. This is “supposed to” be failing right now. We have an actual bug in search at the moment. For metabolites results, we combine metabolites and compartmentalized metabolites before returning the results, where each node type has the limit of 50.

nanjiangshu: I see. Thanks for the explanation! However, I think still `toBeLessThanOrEqual` might be better to interpret the expectation.

e0: Ah yes, I see your point. That would match better with the meaning of the test. The downside is that 0 results would also pass. I specifically wanted to test this on something that would push the limit to enforce we get 50 back, but maybe I can rewrite the description better or add a comment.

nanjiangshu: Aha. I see what you mean. Metabolite for `H2O` has more than maximum hits. Then we should keep it but make a better description for this test.

e0: Added [a commit here](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/830/commits/4b1671d0bb7c53e58ab4fd060e10c87f2e05dd84) to address it. Still sounds a bit awkward I think but do you think it's ok?

nanjiangshu: How about "model search to have 50 results per component type if it exceeds the limit 50"?

e0: That sounds much better. Added another fix for it now. Thank you!

nanjiangshu (approved):
Looks great to me now!

e0:
Hi @MalinAhlberg, I sent you a new request to review the PR since quite a few more tests have been added. Please don't stress about reviewing it though! It would be nice to have input from @inghylt as well when she is back to the project next week since this is something she did a lot of research work on testing.

api/test/metabolites.test.js
```diff
@@ -0,0 +1,21 @@
+import fetch from "node-fetch";
+
+describe("metabolites", () => {
+  test("a metabolite should have a list of subsystems", async () => {
+    const res = await fetch(
+      `${API_BASE}/metabolites/MAM01199m?model=HumanGem&version=${HUMAN_GEM_VERSION}`
+    );
+
+    const { subsystems } = await res.json();
+    expect(subsystems.length).toBeGreaterThan(0);
```
MalinAhlberg, line 10: Check that all sybsystems have `id` and `name`? Is there any specific subsystem for `MAM01199m` that must be present?

MalinAhlberg (commented):
I (still) think this is great, but following on Nanjiang's previous comment, I wonder if it would make sense to make the test more specific and detailed. I suppose this would make it easier for us to catch errors, but of course it's a lot of work.
I added some separate comments with more concrete examples.

e0:
> I (still) think this is great, but following on Nanjiang's previous comment, I wonder if it would make sense to make the test more specific and detailed. I suppose this would make it easier for us to catch errors, but of course it's a lot of work.
> I added some separate comments with more concrete examples.

This is a great point @MalinAhlberg and I should have thought more about it prior to implementing this issue. I really like the specific suggestions/questions you brought up as well.

 It was a bit unclear exactly what the scope of this issue should be. Looking at the state now, we have a series of tests that act as a sanity check to make sure that most of the API endpoints are functional, but not necessarily 100% correct. I think adding more tests to validate the details would be a big improvement but it could take quite sometime to complete that goal. Do you think it would make sense to create a new issue that focuses more on unit testing for the `api`? Then we can write tests to focus more on the `neo4j` queries, similar to the test for search, as opposed to the `http` ones. It would also help us wrap up this current issue.

api/test/crossReference.test.js
```diff
@@ -0,0 +1,19 @@
...
+    expect(components.length).toBeGreaterThan(1);
+  });
+
+  test("should return a TSV file for a given data source", async () => {
+    const res = await fetch(`${API_BASE}/external-db/Recon3D/RE3335C`);
+
+    const {
+      externalDb: { url },
+    } = await res.json();
+    expect(url.length).toBeGreaterThan(1);
```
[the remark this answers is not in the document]
e0: I think it would be possible to make an http request to the url and see if the response is successful. I'm not sure if it would be too beneficial though because it only validates one of many possible URLs we have in the database. The intention here was to make sure that an external reference (to Recon3D as opposed to one of our own cross references like HMR) could have an URL.

api/test/hpa.test.js
```diff
@@ -0,0 +1,21 @@
+import fetch from "node-fetch";
+
+describe("HPA", () => {
+  test("should return a list of genes", async () => {
+    const res = await fetch(`${API_BASE}/hpa/genes`);
+
+    const genes = await res.json();
+    expect(genes.length).toBeGreaterThan(0);
+    expect(genes[0].length).toBe(3);
+    expect(genes[genes.length - 1].length).toBe(3);
```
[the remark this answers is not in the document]
e0: I'm not sure about this one. Do you have any input @mihai-sysbio?

api/test/metabolites.test.js
e0: I think a unit test might be suitable as mentioned in [this comment](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/830#issuecomment-1019984656).

MalinAhlberg:
> This is a great point @MalinAhlberg and I should have thought more about it prior to implementing this issue. I really like the specific suggestions/questions you brought up as well.
> 
> It was a bit unclear exactly what the scope of this issue should be. Looking at the state now, we have a series of tests that act as a sanity check to make sure that most of the API endpoints are functional, but not necessarily 100% correct. I think adding more tests to validate the details would be a big improvement but it could take quite sometime to complete that goal. Do you think it would make sense to create a new issue that focuses more on unit testing for the `api`? Then we can write tests to focus more on the `neo4j` queries, similar to the test for search, as opposed to the `http` ones. It would also help us wrap up this current issue.

Yes, that might be the best way, to open new issues for that and let this PR focus on what's already here :+1: .

e0:
Had a meeting with @inghylt, who brought up some good points about this PR. Similarly to what @MalinAhlberg [mentioned before], it could also be good to validate the counts, for example the number of genes and metabolites returned for Cytosol.

More specifically, @inghylt mentioned the following points that made me change my mind and work a bit more on this PR.

- There is value in validating details for single components since it still helps ensure that we didn't break anything.
- We do not necessarily need unit testing for validating this kind of data.

e0:
Rebased from the dev branch now and force pushed. If you want to test this more, it's probably easiest to replace your local branch with the remote one again. Sorry for the inconvenience.

e0:
Hi all, sorry for the re-requests again! I have added some new tests now for more precise validation of the components. Please have another review at this when you have time.

api/test/crossReference.test.js
```diff
@@ -0,0 +1,19 @@
...
+
+describe("cross reference", () => {
+  test("should return more than one component", async () => {
+    const res = await fetch(`${API_BASE}/external-db/HMR%202.0/HMR_1157`);
+
+    const { components } = await res.json();
+    expect(components.length).toBeGreaterThan(1);
+  });
+
+  test("should return a TSV file for a given data source", async () => {
```
nanjiangshu, line 11:
```suggestion
  test("should return a JSON object for a given data source", async () => {
```

MalinAhlberg:
> Looks great with newly added test @e0. I just added a minor comment for the test description. However, it seems linting is not working for the folder `api/test` since only `api/src` was configured, am I correct @MalinAhlberg ? Maybe we can create a new issue to include `api/test` to linting as well.

Good point! If it's not causing too much problems, we could add that already in this PR? Hopefully it would just mean adding the folder to the configs. Adding it as a separate issue sounds too.

api/test/crossReference.test.js
e0: Thanks for spotting this! The description was actually pasted from another file (data overlay) and I forgot to change the description. I changed it to a more appropriate one here: https://github.com/MetabolicAtlas/MetabolicAtlas/pull/830/commits/fdafc41c6e54b1379ff62da43a20264cc3b869f2

e0:
Good point about the linting! It took some time to figure out how to integrate jest with eslint properly but it seems to be working now. It would be great if you can have another run at this before we merge this PR. There is a new package added now as well so I think you would need to run `build-stack` and `start-stack` again.

MalinAhlberg (approved):
I really like the new tests! It's nice with some negative test cases!

api/test/compartments.test.js
```diff
@@ -0,0 +1,33 @@
...
+    expect(info.subsystemCount).toBe(subsystems.length);
+  });
+
+  test('a compartment should have related reactions', async () => {
+    const res = await fetch(
+      `${API_BASE}/compartments/nucleus/related-reactions?model=HumanGem&version=${HUMAN_GEM_VERSION}`
+    );
+
+    const data = await res.json();
+    expect(data.length).toBe(238);
```
inghylt, line 31: maybe use `NUCLEUS_INFO.reactionsCount` instead of 238?

e0: Good catch! Fixed in: https://github.com/MetabolicAtlas/MetabolicAtlas/pull/830/commits/67eec460463e3d38bda2ae10cff7098a5220397a

inghylt:
Great work with this PR! I'm so excited to see that we finally have some testing in place. Really nice feedback from the other reviewers as well :+1:

inghylt (approved):

nanjiangshu (approved):
Looks great now!

PR #837, opened 2022-01-14
fix: make consistent url when switching between 2D/3D Map Viewer for SVG maps with sub maps 

### LINKED ISSUE #745: Inconsistent and sometimes nonfunctional url for Cytosol

nanjiangshu:
This PR closes #745 

The URL in the address bar is reset to reflect the actual map shows in the Map Viewer, so that a hard refresh (ctrl-r) will not show `map not found` error
**Type of change**  

- [ ] Other
 
**List of changes made**  

**Screenshot of the fix**  

**Testing**  

- Instructions on how to test
Follow the steps described in the issue  #745, the buggy behavior will disappear. 
Moreover, when switching to the 2D map viewer, select e.g. `Part 4`, and then switch to 3D and then switch back to 2D, the original 2D map for `Part 4` will be restored.  

**Further comments**  

When switching to the 2D map for the first time from the 3D map viewer, the SVG map `Part 2`, which is the first element of the array `this.currentMap.svgs`,  is selected. This might be discussed if it is reasonable. 

**Checklist**  

### DIFF
--- a/frontend/src/components/explorer/MapViewer.vue
+++ b/frontend/src/components/explorer/MapViewer.vue
@@ -285,7 +285,7 @@ export default {
         let urlPath = `${this.$route.path}`;
         if (this.currentMap.mapReactionIdSet.length > 1) {
           if (newQuery.dim === '2d') {
-            urlPath = urlPath.replace(new RegExp(this.currentMap.id), `${this.currentMap.id}_1`);
+            urlPath = urlPath.replace(new RegExp(this.currentMap.id), this.currentMap.svgs[0].id);
           } else {
             urlPath = urlPath.replace(new RegExp(this.$route.params.map_id), this.currentMap.id);
           }
--- a/frontend/src/components/explorer/MapViewer.vue
+++ b/frontend/src/components/explorer/MapViewer.vue
@@ -279,11 +279,17 @@ export default {
         .join('&');
 
       if (newQuery.dim === this.$route.query.dim || (newQuery.dim && !this.$route.query.dim)) {
+        // if Map viewer (2D or 3D) is not changed, keep the url path
         const payload = [{}, null, `${this.$route.path}?${queryString}`];
         history.replaceState(...payload); // eslint-disable-line no-restricted-globals
       } else {
         let urlPath = `${this.$route.path}`;
         if (this.currentMap.mapReactionIdSet.length > 1) {
+          // if Map Viewer (2D or 3D) is changed, for the 2D Compartment maps with
+          // sub maps, e.g. Cytosol, replace the map id to the name of submap when
+          // switching to the 2D Map viewer, and replace the map id of the
+          // submap back to the compartment name when switching to the 3D Map
+          // viewer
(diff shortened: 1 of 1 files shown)


frontend/src/components/explorer/MapViewer.vue
```diff
@@ -277,10 +277,20 @@ export default {
       const queryString = Object.entries(newQuery)
         .map(e => e.join('='))
         .join('&');
-      const payload = [{}, null, `${this.$route.path}?${queryString}`];
+
       if (newQuery.dim === this.$route.query.dim || (newQuery.dim && !this.$route.query.dim)) {
+        const payload = [{}, null, `${this.$route.path}?${queryString}`];
         history.replaceState(...payload); // eslint-disable-line no-restricted-globals
       } else {
+        let urlPath = `${this.$route.path}`;
```
MalinAhlberg, line 285: To avoid future confusion, I think a comment about what's going on would be nice here

MalinAhlberg (approved):
Seems to work very well :+1: !

I left a comment about comments. Apart from that, I approve!

frontend/src/components/explorer/MapViewer.vue
nanjiangshu: Thanks for the comment! I've added some explanation to the code in the commit 3722c94

However, I'm not quite sure to which extent the code should be commented. If you go through the code for MetAtlas, there are very little comment. It's great that you raised this question and I think we should discuss about this: whether and how should we add comment in the code.

MalinAhlberg: Thanks! And yes, I have noticed the lack of comments :). I think at least adding very short comments would help in many places, the code base is quite massive and it's sometimes hard to figure out what's going on. But there are of course drawbacks of comments, such as the risk of them not being updated when the code is, etc. So es, a discussion would be good!

e0 (approved):

PR #845, opened 2022-01-19
fix: display data overlay level on mouse hover

e0:
This PR closes https://github.com/MetabolicAtlas/private-issues/issues/141.

**Type of change**  

**List of changes made**  

The data overlay level used to be displayed on mouse hover but was then broken. This PR re-enables it.

**Screenshot of the fix**  

image

**Testing**  

- Instructions on how to test

1. Visit `/explore/Human-GEM/map-viewer/acylglycerides_metabolism?dim=2d&panel=1`.
2. Select a data set, for example 'adipose tissue'.
3. Mouse over genes (rectangles) on the SVG map and verify that a number is displayed.

**Further comments**  

The mouse event initialization was previously done in the `mounted` function. In this function the `componentClassName` may still not be initialized so the mouse events are not setup.

The fix moves the mouse event initialization into a separate function which is called inside a watch handler for `componentClassName` to ensure it's done correctly.

**Checklist**  

### DIFF
--- a/frontend/src/components/explorer/mapViewer/Svgmap.vue
+++ b/frontend/src/components/explorer/mapViewer/Svgmap.vue
@@ -116,6 +116,9 @@ export default {
     async mapData() {
       await this.init();
     },
+    componentClassName() {
+      this.setupHoverEventHandlers();
+    },
     dataSet() {
       this.applyLevelsOnMap();
     },
@@ -133,23 +136,6 @@ export default {
       $('#svg-wrapper').on('click', aClass, async function f() {
         await self.selectElement($(this));
       });
-    });
-    $('#svg-wrapper').on('mouseover', `.${self.componentClassName}`, function f(e) {
-      const id = $(this).attr('class').split(' ')[1].trim();
-      if (id in self.computedLevels) {
-        self.$refs.tooltip.innerHTML = self.computedLevels[id][1]; // eslint-disable-line prefer-destructuring
-      } else if (Object.keys(self.computedLevels).length !== 0) {
-        self.$refs.tooltip.innerHTML = self.computedLevels['n/a'][1]; // eslint-disable-line prefer-destructuring
-      } else {
-        return;
-      }
-      self.$refs.tooltip.style.top = `${e.pageY - $('.svgbox').first().offset().top + 15}px`;
-      self.$refs.tooltip.style.left = `${e.pageX - $('.svgbox').first().offset().left + 15}px`;
-      self.$refs.tooltip.style.display = 'block';
-    });
-    $('#svg-wrapper').on('mouseout', `.${self.componentClassName}`, () => {
-      self.$refs.tooltip.innerHTML = '';
-      self.$refs.tooltip.style.display = 'none';
     });
     $('.svgbox').on(
       'webkitfullscreenchange mozfullscreenchange fullscreenchange mozFullScreen MSFullscreenChange',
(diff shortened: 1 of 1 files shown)


nanjiangshu (approved):
Great fix @e0, although I do not fully understand the logic of the code - need to learn more about vuejs :stuck_out_tongue_winking_eye:

Is that you redefined `componentClassName` by overriding the function from dataOverlay?

e0:
> Great fix @e0, although I do not fully understand the logic of the code - need to learn more about vuejs 😜
> 
> Is that you redefined `componentClassName` by overriding the function from dataOverlay?

This is a great question!

To start, we can see that the `componentClassName` is a store (Vuex) binding as seen in the screenshot below. In this case, the store binding is associated with an async process (fetching data from the API), so it could take a bit of time for this value to be set after the component is mounted. 

image

In the next screenshot, we can see that we intend to use `componentClassName` to setup the mouse-over event listener in the `mounted` Vue lifecycle hook.

image

The problem is that the `componentClassName` binding may not be set at this point. During debugging, this was indeed the case and `componentClassName` appeared to be `null`. This means that the mouse-over event listener was not correctly setup.

To fix this, we could make sure to wait for the `componentClassName` to be set before setting up the mouse-over event listeners. In Vue, for any variable defined in the `data() {`, or `computed: {` sections, we could monitor their changes in the `watch: {` section. As you see in the screenshot below, we are now watching the `componentClassName` variable and whenever it is changed, we try to setup the mouse-over event listeners.

image

inghylt (approved):
Nice work! Not only regarding finding and fixing the bug but also making the code more readable :clap:

nanjiangshu:
Thanks @e0 for the clear explanation!

mihai-sysbio:
Because of my skepticism of adding unnecessary overhead of watching variables, I looked into this a bit more.

Unlike with API calls, the Vuex store is used in this case purely client-side, and it doesn't seem to need to be asynchronous:
https://github.com/MetabolicAtlas/MetabolicAtlas/blob/bbbd7d02059a1e927f5bd5674c3300ad72574e02/frontend/src/store/modules/dataOverlay.js#L46

Also, unlike `componentType` and `componentDefaultColor`, `componentClassName` is used only in `Svgmap.vue` and not shared with  `ThreeDviewer.vue`.

e0:
> Because of my skepticism of adding unnecessary overhead of watching variables, I looked into this a bit more.
> 
> Unlike with API calls, the Vuex store is used in this case purely client-side, and it doesn't seem to need to be asynchronous:
> 
> https://github.com/MetabolicAtlas/MetabolicAtlas/blob/bbbd7d02059a1e927f5bd5674c3300ad72574e02/frontend/src/store/modules/dataOverlay.js#L46
> 
> Also, unlike `componentType` and `componentDefaultColor`, `componentClassName` is used only in `Svgmap.vue` and not shared with `ThreeDviewer.vue`.

Interesting point. Can you elaborate on what you mean by overhead? Seems reasonable to me to bind the event handler as soon as the `componentClassName` is modified. This should also play well with when we want to support multiple overlays at the same time.

e0:
> > Because of my skepticism of adding unnecessary overhead of watching variables, I looked into this a bit more.
> > Unlike with API calls, the Vuex store is used in this case purely client-side, and it doesn't seem to need to be asynchronous:
> > https://github.com/MetabolicAtlas/MetabolicAtlas/blob/bbbd7d02059a1e927f5bd5674c3300ad72574e02/frontend/src/store/modules/dataOverlay.js#L46
> > 
> > Also, unlike `componentType` and `componentDefaultColor`, `componentClassName` is used only in `Svgmap.vue` and not shared with `ThreeDviewer.vue`.
> 
> Interesting point. Can you elaborate on what you mean by overhead? Seems reasonable to me to bind the event handler as soon as the `componentClassName` is modified. This should also play well with when we want to support multiple overlays at the same time.

Update: We had a short discussion during standup and @mihai-sysbio clarified on two interesting points. Here is a summary of the points and conclusion for now.

1. Does watching too many things slow down the website? Is there a better way than using `watch` for these situations? It might be worth to look into this a bit but no need to create an issue for now.
2. Sometimes we have things in the store (Vuex) that are only used in one component. For these situations, should this be the default or should we only try to put things in the store if they are shared in multiple components? For this specific case, the `componentClassName` could be used in multiple places (map viewer and interaction partners), so no changes are made for now.

It would be interesting to look into both of these a bit more. For now we have concluded that this PR can be merged in its current state.

If I missed anything in this summary, please feel free to add a reply as clarification.

PR #875, opened 2022-02-09
fix: run GH Actions workflows on all branches except main

mihai-sysbio:
This PR changes workflows to run on all branches except `main`, so that bugs don't squeeze through PRs between branches which are not `develop`. Here is the relevant [documentation](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-branches-and-tags).

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -2,9 +2,11 @@ name: "CodeQL"
 
 on:
   push:
-    branches: [ develop ]
+    branches-ignore: 
+      - 'main'
   pull_request:
-    branches: [ develop ]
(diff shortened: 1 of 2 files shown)


e0 (approved):
Looks good!

As a note, we had a group discussion and decided that it was good to have this safety without it necessarily dictating the workflow.

Minor detail, it looks like [it's ok to omit the single quotes around main](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-branches) but of course either way is fine.

nanjiangshu (approved):
Looks good!

MalinAhlberg (approved):

Issue #856, opened 2022-01-28
Sanitize user input


inghylt:
**Please describe the feature**

As a developer, I want to make sure that we sanitize our input so that I know the codebase is safe from malicious user input

**Describe the solution you would like**

- A summary of where sanitation is needed
- Fixes for these files. If fixing these are too complicated, create a new issue for fixing

**Additional context**

This is obv. good practice in general, but the trigger to create the issue was the [conversation](https://github.com/chalk/ansi-regex/commit/8d1d7cdb586269882c4bdc1b7325d0c58c8f76f9) related to CVE-2021-3807, an ansi-regex vulnerability

Issue #686, opened 2021-09-09
search support for composite(complex) metabolite name or aliases


mihai-sysbio:
_[Originally reported by @ANiknejad in Human-GEM](https://github.com/SysBioChalmers/Human-GEM/issues/319)_
> hello!
> 
> for example:
> MAM02911c sn-glycero-3-phospho-1-inositol
> 
> I can get 'MAM02911c' from
> https://metabolicatlas.org/explore
> or from
> https://metabolicatlas.org/search
> 
> but not 'sn-glycero-3-phospho-1-inositol', neither from
> https://metabolicatlas.org/explore
> or from
> https://metabolicatlas.org/search
> 
> BUT WITH simple name like 'itaconate' both search engine work fine

PR #880, opened 2022-02-16
fix: clean search term and escape bad characters for neo4j

### LINKED ISSUE #764: Quick search throws error when using charachter :
**Describe the bug**

Using quick search and typing in `:` leads to an error

**Steps to reproduce**

Type for example `gene:` in quick search, note the error being reported

**Expected behavior**

That no error is reported but that the quick search works as normally

**Screenshots**

nanjiangshu:
This PR closes #764 and (probably) #884

**Type of change**  

**List of changes made**  

The search term for the quick search is cleaned to avoid unexpected neo4j querying error.
1. Special characters are escaped by adding `\ `
2. The character `"` and `\` are removed since it may cause errors even after escaping- 
3. Multiple white spaces are replaced with a single space
4. the trailing white spaces are removed 

Besides, a test is added for searching of terms with special characters.

**Screenshot of the fix**  

**Testing**  

- Instructions on how to test
Test in the quick search with searching terms e.g.  
1. `Node: rna`, 
2. `Tricarboxylic acid cycle and glyoxylate/dicarboxylate ` 
3. `°!#€€#€%&&///(/(())())=?``^ÅPOIUUYYTTRRE EWW_:;M:*ÄÖ*^ÄÖLKK;:;M_:;>ZXZXCVV°!"©@££$∞§ ∞§|[]]≈± ø–‘›‹ç≈÷≈çç‹›‘›‘ ‘’ ’ –ßß∂ƒƒ¸ ƒ¸˛˛√√ªﬁøæıüµ•Ωé®†µµüıı©@£$∞§§§||[ []] []≈±´`  😄  

**Further comments**  

**Checklist**  

### DIFF
--- a/api/src/neo4j/queries/search.js
+++ b/api/src/neo4j/queries/search.js
@@ -328,9 +328,7 @@ const _search = async ({
   let term = sanitizeSearchString(searchTerm, true);
   // the EC field for reaction could contain ":", which is a special character
   // in this case th search term is modified to be escape and perform an exact match
-  term = term.includes('EC\\:')
-    ? `\\"${term}~\\"`
-    : `${term}~`;
+  term = term.includes('EC\\:') ? `\\"${term}~\\"` : `${term}~`;
 
   // Metabolites are not included as it would mess with the limit and
   // relevant metabolites should be matched through CompartmentalizedMetabolites
--- a/api/src/neo4j/queries/search.js
+++ b/api/src/neo4j/queries/search.js
@@ -326,9 +326,17 @@ const _search = async ({
 
   // the EC field for reaction could contain ":", which is a special character
   // in this case th search term is modified to be escape and perform an exact match
-  const term = searchTerm.includes('EC:')
+  let term = searchTerm.includes('EC:')
     ? `\\"${searchTerm}~\\"`
     : `${searchTerm}~`;
+
+  // clean the search term and excape special characters that may cause errors
+  // in neo4j querying
+  term = term
+    .replace(/\s\s+/g, ' ')
+    .replace(/^\s|\s$/g, '')
+    .replace(/([\"])/g, '')
+    .replace(/([:/^!\"\(\)\[\]])/g, '\\$1');
 
   // Metabolites are not included as it would mess with the limit and
   // relevant metabolites should be matched through CompartmentalizedMetabolites
--- a/api/src/neo4j/queries/search.js
+++ b/api/src/neo4j/queries/search.js
@@ -1,4 +1,5 @@
 import queryListResult from 'neo4j/queryHandlers/list';
+import { sanitizeSearchString } from 'utils/utils';
 const INTEGRATED_MODELS = require('data/integratedModels');
 
 const componentTypes = [
@@ -324,11 +325,12 @@ const _search = async ({
 }) => {
   const v = version ? `:V${version}` : '';
 
+  let term = sanitizeSearchString(searchTerm, true);
   // the EC field for reaction could contain ":", which is a special character
   // in this case th search term is modified to be escape and perform an exact match
-  let term = searchTerm.includes('EC\\:')
-    ? `\\"${searchTerm}~\\"`
-    : `${searchTerm}~`;
+  term = term.includes('EC\\:')
+    ? `\\"${term}~\\"`
+    : `${term}~`;
 
   // Metabolites are not included as it would mess with the limit and
   // relevant metabolites should be matched through CompartmentalizedMetabolites
--- a/api/src/neo4j/queries/search.js
+++ b/api/src/neo4j/queries/search.js
@@ -330,7 +330,6 @@ const _search = async ({
     ? `\\"${searchTerm}~\\"`
     : `${searchTerm}~`;
 
-  console.log(`term: ${term}`); //eslint-disable-line
   // Metabolites are not included as it would mess with the limit and
   // relevant metabolites should be matched through CompartmentalizedMetabolites
   let statement = `
--- a/api/src/neo4j/queries/search.js
+++ b/api/src/neo4j/queries/search.js
@@ -325,15 +325,12 @@ const _search = async ({
 }) => {
   const v = version ? `:V${version}` : '';
 
-  let term = sanitizeSearchString(searchTerm, true);
-  // the EC field for reaction could contain ":", which is a special character
-  // in this case th search term is modified to be escape and perform an exact match
-  term = term.includes('EC\\:') ? `\\"${term}~\\"` : `${term}~`;
+  const term = sanitizeSearchString(searchTerm, true);
 
   // Metabolites are not included as it would mess with the limit and
   // relevant metabolites should be matched through CompartmentalizedMetabolites
   let statement = `
-CALL db.index.fulltext.queryNodes("fulltext", "${term}")
+CALL db.index.fulltext.queryNodes("fulltext", "${term} ${term}~")
 YIELD node, score
 WITH node, score, LABELS(node) as labelList
 OPTIONAL MATCH (node)-[${v}]-(parentNode:${model})
--- a/api/src/neo4j/queries/search.js
+++ b/api/src/neo4j/queries/search.js
@@ -335,7 +335,7 @@ const _search = async ({
   term = term
     .replace(/\s\s+/g, ' ')
     .replace(/^\s|\s$/g, '')
-    .replace(/([\"])/g, '')
+    .replace(/([\"\\])/g, '')
     .replace(/([:/^!\"\(\)\[\]])/g, '\\$1');
 
   // Metabolites are not included as it would mess with the limit and
--- a/api/src/neo4j/queries/search.js
+++ b/api/src/neo4j/queries/search.js
@@ -326,18 +326,11 @@ const _search = async ({
 
   // the EC field for reaction could contain ":", which is a special character
   // in this case th search term is modified to be escape and perform an exact match
-  let term = searchTerm.includes('EC:')
+  let term = searchTerm.includes('EC\\:')
     ? `\\"${searchTerm}~\\"`
     : `${searchTerm}~`;
 
-  // clean the search term and excape special characters that may cause errors
-  // in neo4j querying
-  term = term
-    .replace(/\s\s+/g, ' ')
-    .replace(/^\s|\s$/g, '')
-    .replace(/([\"\\])/g, '')
-    .replace(/([:/^!\"\(\)\[\]])/g, '\\$1');
-
+  console.log(`term: ${term}`); //eslint-disable-line
   // Metabolites are not included as it would mess with the limit and
   // relevant metabolites should be matched through CompartmentalizedMetabolites
   let statement = `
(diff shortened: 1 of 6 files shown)


mihai-sysbio:
It sounds like this would also partly resolve #856 - if so, maybe we should refine that issue.

nanjiangshu:
Thanks @mihai-sysbio for pointing out the problem with backslash in the search term. A fix has been added in the commit 17c7f4e8

e0:
> Thanks @mihai-sysbio for pointing out the problem with backslash in the search term. A fix has been added in the commit [17c7f4e](https://github.com/MetabolicAtlas/MetabolicAtlas/commit/17c7f4e89e10d820bda181d943ff5b7f21540f9e)

Works well from what I tested but the backslash escaping problem still seems unsolved.

nanjiangshu:
> > Thanks @mihai-sysbio for pointing out the problem with backslash in the search term. A fix has been added in the commit [17c7f4e](https://github.com/MetabolicAtlas/MetabolicAtlas/commit/17c7f4e89e10d820bda181d943ff5b7f21540f9e)
> 
> Works well from what I tested but the backslash escaping problem still seems unsolved.

Hi @e0 Thanks for testing. Could you please share the search term you used? I tested various terms with backslashes, such as `\\ askd\( \[\\// \\'"\"2\\\'.*///\/\/`, `!\/\(\\//`, but the error message is not prompted.

e0:
> > > Thanks @mihai-sysbio for pointing out the problem with backslash in the search term. A fix has been added in the commit [17c7f4e](https://github.com/MetabolicAtlas/MetabolicAtlas/commit/17c7f4e89e10d820bda181d943ff5b7f21540f9e)
> > 
> > 
> > Works well from what I tested but the backslash escaping problem still seems unsolved.
> 
> Hi @e0 Thanks for testing. Could you please share the search term you used? I tested various terms with backslashes, such as `\\ askd\( \[\\// \\'"\"2\\\'.*///\/\/`, `!\/\(\\//`, but the error message is not prompted.

The searches worked fine actually. It's just about the CodeQL thing: https://github.com/MetabolicAtlas/MetabolicAtlas/security/code-scanning/28?query=ref%3Arefs%2Fpull%2F880%2Fmerge+ref%3Arefs%2Fpull%2F880%2Fhead+ref%3Arefs%2Fheads%2Ffix%2Fquick-search-error-chars

nanjiangshu:
```javascript
  term = term
    .replace(/\s\s+/g, ' ')
    .replace(/^\s|\s$/g, '')
    .replace(/([\"\\])/g, '')
    .replace(/([:/^!\"\(\)\[\]])/g, '\\$1');
```
> This does not escape backslash characters in the input. CodeQL

I don't quite understand the warning made by CodeQL. Firstly, it does not make any warning previously when there is no sanitizing at all.  Secondly, the third `replace()` in the code removes all backslashes from the original search term, and then the fourth search term adds backslashes if we still want to parse special characters such as `(`, `'` to the neo4j query. CodeQL says `This does not escape backslash characters in the input`, is this because of the addition of backslashes in the fouth replace(). I'm not sure of removal of all the special characters will undo this CodeQL warning. Some characters like `!` might be removed, but we probably need to keep characters such as `(` and `:`?

e0:
> I don't quite understand the warning made by CodeQL. Firstly, it does not make any warning previously when there is no sanitizing at all.  Secondly, the third replace() in the code removes all backslashes from the original search term, and then the fourth search term adds backslashes if we still want to parse special characters such as (, ' to the neo4j query. CodeQL says This does not escape backslash characters in the input, is this because of the addition of backslashes in the fouth replace(). I'm not sure of removal of all the special characters will undo this CodeQL warning. Some characters like ! might be removed, but we probably need to keep characters such as ( and :?

I tested this a bit more and see that the search breaks if you type in `\\\`. Trying this in the browser JavaScript console confirms the error. In the linked page from my previous comment, they have some good advice if you click "Show more".

image

nanjiangshu:
Hi @e0 , thanks for pointing out that the search string `\\\` caused the error message. This is because backslashes are removed and thus results in an empty search string, but the minimal length checking is still passed since it is checking the length of the original input string. I have moved the sanitizing function from the `api` to the `frontend` and the minimal length checking is for the sanitized string. In addition, the CodeQL warning is also solved with a more complete regexp. 

There are still room for improvement. For example, the maximum length of input is not restricted. But I think this is a more general problem but limited to the GemSearch, so it is better to be solved in https://app.zenhub.com/workspaces/metatlas-sprint-607745622d49260019ce115a/issues/metabolicatlas/metabolicatlas/856.

api/src/neo4j/queries/search.js
```diff
@@ -326,7 +326,7 @@ const _search = async ({
 
   // the EC field for reaction could contain ":", which is a special character
   // in this case th search term is modified to be escape and perform an exact match
```
inghylt, line 328:
```suggestion
  // in this case the search term is modified to be escaped and perform an exact match
```

inghylt (commented):
I think the changes look good! Two things though:

- The highlighting of matching text in result seems to be affected. Compare searching for `MAR10368` on localhost and metabolicatlas.org. It's not completely broken though, for example when searching for `Glycosylphosphatidylinositol (GPI)-anchor` you get highlighted text as expected

- Seems like you get some kind of result for all search terms except variants of just using `\\`, then it just keeps on loading (see image below). Not sure if this is something we should care about though since users probably won't be affected negatively by this.
image

nanjiangshu:
Thanks for reviewing @inghylt 
> I think the changes look good! Two things though:
> 
>     * The highlighting of matching text in result seems to be affected. Compare searching for `MAR10368` on localhost and metabolicatlas.org. It's not completely broken though, for example when searching for `Glycosylphosphatidylinositol (GPI)-anchor` you get highlighted text as expected
> 
Interesting. I do see highlight when searching for the key word `MAR10368` as seen in the screenshot below.
Screenshot 2022-02-18 at 11 38 27


>     * Seems like you get some kind of result for all search terms except variants of just using `\\`, then it just keeps on loading (see image below). Not sure if this is something we should care about though since users probably won't be affected negatively by this.
This is made by purpose since I have removed the backslash before passing the search term to the API and moreover, the length of cleaned search term is used when validating for the minimal length (=2). In that case, the GemSearch is waiting for a valid input with at least two characters after input of `\\`. 

I have also tried by not removing `\`, in that case, the escaped search term will be passed to the search API and one actually got hit after typing in `\\` which I think is weird. 

Nevertheless, I don't think any real user will type in `\\` in the search bar. For people who want to break the website, removing `\` is enough. 

I do think we should not overdo things and spend much time on things users might not pay an attention.

e0:
@nanjiangshu Quick follow up. If we do move the sanitizing logic to the backend again and you want to add some tests, you can continue working in [this file](https://github.com/MetabolicAtlas/MetabolicAtlas/blob/develop/api/test/search.test.js). I also find the [Jest documentation](https://jestjs.io/docs/using-matchers) useful to have open when writing tests.

nanjiangshu:
> @nanjiangshu Quick follow up. If we do move the sanitizing logic to the backend again and you want to add some tests, you can continue working in [this file](https://github.com/MetabolicAtlas/MetabolicAtlas/blob/develop/api/test/search.test.js). I also find the [Jest documentation](https://jestjs.io/docs/using-matchers) useful to have open when writing tests.

Great idea to add a test for this API!
We can not simply move the sanitizing logic to the API, it is needed also in the frontend. 
Firstly, the [RegExp of searchTerm](https://github.com/MetabolicAtlas/MetabolicAtlas/blob/81f209a2d7431c65b171cfb8c49a44f930c4f073/frontend/src/components/explorer/gemBrowser/GemSearch.vue#L245) will report an error by vuejs with `invalid syntax` if some special letters (e.g. `(`)are not escaped. This error was not reported by vuejs previously because the `unknowError` was triggered.
Secondly, the [minimal length checking](https://github.com/MetabolicAtlas/MetabolicAtlas/blob/81f209a2d7431c65b171cfb8c49a44f930c4f073/frontend/src/components/explorer/gemBrowser/GemSearch.vue#L207) needs to check the length of the filtered string.

A good solution would be if this function can be used at both the `api` and the `frontend` and a quick and dirty solution would be we define the function at both places.

inghylt:
We should aim to solve #686 and #884 with this issue as well

nanjiangshu:
I think the issue #686 will be solved by the PR #886

nanjiangshu:
I checked with the term @inghylt mentioned in the issue #884 and it is highlighted. 
Could you please verify that the issue #884 is actually solved by this PR?

api/src/neo4j/queries/search.js
nanjiangshu: Special treatment with `EC:` is not needed anymore since the symbol `:` is already sanitized. I have removed the both the code and this comment in the commit a9a85114

MalinAhlberg (approved):
Apart from the complaint from Prettier, I think this looks good! #884 seems to be solved by this too, as you mention.
Good job :+1: !

api/test/search.test.js
```diff
@@ -43,4 +43,23 @@ describe('search', () => {
...
+        model: 'HumanGem',
+        version: HUMAN_GEM_VERSION,
+      }),
+    ]);
+    for (const data of [data1, data2]) {
+      const { name } = data['Human-GEM'];
+      expect(name).toEqual('Human-GEM');
+    }
+  });
+});
```
inghylt, line 65: Add newline?

nanjiangshu: Not quite sure. It's not added in the rest of testing js files.

api/src/utils/utils.js
```diff
@@ -0,0 +1,10 @@
+export function sanitizeSearchString(term, isAddBackSlash = true) {
```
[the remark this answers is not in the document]
nanjiangshu: When checking for the minimal length of the input string, the backslash should not be added, otherwise, backslashes will also be counted.

api/test/search.test.js
inghylt: `ma-exec api yarn format` does react to this as well so I think it makes sense to add it

nanjiangshu: It seems there is a prettier issue as @MalinAhlberg also pointed out. I guess this is caused when I solving the merging conflict in the browser, which ignored the prettier.

api/src/utils/utils.js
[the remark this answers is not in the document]
inghylt: Maybe worth adding a comment about that in GemSearch.vue, but since no one else has asked about it might be superfluous. Up to you :)

inghylt (approved):
Really nice work!

api/src/utils/utils.js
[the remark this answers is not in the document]
nanjiangshu: Added a comment in c5b7f090

nanjiangshu:
Hi @e0, do you have any other comments for this issue?

e0:
> Hi @e0, do you have any other comments for this issue?

Hi, sorry I didn't mention but since it already got two approvals I didn't take another look. It worked well originally and I trust that your work along with the two approvals are good enough.

nanjiangshu:
Thinking about the actual amount of effort spent on this issue, I would bump it to at least 4.

PR #882, opened 2022-02-16
fix: add delay for blur event

### LINKED ISSUE #729: Window does not switch to global search after clicking "Search all integrated GEMs" on safari
When doing the gem searching, the webpage does not switch to global search after clicking "Search all integrated GEMs" 

**To reproduce**
    1. Open the webpage
    2. Click the GemSearch and input `ENSG00000002587` with model as Fruitfly-GEM
    3. At the popup window, click "Search all integrated GEMs"
    4. The popup window disappears but it does not switch to the global search page rather stay on the current page.

**Additional context**
This bug was found on Safari but neither Firefox nor Chrome.

Related issue: https://app.zenhub.com/workspaces/metatlas-sprint-607745622d49260019ce115a/issues/metabolicatlas/private-issues/143

e0:
This PR closes #729.

**Type of change**  

**List of changes made**  

- Add small delay (100ms) to the blur event for the gem search. Please note that this does not delay the global search request.
- Removed usages of jQuery in this file with plain JavaScript. Not needed for this issue but I saw some related error output in Safari.

**Screenshot of the fix**  

safari_search

**Testing**  

- Instructions on how to test
1. Click the GemSearch and input `ENSG00000002587` with model as `Fruitfly-GEM`.
2. Click the "Search all integrated GEMs" button.
3. Verify that the global page loads and performs a search on `ENSG00000002587` and the gem search hides.

**Further comments**  

Please let me know if you think the jQuery stuff should not be included here.

**Checklist**  

### DIFF
--- a/frontend/src/components/explorer/gemBrowser/GemSearch.vue
+++ b/frontend/src/components/explorer/gemBrowser/GemSearch.vue
@@ -115,7 +115,6 @@
 
 <script>
 import { mapGetters, mapState } from 'vuex';
-import $ from 'jquery';
 import { default as messages } from '@/content/messages';
 
 export default {
@@ -202,7 +201,7 @@ export default {
       }
     },
     async search() {
-      $('#search').focus();
+      document.getElementById('search').focus();
       if (this.searchTermString.length < 2) {
         return;
       }
@@ -277,9 +276,9 @@ export default {
         } else if (
           document.getElementById('model-select').contains(document.activeElement) === false
         ) {
-          $('#search').focus();
+          document.getElementById('search').focus();
         }
-      });
+      }, 100);
     },
   },
 };


mihai-sysbio:
> * Removed usages of jQuery in this file with plain JavaScript. Not needed for this issue but I saw some related error output in Safari.

Awesome, that's very much in line with https://github.com/MetabolicAtlas/private-issues/issues/28 .

frontend/src/components/explorer/gemBrowser/GemSearch.vue
```diff
@@ -277,9 +276,9 @@ export default {
         } else if (
           document.getElementById('model-select').contains(document.activeElement) === false
         ) {
-          $('#search').focus();
+          document.getElementById('search').focus();
         }
-      });
+      }, 100);
```
mihai-sysbio, line 281: Have you tested smaller values, eg 50 ms? I'm a bit wary of introducing lag in the UI.

mihai-sysbio (approved):
The changes make sense to me, but I haven't tested the difference. I'm hoping @nanjiangshu will, to confirm that this resolved the problem.

frontend/src/components/explorer/gemBrowser/GemSearch.vue
e0: I tried 10 initially but that did not work occasionally. I did actually try 50 too and it worked fine for me but thought 100 was safer and also is not in the way of the user since it's a non-blocking lag. We can reduce it to 50 if you want though. In general I agree with you on having as small a number as possible. With this specific component though, I think we can afford to have 100 here as it plays well with the hiding animation.

nanjiangshu (commented):
Hi @e0, this is really an interesting fix, especially with the increase of timeout. When I tested on my localhost. The behavior is still like before. Even when I changed this timeout value to a very larger number, such as 50000, it still did not work. Nevertheless, I don't see any UI lags despite of this large timeout setting. 

Interestingly, when I deployed this branch to the dev server, (with timeout set to 100), the windows switch works generally well, although I do encounter window non switching problem once in a while. 

I have also deployed this branch to http://134.158.151.84, when opened in private window on Safari, it didn't work neither. 

I guess there might be something difference with the `docker-compose-remote.yml` and `docker-compose-local.yml` or settings in nginx. 

As for the timeout value, since I don't see lags when changing from 100 to 500, I think we can set it to a larger value.

nanjiangshu (approved):
I updated my OS this morning and made another test. It seems it works well with window switching on all three servers, my local server, the dev server and http://134.158.151.84/.

It looks like the problem is with Safari on my computer.

PR #886, opened 2022-02-21
feat: improve search scoring

### LINKED ISSUE #707: Research on how to improve scores returned by neo4j
As a developer, I would like the score returned by `db.index.fulltext.queryNodes` in api/src/neo4j/queries/search.js to be dependable, so that I can use it to rank the results returned.

Currently, there are two sorting functions used related to sorting results returned from _search in api/src/neo4j/queries/search.js, `sortResultsScore(a, b)` and `sortResultsSearchTerm(a, b, this.searchedTerm)`. It would be optimal to just having to sort based on score. However, in some cases neo4j returns a higher score for nodes which should not be ranked as high. An example:

Gene with id `ENSG00000132664` is connected to GeneState with name `POLR3F`and Gene with id `ENSG00000181222` is connected to GeneState with name `POLR2A`

Thus, when searching on `POLR3F` it would be expected for `ENSG00000132664` to receive a higher score than `ENSG00000181222`, but this is not the case

This has the consequence that `sortResultsSearchTerm(a, b, this.searchedTerm)` has to be used as well as `sortResultsScore(a, b)` within the codebase. Just sorting based on `sortResultsSearchTerm(a, b, this.searchedTerm)` is not working either as this does not work well when sorting for matches such as external Ids (noticeable when using global search).

e0:
This PR closes #707 and #686.

**Type of change**  

- [x] Other
 
**List of changes made**  

Use a combination of exact match and fuzzy match when searching.

**Testing**  

- Instructions on how to test

Use the quick search and see if the results are sensible. Specifically the following search terms caused problems before. You can use the web inspector to check the scoring of the search results.

1. `POLR3F`: This should have an exact match on the gene name but it was ranked 2nd. It should be ranked 1st now.
2. `Asparaginyl-Cysteinyl`: Unsure exactly what the problem is but this does not return any matches when using fuzzy match. Using exact match does return sensible results and you should see some results returned with the top ranking results having partial matches with the metabolite name.

**Further comments**  

There is a new test case added that verifies the above two keywords. You can verify this locally with `ma-exec api yarn test` and the test named `model search should receive sensible ranking scores` should pass.

**Checklist**  

### DIFF
--- a/api/src/neo4j/queries/search.js
+++ b/api/src/neo4j/queries/search.js
@@ -332,6 +332,8 @@ const _search = async ({
 
   // Metabolites are not included as it would mess with the limit and
   // relevant metabolites should be matched through CompartmentalizedMetabolites
+  // The search term is used twice, once with exact match and once with
+  // fuzzy match. This seems to produce optimal results.
   let statement = `
 CALL db.index.fulltext.queryNodes("fulltext", "${term} ${term}~")
 YIELD node, score
--- a/api/src/neo4j/queries/search.js
+++ b/api/src/neo4j/queries/search.js
@@ -327,13 +327,13 @@ const _search = async ({
   // the EC field for reaction could contain ":", which is a special character
   // in this case th search term is modified to be escape and perform an exact match
   const term = searchTerm.includes('EC:')
-    ? `\\"${searchTerm}~\\"`
-    : `${searchTerm}~`;
+    ? `\\"${searchTerm}\\"`
+    : `${searchTerm}`;
 
   // Metabolites are not included as it would mess with the limit and
   // relevant metabolites should be matched through CompartmentalizedMetabolites
   let statement = `
-CALL db.index.fulltext.queryNodes("fulltext", "${term}")
+CALL db.index.fulltext.queryNodes("fulltext", "${term} ${term}~")
 YIELD node, score
 WITH node, score, LABELS(node) as labelList
 OPTIONAL MATCH (node)-[${v}]-(parentNode:${model})
--- a/api/test/search.test.js
+++ b/api/test/search.test.js
@@ -18,4 +18,29 @@ describe('search', () => {
     const { metabolite } = data['Human-GEM'];
     expect(metabolite.length).toBe(50);
   });
+
+  test('model search should receive sensible ranking scores', async () => {
+    const [data1, data2] = await Promise.all([
+      search({
+        searchTerm: 'POLR3F',
+        model: 'HumanGem',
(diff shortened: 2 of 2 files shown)


mihai-sysbio:
Really cool solution! Just to make sure, does this cover both _GemSearch_ and _GlobalSearch_?

e0:
> Really cool solution! Just to make sure, does this cover both _GemSearch_ and _GlobalSearch_?

It covers both. The change is in the function that’s shared by both.

nanjiangshu (approved):
Nice fix! 

I think this also solved the issue  #686
The only problem is that there will be merging conflict with the PR #880  😅

e0:
> Nice fix!
> 
> I think this also solved the issue #686 The only problem is that there will be merging conflict with the PR #880 😅

Haha thanks! Hope the merging is not too bad since it's just a couple of lines here but let me know if you want any help with merging it.

Great that you mentioned the other issue too. I can confirm that it appears to be solved.

api/src/neo4j/queries/search.js
```diff
@@ -327,13 +327,13 @@ const _search = async ({
...
-    ? `\\"${searchTerm}~\\"`
-    : `${searchTerm}~`;
+    ? `\\"${searchTerm}\\"`
+    : `${searchTerm}`;
 
   // Metabolites are not included as it would mess with the limit and
   // relevant metabolites should be matched through CompartmentalizedMetabolites
   let statement = `
-CALL db.index.fulltext.queryNodes("fulltext", "${term}")
+CALL db.index.fulltext.queryNodes("fulltext", "${term} ${term}~")
```
MalinAhlberg, line 336: A comment would be nice here, to avoid future confusion

MalinAhlberg (approved):
Cool fix!

As far as I can tell, this works well! I however noticed that searching for `ATP => cAMP + PPi` (one of the examples in the search field place holder) gives a different result using this branch. Not sure if it's irrelevant, better or worse.

api/src/neo4j/queries/search.js
e0: Great point @MalinAhlberg ! Added a comment about this now.

e0:
> Cool fix!
> 
> As far as I can tell, this works well! I however noticed that searching for `ATP => cAMP + PPi` (one of the examples in the search field place holder) gives a different result using this branch. Not sure if it's irrelevant, better or worse.

Good that you spotted this. I took a look at the new results for this one and I think they are ok. Different but not necessarily worse imo. I think this specific example might generate even more sensible results after @nanjiangshu's work involving special characters.

PR #887, opened 2022-02-22
Feat/quick search layout

inghylt:
Draft PR for private issue 151 by @inghylt and @MalinAhlberg 

**Type of change**
- The categories are sorted based on the highest score in their respective results list
- The amount of results per category are limited to 6 to make it easier to see the results from different categories and limit the amount of scrolling

**Testing**
Try for example `AMY1A` and `Pantothenate and CoA biosynthesis` for different models in quick search and global search

**Questions**
1. In the issue there is a [comment](https://github.com/MetabolicAtlas/private-issues/issues/151#issuecomment-1041407234)  _"The GemSearch results should be tabulated like in Bio.Tools"_, but it is not part of the issue description or something we discussed during backlog refinement. Is it expected that this redesign should be done in this issue?
2. It would probably be easy to improve global search as well to make the category with the highest score the default chosen one instead of metabolites. Good idea to fix in this PR while we have the knowledge fresh?
3. Is the current limit of 6 results per category reasonable?

 <br />
 <br />
 
### Fields below will be filled in once not in draft mode

This PR closes [issue number] and relates to [issue or PR number].

**Type of change**  

- [ ] Other
 
**List of changes made**  

**Screenshot of the fix**  

**Testing**  

- Instructions on how to test

**Further comments**  

**Checklist**  

- [ ] I have performed a self-review of my own code

### DIFF
--- a/frontend/src/components/explorer/gemBrowser/GemSearch.vue
+++ b/frontend/src/components/explorer/gemBrowser/GemSearch.vue
@@ -164,6 +164,7 @@ export default {
     placeholder() {
       return 'uracil, SULT1A3, ATP => cAMP + PPi, subsystem or compartment';
     },
+
     categoryOrder() {
       function compare(a, b) {
         if (b[1].maxScore < a[1].maxScore) {
--- a/frontend/src/components/explorer/gemBrowser/GemSearch.vue
+++ b/frontend/src/components/explorer/gemBrowser/GemSearch.vue
@@ -50,7 +50,7 @@
         class="notification is-large is-unselectable has-text-centered is-clickable py-1 mb-1"
         @mousedown="globalSearch()"
       >
-        Limited to 50 results per type. Click here to search all integrated GEMs
+        Limited to 6 results per type. Click here to search all integrated GEMs
       </div>
       <div v-show="!showLoader" v-if="searchResults.length !== 0" class="resList">
         <template v-for="type in categoryOrder">
--- a/frontend/src/components/explorer/gemBrowser/GemSearch.vue
+++ b/frontend/src/components/explorer/gemBrowser/GemSearch.vue
@@ -53,7 +53,7 @@
         Limited to 50 results per type. Click here to search all integrated GEMs
       </div>
       <div v-show="!showLoader" v-if="searchResults.length !== 0" class="resList">
-        <template v-for="type in resultsOrder">
+        <template v-for="type in categoryOrder">
           <div
             v-if="searchResults[type] && searchResults[type]['results'].length !== 0"
             :key="type"
@@ -155,7 +155,6 @@ export default {
   computed: {
     ...mapState({
       model: state => state.models.model,
-      resultsOrder: state => state.search.categories,
       searchTermString: state => state.search.searchTermString,
     }),
     ...mapGetters({
@@ -164,6 +163,17 @@ export default {
     }),
     placeholder() {
       return 'uracil, SULT1A3, ATP => cAMP + PPi, subsystem or compartment';
+    },
+    categoryOrder() {
+      function compare(a, b) {
+        if (b[1].maxScore < a[1].maxScore) {
+          return -1;
+        }
+        return 1;
+      }
+      const cats = Object.entries(this.searchResults).sort(compare);
+
+      return cats.map(x => x[0]);
     },
   },
   watch: {
--- a/frontend/src/components/explorer/gemBrowser/GemSearch.vue
+++ b/frontend/src/components/explorer/gemBrowser/GemSearch.vue
@@ -55,38 +55,49 @@
       <div v-show="!showLoader" v-if="searchResults.length !== 0" class="resList">
         <template v-for="type in resultsOrder">
           <div
-            v-for="(r, i2) in searchResults[type]"
-            :key="`${r.id}-${i2}`"
-            class="searchResultSection px-1 py-0"
+            v-if="searchResults[type] && searchResults[type]['results'].length !== 0"
+            :key="type"
           >
-            <hr v-if="i2 !== 0" class="m-0" />
-            <div>
-              <span v-if="type === 'metabolite' || type === 'gene'" class="pr-1">
-                <span class="has-text-primary is-clickable" @mousedown="handleClickResult(type, r)">
-                  <span class="icon is-medium is-left" title="Gem Browser">
-                    <i class="fa fa-table" />
+            <div
+              v-for="(r, i2) in searchResults[type]['results']"
+              :key="`${r.id}-${i2}`"
+              class="searchResultSection px-1 py-0"
+            >
+              <hr v-if="i2 !== 0" class="m-0" />
+              <div>
+                <span v-if="type === 'metabolite' || type === 'gene'" class="pr-1">
+                  <span
+                    class="has-text-primary is-clickable"
+                    @mousedown="handleClickResult(type, r)"
+                  >
+                    <span class="icon is-medium is-left" title="Gem Browser">
+                      <i class="fa fa-table" />
+                    </span>
+                  </span>
+                  <span
+                    class="has-text-icon-interaction-partner is-clickable"
+                    @mousedown="handleClickResult('interaction', r)"
+                  >
+                    <span class="icon is-medium is-left" title="Interaction Partners">
+                      <i class="fa fa-connectdevelop" />
+                    </span>
                   </span>
                 </span>
-                <span
-                  class="has-text-icon-interaction-partner is-clickable"
-                  @mousedown="handleClickResult('interaction', r)"
-                >
-                  <span class="icon is-medium is-left" title="Interaction Partners">
-                    <i class="fa fa-connectdevelop" />
-                  </span>
+                <span class="has-text-link is-clickable" @mousedown="handleClickResult(type, r)">
+                  <b class="is-capitalized">{{ type }}:</b>
+                  <label
+                    class="is-clickable"
+                    v-html="formatSearchResultLabel(type, r, searchTermString)"
+                  ></label>
                 </span>
-              </span>
-              <span class="has-text-link is-clickable" @mousedown="handleClickResult(type, r)">
-                <b class="is-capitalized">{{ type }}:</b>
-                <label
-                  class="is-clickable"
-                  v-html="formatSearchResultLabel(type, r, searchTermString)"
-                ></label>
-              </span>
+              </div>
             </div>
           </div>
           <!-- eslint-disable-next-line vue/valid-v-for vue/require-v-for-key -->
-          <hr v-if="searchResults[type] && searchResults[type].length !== 0" class="bhr p-0" />
+          <hr
+            v-if="searchResults[type] && searchResults[type].results.length !== 0"
+            class="bhr p-0"
+          />
         </template>
       </div>
       <div v-show="showLoader" class="has-text-centered">
@@ -215,7 +226,7 @@ export default {
         const keyList = Object.keys(this.searchResults);
         for (let i = 0; i < keyList.length; i += 1) {
           const k = keyList[i];
-          if (this.searchResults[k].length) {
+          if (this.searchResults[k].results.length) {
             this.showSearchCharAlert = false;
             this.noResult = false;
             break;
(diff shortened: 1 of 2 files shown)


e0 (commented):
Looks really good so far!

I think it's a great idea to make the change for the global search page as well. It's closely related enough to the original issue and should be a relatively small amount of work.

frontend/src/components/explorer/gemBrowser/GemSearch.vue
```diff
@@ -50,43 +50,54 @@
         class="notification is-large is-unselectable has-text-centered is-clickable py-1 mb-1"
         @mousedown="globalSearch()"
       >
-        Limited to 50 results per type. Click here to search all integrated GEMs
+        Limited to 6 results per type. Click here to search all integrated GEMs
```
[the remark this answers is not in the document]
mihai-sysbio: If the search performance allows it, 10 per category sounds better to me.

mihai-sysbio:
> 1. In the issue there is a [comment](https://github.com/MetabolicAtlas/private-issues/issues/151#issuecomment-1041407234)  _"The GemSearch results should be tabulated like in Bio.Tools"_, but it is not part of the issue description or something we discussed during backlog refinement. Is it expected that this redesign should be done in this issue?

Indeed. It's also that I do not interpret a different sorting to be a _layout_ improvement - some further clarifications are needed I think. Let's take it up in the standup.

frontend/src/store/modules/search.js
```diff
@@ -57,10 +69,15 @@ const getters = {
       Object.entries(results).map(([k, v]) => [
         k,
         (() => {
-          if (v === 0) {
-            return v;
+          if (v.results === 0) {
```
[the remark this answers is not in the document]
inghylt: Lol saw your commit fixing this, never mind

frontend/src/store/modules/search.js
```diff
@@ -57,14 +74,21 @@ const getters = {
...
-            return v;
+          if (v.results.length === 0) {
+            return { topScore: v.topScore, results: v.results };
           }
-          return v.sort((a, b) => sortResultsSearchTerm(a, b, state.searchTermString));
+          return {
+            topScore: v.topScore,
+            results: v.results
+              .sort((a, b) => sortResultsSearchTerm(a, b, state.searchTermString))
+              .slice(0, 10),
```
inghylt, line 84: This previous number was not completely random, with 6 instead of 10 you can actually see that there are other components as well in the result list + you don't have to scroll through as many items

e0: Sorry, I forgot to check in a local change that I mentioned in [this comment](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/887#discussion_r812718805).

inghylt: Cool! I'm still feel like it is looks better with fewer items though. Would be interesting to get the advisors feedback on this (or the team can look at it together on Friday perhaps)

frontend/src/store/modules/search.js
```diff
@@ -39,24 +39,29 @@ const categorizeResults = results => {
...
     }, 0) === 0,
 
-  categorizedGlobalResults: state => {
-    const results = categorizeResults(state.globalResults);
-    return Object.fromEntries(Object.entries(results).map(([k, v]) => [k, v.results]));
-  },
-
-  // eslint-disable-next-line no-unused-vars
-  categorizedGlobalResultsCount: (state, _getters) =>
+  categorizedGlobalResultsWithScores: state => categorizeResults(state.globalResults),
```
inghylt, line 54: Maybe just name it categorizedGlobalResults? Since it's with scores by default from now on. And maybe nice with some newlines between the  variables to improve readability?

inghylt:
Since  #886 have been merged, I guess we should try to remove function sortResultsSearchTerm and solely sort based on the scores within the different components? Either in a new issue or in this PR. Feels related to this PR but it might be that we introduce to many changes?

frontend/src/components/explorer/gemBrowser/GemSearch.vue
```diff
@@ -144,12 +155,12 @@ export default {
...
-      resultsOrder: state => state.search.categories,
       searchTermString: state => state.search.searchTermString,
     }),
     ...mapGetters({
       models: 'models/models',
       searchResults: 'search/categorizedAndSortedResults',
+      componentTypeOrder: 'search/resultsComponentTypeOrder',
     }),
     placeholder() {
       return 'uracil, SULT1A3, ATP => cAMP + PPi, subsystem or compartment';
```
nanjiangshu, line 166: Not sure if it is appropriate to raise this question in this PR. I think the examples given in the placeholder is not very good. 
1. the first three terms are detailed examples but the last one `subsystem or compartment` are types, better to be consistent
2. The example for the equation  `ATP => cAMP + PPi` might not be optimal, since the top ranked hits are not that relevant to this equation. I vaguely remember @MalinAhlberg has also raised this question previously.

frontend/src/components/explorer/gemBrowser/GemSearch.vue
```diff
@@ -317,7 +328,7 @@ export default {
     z-index: 30;
 
     .resList {
-      max-height: 22rem;
+      max-height: 33rem;
```
nanjiangshu, line 331: Can we even make the width larger? For desktop users, the current width of the GemSearch window is usually far away from the width of the browser. For mobile user, its the same.

frontend/src/store/modules/search.js
nanjiangshu: I agree with @inghylt that a small number is better in the quick search. As a user point of view, when using quick search, I'm interested if there is any hit and take a look at it. I don't even bother to check details of all 10 found hits. Maybe even 5 is enough.

If we are afraid that users may be confused why there are so little hits because they have missed the option for the global search, we can probably make some effort to highlight the notification of global search so that it won't be missed.

nanjiangshu:
Nice work! I think the new design by ranking the category with topScore hit on top is a great improvement to return the most relevant search result to the user. 

However, I have a feeling that it is usually `Genes` have the highest score when you search for terms in other components, for example, `cis`, `fatty`,  `trans`, `bile acid`, not sure if it is ideal.

frontend/src/store/modules/search.js
mihai-sysbio: The reasoning behind having more matches is to capture the compartmentalized metabolites, which can be more that 5.

inghylt: Summary of discussion: The chosen number is not too big of a deal but choosing either 5 or 10 seems more natural. Suggested that during demo, show advisors both 6 (5?) and 10 and let them decide what's best

frontend/src/store/modules/search.js
```diff
@@ -10,21 +10,30 @@ const data = {
 
 const categorizeResults = results => {
   const categorizedResults = data.categories.reduce(
-    (obj, category) => ({ ...obj, [category]: [] }),
+    (obj, category) => ({ ...obj, [category]: { maxScore: 0, results: [] } }),
```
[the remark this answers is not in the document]
MalinAhlberg: Resolved by 2a77aa2d0145a835edbb793f785fc82c505076a6

frontend/src/components/explorer/gemBrowser/GemSearch.vue
```diff
@@ -154,6 +164,18 @@ export default {
...
       return 'uracil, SULT1A3, ATP => cAMP + PPi, subsystem or compartment';
     },
+
+    categoryOrder() {
+      function compare(a, b) {
+        if (b[1].maxScore < a[1].maxScore) {
+          return -1;
+        }
+        return 1;
+      }
```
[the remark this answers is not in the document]
MalinAhlberg: Resolved by  0266588f88bfa94484dc63db736456a66df66e2b

frontend/src/components/explorer/gemBrowser/GemSearch.vue
inghylt: Resolved in commit 809b9df34a4e1e30b1bc5e3292835a7f43661afc

frontend/src/store/modules/search.js
inghylt: Resolved by cd3fecc210c7c76fc2a111de695b003f42e5a044 and de2a5d0af60b5f313613608bc8bd2b04a7cfda7e

frontend/src/components/explorer/gemBrowser/GemSearch.vue
inghylt: Nice that you noticed this! We think it makes sense to discuss this live at next standup. Another point to talk about is that if you start typing `ATP` the placeholder text disappear, i.e. you have to remember `ATP => cAMP + PPi` since you can't copy paste

MalinAhlberg:
> Apart from that, I noticed another weird behavior in the global search. The keyboard response in the search bar is very slow and responsiveness. But in the same page, if you type in the filtering field such as `Name` or `Subsystem`, the keyboard response is instant. I don't see a similar behavior neither on https://dev-metatlas.csbi.chalmers.se/search nor https://metabolicatlas.org/search.

Nice that you noticed this! This happens when there is a large search result loaded. When typing in the search field at `/search` (empty result) the responsivness is good. We tested this for different versions, and the behavior seems to be the same at least back to d59957b1.

Regarding your other findings, we suggest that we discuss them during stand-up today.

e0 (approved):
Looks great and seems to solve the related issue! I like the new size of the search results section as well.

frontend/src/components/explorer/gemBrowser/GemSearch.vue
```diff
@@ -26,7 +26,6 @@
           data-hj-whitelist
           type="text"
           class="input"
-          :placeholder="placeholder"
```
nanjiangshu, line 29: I'm a bit confused. Have we decided to remove the placeholder completely or just remove the formula?

frontend/src/components/explorer/gemBrowser/GemSearch.vue
inghylt: 6943af784047ef93c66e4872ec5940e35d505b93 removes the placeholder and issue #897 covers adding a more appropriate one. Please resolve this comment if you think this covers the problems raised

frontend/src/components/explorer/gemBrowser/GemSearch.vue
inghylt: I removed it completely, but I was not present during the discussion today so please add a commit as you see fit if you don't think this is what was suggested

frontend/src/components/explorer/gemBrowser/GemSearch.vue
nanjiangshu: I see. make sense to me.

MalinAhlberg:
Thanks @e0 for the refactoring and global search fixes you did to this! Appreciated!

MalinAhlberg:
New issues:
- #895 
- #896 
- #897

e0:
I have done some research and made some early mockups of the important components and ideas. After the mockups there is some additional notes. I'm happy to provide more detailed mockups but would appreciate some input before dedicating more time to a certain direction.

## Layout

image

The filter area is intended to be expandable and collapsable on mobile devices. When expanded, it should take up the entire screen except the navbar.

## Filter

image


There are three kinds of filters included in this mockup. 

1. Checkbox group: Ideal for multi-select of filters that have few options.
2. Text input: Ideal for fields such as name and formula.
3. Multi-select with search: Ideal for multi-select of filters that have many options.

Ideas for other types of filters. Not sure if these are something we want to implement. Let me know if that's the case and I can add mockups for them.

- Range slider: Ideal for component counts, such as reactions count for a subsystem.
- Component that handles lists of values, such as alternate names and synonyms. Perhaps this can be included in the Text input?
- Custom component for formula: Would be neat to be able to filter formula by included elements.
- Custom component for gene rule: No concrete idea for this yet. The gene rules could be complex and nested. Happy to discuss this in more details if there is interest.

## Search results

Three alternatives are included here. Is there anyone you prefer? We could even have all three views and add a toggle (button group) for the user to pick.

1. Search engine style. Should be mobile friendly without requiring much adaptation. Can fit a lot of info.
image

2. Table view. Similar to what we currently have. Certain fields, such as id and name, could be merged into a single column if it makes sense.
image

3. Random components style. Nice way to visualize different types of components.
image

## Inspiration

- http://ensembl.org
- https://www.uniprot.org
- http://prisjakt.nu

Issue #710, opened 2021-10-05
Improve no results found message in global search


inghylt:
As a user, I would like to have a clear message if my global search returns no hits, so that I am not confused

The image below displays the current message, which is a bit unclear

![image.png](https://images.zenhubusercontent.com/5fc0f648ba2981921764866b/96d3c773-7a9c-4aec-a3de-65d5c560747c)

mihai-sysbio:
- [ ] an introductory sentence
- [ ] examples (urls) for each

PR #888, opened 2022-02-22
feat: improve NotFound message in the global search

### LINKED ISSUE #744: Remove legacy info in global search
As a user, I want information regarding the global search to be up to date, so that I know that I can for example search for external id:s

**Describe the solution you'd like**
The information returned if there are not matches `"If this is an alias or external identifier, it means it is not present in any of the models."` should probably be replaced by `"If this is an alias it means it is not present in any of the models"` or similar. We might want to add new information to the Documentation as well

**Additional context**
Add any other context or screenshots about the feature request here.

nanjiangshu:
This PR closes #744  and #710

1. make the NotFound message more precisely
2. Add examples in the suggested items that can be searched in the global search.
3. Changed the term `External identifiers` to `Cross references`

**Type of change**  

- [x] Other: message improvement
 
**List of changes made**  

Show clearer message to the user when there is no hit in the global search

**Screenshot of the fix**  

**Testing**  

- Instructions on how to test
e.g. search with `/search?term=Cytocsolacbcd`

**Further comments**  

I could not find `cross references` for subsystems and thus I have removed it from the list. Please double check.

**Checklist**  

### DIFF
--- a/frontend/src/components/SearchTable.vue
+++ b/frontend/src/components/SearchTable.vue
@@ -175,8 +175,8 @@
                   </li>
                   <li>
                     Cross references (e.g.
-                    <router-link :to="{ name: 'search', query: { term: 'R03004' } }">
-                      R03004
+                    <router-link :to="{ name: 'search', query: { term: 'RCR13746' } }">
+                      RCR13746
                     </router-link>
                     )
                   </li>
--- a/frontend/src/components/SearchTable.vue
+++ b/frontend/src/components/SearchTable.vue
@@ -101,15 +101,15 @@
                 <ul>
                   <li>
                     ID (e.g.
-                    <router-link :to="{ name: 'search', query: { term: 'MAM03146l' } }">
-                      MAM03146l
+                    <router-link :to="{ name: 'search', query: { term: 'MAM01738p' } }">
+                      MAM01738p
                     </router-link>
                     )
                   </li>
                   <li>
                     Name or aliases (e.g.
-                    <router-link :to="{ name: 'search', query: { term: 'VLDL remnant' } }">
-                      VLDL remnant
+                    <router-link :to="{ name: 'search', query: { term: 'dopamine-O-quinone' } }">
+                      dopamine-O-quinone
                     </router-link>
                     )
                   </li>
@@ -118,17 +118,17 @@
                     <router-link
                       :to="{
                         name: 'search',
-                        query: { term: 'C136286H232084N13476O38613P3271S174R7267' },
+                        query: { term: 'C8H10NO2' },
                       }"
                     >
-                      C136286H232084N13476O38613P3271S174R7267
+                      C8H10NO2
                     </router-link>
                     )
                   </li>
                   <li>
                     Cross references (e.g.
-                    <router-link :to="{ name: 'search', query: { term: 'M03146' } }">
-                      M03146
+                    <router-link :to="{ name: 'search', query: { term: 'CE5276' } }">
+                      CE5276
                     </router-link>
                     )
                   </li>
--- a/frontend/src/components/SearchTable.vue
+++ b/frontend/src/components/SearchTable.vue
@@ -99,27 +99,106 @@
                 <h5 class="title is-5">You may search for the following items:</h5>
                 <h6 class="title is-6 m-0">Metabolites</h6>
                 <ul>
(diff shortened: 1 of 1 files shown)


MalinAhlberg:
Nice!
Would it make sense to make the examples clickable (eg adding links to `/search?term=MAM03146l` etc)?

frontend/src/components/SearchTable.vue
```diff
@@ -93,35 +93,33 @@
...
+                <h5 class="title is-5">You may search for the following items:</h5>
+                <h6 class="title is-6 m-0">Metabolites</h6>
                 <ul>
-                  <li>ID</li>
-                  <li>Name or aliases</li>
-                  <li>Formula without charge</li>
-                  <li>External identifiers</li>
+                  <li>ID (e.g. MAM03146l)</li>
+                  <li>Name or aliases (e.g. VLDL remnant)</li>
+                  <li>Formula without charge (e.g. C136286H232084N13476O38613P3271S174R7267)</li>
```
inghylt, line 104: Haha `C136286H232084N13476O38613P3271S174R7267` is obv. correct but it is very long, maybe something shorter like `C8H10NO2` of `MAM01738p` would make sense?

frontend/src/components/SearchTable.vue
```diff
@@ -93,35 +93,33 @@
                 <b>
                   <i>{{ searchedTerm }}</i>
                 </b>
-                <br />
-                If this is an alias or external identifier, it means it is not present in any of the
-                models.
+                among all {{ Object.keys(models).length }} integrated GEMs
```
inghylt, line 96: Nice that this is not hard coded :)

frontend/src/components/SearchTable.vue
nanjiangshu: changed in 0faf5103

nanjiangshu:
> Nice! Would it make sense to make the examples clickable (eg adding links to `/search?term=MAM03146l` etc)?

Good suggestion! Updated in 15226678 
However, prettier added unexpected white space before `)` as shown below
Screenshot 2022-02-22 at 15 38 56

I think this is related to this issue https://app.zenhub.com/workspaces/metatlas-roadmap-60d96b68da0fb10019d26deb/issues/metabolicatlas/metabolicatlas/870

MalinAhlberg:
Really nice with the links!
I find two examples which might be worth changing:
-  "Reaction PMID 11579996", no  reactions are found in this search. 
- "Reaction cross reference R03004", this gives results, but as far as I can tell, the top results are incorrect.
I guess this is not a big deal, but if we give examples, maybe we should show things that are working really well.

e0:
Taking a look at this one now.

e0 (approved):
Looks really good!

I think it's nice with links (looks fine imo and better UX) and I remember us mentioning it during the backlog refinement as well when voting for this issue.

It's really nice that @MalinAhlberg spotted the few test results that didn't show the desired matches at the top. If it's not too much work, it would be nice to replace them. Not too big of a deal though so I will approve the PR already and you can decided if you want to change this or not.

frontend/src/components/SearchTable.vue
inghylt: It seems to be the same long formula still though?

nanjiangshu: Sorry, I forgot to push the changes

nanjiangshu:
Hi @MalinAhlberg and @e0, good spot for the problem with the PMID `11579996` and the cross reference `R03004`. I've updated the example cross reference for the Reactions in the commit 4b3cb938.

As to the PMID `11579996`, I test with several PMIDs but none of them returned reasonable hits. It seems searching does not work well for PMID if you look at the code for `search` in the API. I will create a new issue for that.

nanjiangshu:
Issue #889 is created

frontend/src/components/SearchTable.vue
mihai-sysbio: ➕

mihai-sysbio:
Well done! I very much agree with all the previous comments, so I could approve, but I think it's nicer to re-request reviews to confirm all feedback has been addressed.

nanjiangshu:
Hi @inghylt I have re-requested your review to see if you have any other comments for this PR.

inghylt (approved):
Nice work!

PR #892, opened 2022-02-28
feat: separate model search and global search implementation and reduce model search response content

### LINKED ISSUE #705: separate Neo4j queries in API that power GemSearch and GlobalSearch
As a developer, I want the code for the search functionality to be maintainable, so that I can work with it easily.

_Originally reported [here](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/700#issuecomment-930864502)_

e0:
This PR closes #705.

**Type of change**  

- [x] Other
 
**List of changes made**  

- Separate model search and global search code.
- Reduce response content for model search.
- Fix bug where too many metabolites may be returned, as mentioned [here](https://github.com/MetabolicAtlas/private-issues/issues/101#issuecomment-1047937655).

**Testing**  

- Instructions on how to test

1. Click the search icon in the nav bar and select `Human-GEM`.
2. Search for `h2o` and make sure results for `Metabolite`, `Gene`, and `Reaction` look correct.
3. Search for `peroxisom` and make sure results for `Subsystem`, and `Compartment` look correct.

**Further comments**  

- If it helps, you can use the dev or prod server to compare the search results.
- The search results for metabolites may be slightly different as before due to [the bug fix](https://github.com/MetabolicAtlas/MetabolicAtlas/commit/3b3b01e5388e2fce301f841f824e350a2ddc32f9).

**Checklist**  

- [ ] I have checked so that the same data as before is returned by the API (leaving this unchecked this time as the data is different but more correct now)

### DIFF
--- a/api/src/neo4j/queries/search/globalSearch.js
+++ b/api/src/neo4j/queries/search/globalSearch.js
@@ -1,0 +1,400 @@
+import queryListResult from 'neo4j/queryHandlers/list';
+import { sanitizeSearchString, intersect } from 'utils/utils';
+import { MODELS, COMPONENT_TYPES } from 'neo4j/queries/search/helper';
+
+const fetchCompartmentalizedMetabolites = async ({
+  ids,
+  model,
+  version,
+  limit,
+  viaMetabolties,
+}) => {
+  if (!ids) {
+    return null;
+  }
+
+  let statement = ``;
+
+  if (viaMetabolties) {
+    statement += `
+WITH ${JSON.stringify(ids)} as mids
+UNWIND mids as mid
+MATCH (:Metabolite:${model} {id:mid})-[${version}]-(cm:CompartmentalizedMetabolite)
+WITH DISTINCT(cm.id) as cmid
+`;
+  } else {
+    statement += `
+WITH ${JSON.stringify(ids)} as cmids
+UNWIND cmids as cmid
+`;
+  }
+
+  statement += `
+CALL apoc.cypher.run('
+  MATCH (ms:MetaboliteState)-[${version}]-(:Metabolite)-[${version}]-(:CompartmentalizedMetabolite:${model} {id: $cmid})
+  RETURN ms { id: $cmid, .* } as data
+  
+  UNION
+  
+  MATCH (:CompartmentalizedMetabolite:${model} {id: $cmid})-[${version}]-(c:Compartment)-[${version}]-(cs:CompartmentState)
+  RETURN { id: $cmid, compartment: cs { id: c.id, .* } } as data
+  
+  UNION
+  
+  MATCH (:CompartmentalizedMetabolite:${model} {id: $cmid})-[${version}]-(:Reaction)-[${version}]-(s:Subsystem)
+  WITH DISTINCT s
+  MATCH (s)-[${version}]-(ss:SubsystemState)
+  RETURN { id: $cmid, subsystem: COLLECT({id: s.id, name: ss.name}) } as data
+', {cmid:cmid}) yield value
+RETURN apoc.map.mergeList(apoc.coll.flatten(
+	apoc.map.values(apoc.map.groupByMulti(COLLECT(value.data), "id"), [value.data.id])
+)) as metabolites
+`;
+
+  if (limit) {
+    statement += `
+LIMIT ${limit}
+`;
+  }
+
+  return queryListResult(statement);
+};
+
+const fetchGenes = async ({ ids, model, version }) => {
+  if (!ids) {
+    return null;
+  }
+
+  const statement = `
+WITH ${JSON.stringify(ids)} as gids
+UNWIND gids as gid
+CALL apoc.cypher.run("
+  MATCH (gs:GeneState)-[${version}]-(:Gene:${model} {id: $gid})
+  RETURN { id: $gid, name: gs.name } as data
+  
+  UNION
+  
+  MATCH (:Gene:${model} {id: $gid})-[${version}]-(r:Reaction)
+  WITH DISTINCT r
+  MATCH (r)-[${version}]-(s:Subsystem)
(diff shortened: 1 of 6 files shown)


api/src/neo4j/queries/search/helper.js
```diff
@@ -0,0 +1,17 @@
+const INTEGRATED_MODELS = require('data/integratedModels');
+
+const MODELS = INTEGRATED_MODELS.map(m => ({
+  label: m.short_name.replace('-GEM', 'Gem'),
+  name: m.short_name,
+}));
```
mihai-sysbio, line 6: Just looking at this helper, it gives the vibe that it would be useful across the entire `api`, so maybe there's a better place for it?

api/test/search.test.js
```diff
@@ -6,7 +6,7 @@ describe('search', () => {
     await driver.close();
   });
 
-  test('model search to have 50 results per component type if it exceeds the limit 50', async () => {
+  test('model search should have max 50 results per component type', async () => {
```
mihai-sysbio, line 9: I'm confused here, the limit was 50 for all results and now it's 50 per component type?

mihai-sysbio (commented):
From what I'm seeing in the request response for the quick search, the details for almost all components are relatively compact. However, for the reactions they are perhaps too rich for now and also in the future:
image

e0:
> From what I'm seeing in the request response for the quick search, the details for almost all components are relatively compact. However, for the reactions they are perhaps too rich for now and also in the future:
> image

For the reactions, the metabolites are needed to display the equation. Perhaps we can consider removing it from the frontend in the future but that should be in a new issue imo.

image

api/src/neo4j/queries/search/helper.js
e0: That could be the case but I generally prefer to refactor things only they are needed. For example if we want to use `MODELS` in another file later but they might need to be formatted differently.

api/test/search.test.js
e0: The search runs in two steps.

1. Full text search that returns ids from all nodes, including `Metabolite` and `CompartmentalizedMetabolite`. In this case, the limit is 50.
2. More data is fetched for each component type. For the case of `Metabolite`, for example if there are 20 ids found, this could lead to more than 50 results, so the limit is further passed here to make sure it does not exceed 50. The problem is that we were fetching additional data for `Metabolite` ids and `CompartmentalizedMetabolite` ids separately before. So the combination of results returned for `Metabolite` ids and `CompartmentalizedMetabolite` ids are then merged back together, which could result in more than 50 results and that is why the test was failing before.

Does that make sense? This was a tricky implementation so let me know if you want me to elaborate on anything.

mihai-sysbio:
> For the reactions, the metabolites are needed to display the equation. Perhaps we can consider removing it from the frontend in the future but that should be in a new issue imo.

At a glance, metabolite fields like `isCurrency`, `formula`, `id`, `charge` should not play any role in how the equation is displayed. Most likely, between `name`, `fullName` and `compartment` one or two of these fields could be dropped as well.

api/test/search.test.js
mihai-sysbio: Ah I see, it does make sense now!

e0:
> > For the reactions, the metabolites are needed to display the equation. Perhaps we can consider removing it from the frontend in the future but that should be in a new issue imo.
> 
> At a glance, metabolite fields like `isCurrency`, `formula`, `id`, `charge` should not play any role in how the equation is displayed. Most likely, between `name`, `fullName` and `compartment` one or two of these fields could be dropped as well.

I think `id` is needed but you are right about some other ones. Will try to give this another look now. Thanks!

api/test/search.test.js
```diff
@@ -62,4 +63,16 @@ describe('search', () => {
       expect(name).toEqual('Human-GEM');
     }
   });
+
+  test('global search should return 50 results for multiple models', async () => {
```
inghylt, line 67: Maybe we should remove the number 50? Since we are not specifying a limit for global search (right?) and we are only testing `toBeGreaterThan(0)` the title is a bit confusing

inghylt (approved):
Nice work! Have tested and seems to work fine. I like the separation and renaming to modelSearch and globalSearch, makes it much clearer.

Would we like to add a test for sensible ranking scores for globalSearch?

api/test/search.test.js
e0: Yea this seems like a blunder, thank you for spotting it! Fixed in: https://github.com/MetabolicAtlas/MetabolicAtlas/pull/892/commits/ffd0952393a62201f4ca49cec77f07d540276560

e0:
> Would we like to add a test for sensible ranking scores for globalSearch?

This would be nice to have but maybe we can wait with it until the work for redoing the global page? Just in case the searching logic or results should be adjusted.

api/src/neo4j/queries/search/globalSearch.js
```diff
@@ -286,43 +270,12 @@ const globalSearch = async ({ searchTerm, version, limit }) => {
...
-      })),
-    },
-  };
-};
-
 /*
  * The search consists of two steps
  * 1. Do a fuzzy search over all nodes covered by full-text search index
  * 2. Fetch results for each component type (parallelly) and return result
  */
```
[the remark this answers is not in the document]
e0: Hi @MalinAhlberg, thanks for noting this. Not sure if you have marked the right lines here but I think the following comments could be removed.

```
  // Metabolites are not included as it would mess with the limit and
  // relevant metabolites should be matched through CompartmentalizedMetabolites
```

e0:
Thanks for the reviews everyone! I have made some small docs related changes since the approvals. Since this issue is a blocker for some other issues and the changes are minimal, I will not re-request the reviews and merge this.

Issue #784, opened 2021-11-22
Explore if we can speed up neo4j queries through parallelization


inghylt:
We should spend some time exploring if we can speed up neo4j queries through parallelization. This was done for Interaction Partners in #728, where the response time was reduced by 75%. Might be that this is useful in other queries as well.

Suggested to list which queries have the slowest response time today and focus on those.

**Acceptance criteria**
- [ ] find slow queries that might be improved if parallelization would be used
- [ ] implement parallelization where needed, but if it's too much (too many places) create separate issue(s)

mihai-sysbio:
- [ ] related reactions which are capped to 200 for the same latency issue

inghylt:
As noted in [this comment](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/900#issuecomment-1062792951), the response time is slow for the reactions of https://metabolicatlas.org/explore/Fruitfly-GEM/gem-browser/metabolite/MAM02040c. Might be that neo4j parallelization can improve that

MalinAhlberg:
Slow queries:
- comparison of three model:
   - ` api/src/neo4j/queries/compare.js: compareThree()`
  - eg
       ```
    MATCH (a:Reaction:FruitflyGem)-[:V1_2_0]-(e:ExternalDb)-[:V1_12_0]-(b:Reaction:HumanGem)
    USING JOIN on e
    MATCH (a)-[:V1_2_0]-(e)-[:V1_3_0]-(c:Reaction:RatGem)
    RETURN { FruitflyGem: COUNT(DISTINCT(a)), HumanGem: COUNT(DISTINCT(b)), RatGem: COUNT(DISTINCT(c)) }
      ```

 - related reactions:
   -  `api/src/neo4j/queries/relatedReactions.js: getRelatedReactions()`
   - eg 
 
       ```
     MATCH (:CompartmentalizedMetabolite:HumanGem {id: 'MAM01597c'})-[:V1_12_0]-(r:Reaction)
     WITH r.id as rid
     
     LIMIT 1000
     
     CALL apoc.cypher.run("
       MATCH (rs:ReactionState)-[:V1_12_0]-(:Reaction:HumanGem {id: $rid})
       RETURN rs { id: $rid, .* } as data
      
       UNION
      
       MATCH (:Reaction:HumanGem {id: $rid})-[:V1_12_0]-(cm:CompartmentalizedMetabolite)
       WITH DISTINCT cm
[...]

PR #900, opened 2022-03-08
fix: increase timeout to fix error for time-consuming global searches

### LINKED ISSUE #765: Searching for citric acids in global search leads to error
**Describe the bug**

When searching for citric acid in quick search, and then going to global search from there, leads to an error 

**Steps to reproduce**

1. Search for citric acid in quick search
2. Press link to global search
3. Note the error being thrown

**Expected behavior**

That going from quick to global search works as when searching for citric acid in just global search, i.e. that only `No matches for citric acid` will be returned and not an error message

**Screenshots**

**Additional context**

Note that this is only applicable when going from quick search to global search, and not when searching in global search directly

inghylt:
This PR closes #765

**Type of change**  

**List of changes made**  

Limit for neo4j transaction timeout increased to make sure that time consuming global searches does not lead to 400 Bad Request.

**Screenshot of the fix**  

**Testing**  

This has been tested, i.e. deployed,  on the dev machine. Following links cause error with a lower limit but work with 40 sek:
`https://dev-metatlas.csbi.chalmers.se/search?term=citric%20acid`
`https://dev-metatlas.csbi.chalmers.se/search?term=C5-branched%20dibasic%20acid%20metabolism`
But please experiment more with global search and see if everything is working correctly

**Further comments**  

A limit of the previously suggested 30 sek continued to cause 400 Bad Request for the example links above, but 40 sek seems to do the trick.

**Checklist**  

- [x My changes generate no new warnings

### DIFF
--- a/docker-compose-remote.yml
+++ b/docker-compose-remote.yml
@@ -40,7 +40,7 @@ services:
 
   neo4j:
     environment:
-      - NEO4J_dbms_transaction_timeout=15s
+      - NEO4J_dbms_transaction_timeout=40s
       - NEO4J_dbms_memory_heap_initial__size=5G
       - NEO4J_dbms_memory_off__heap_max__size=8G
     restart: unless-stopped


MalinAhlberg (commented):
Nice! I played around a bit but weren't able to break anything... :+1:  Even though it's a bit tedious to wait for 40s, it's still better then getting an error message, I think.

Don't know if this applies, but just read [here](https://neo4j.com/developer/kb/understanding-transaction-and-lock-timeouts/) that it might be good to set `dbms.lock.acquisition.timeout` as well?

nanjiangshu:
I can not reproduce the error on the `dev` server when searching for `citric acid`. Is that because  @inghylt  have deployed this branch to the `dev` server?

inghylt:
> This has been tested, i.e. deployed, on the dev machine. 

@nanjiangshu with this I tried to communicate that it was deployed on the dev machine but obv. not clear enough, sorry about that

nanjiangshu:
> Another thing to try is multiple request at once, and check the resource usage with `docker stats`.
I tried to do 40 requests a the same time with  
```
for i in `seq 20` ; do open  "https://dev-metatlas.csbi.chalmers.se/search?term=citric%20acid" -a firefox ; done 

for i in `seq 20` ; do open  "https://dev-metatlas.csbi.chalmers.se/search?term=C5-branched%20dibasic%20acid%20metabolism" -a firefox ; done 
```
and no error message was shown.

e0 (approved):
This worked well for most of the searches I tried. I did notice some problems when making two searches for `00` and `01` at about the same time (two different tabs). Afterwards I did search for just `00` again and it loaded but took ~38s. This was almost immediately afterwards though. After I waited a while and searched for `01` again it took ~26s.

Since it seems to retry for failed searches, instead of caching the error response, I think increasing the timeout is a good solution. I'm still surprised that this works without having to increase the nginx timeout though.

MalinAhlberg (approved):
> @MalinAhlberg Nice that you made some research! I don't think it really relates to this issue as the transaction does not hang indefinitely, it's just slower than our other transactions, but I'm up for creating an issue regarding adding `dbms.lock.acquisition.timeout`.

The way I understand it, it does not need to hang indefinitely for this to be a potential problem. But I agree, it's not in the scope of this PR. Also tested with a three concurrent searches (with different search strings), still didn't break (although slow). So I approve!

e0:
Just noticed this on the prod server. Also tested on the dev server and it loaded in ~19s so doesn't look like we need to report an additional bug. 👍

image

inghylt:
@e0 Nice! Should we create an issue perhaps to try to speed up the response time?

mihai-sysbio:
@e0 that example is H20 in cytosol - one of the most popular compartmentalized metabolites (and also a currency metabolite), so it is understandable. Also I think the limit was increased from 200 to 1000, and I'm not sure this special case was tested.

@inghylt perhaps describe in https://github.com/MetabolicAtlas/MetabolicAtlas/issues/784 ?

inghylt:
@MalinAhlberg Created #903 for exploring `dbms.lock.acquisition.timeout`

Issue #819, opened 2021-12-23
Broken link for gembrowser compartments


inghylt:
**Describe the bug**

When visiting a compartment in the gem browser and clicking the link for "complete list in JSON format" you get an error as result

**Steps to reproduce**

1. Visit https://metabolicatlas.org/explore/Human-GEM/gem-browser/compartment/golgi_apparatus
2. Press "complete list in JSON format" in the sentence "The complete list in JSON format of reactions / metabolites / genes is available using our API"
3. Notice that "Cannot GET /api/v2/1_10_0/compartments/golgi_apparatus" is returned

This happens for other compartments as well

**Expected behavior**

That the users get a list of reactions / metabolites / genes in JSON format

**Screenshots**

![image.png](https://images.zenhubusercontent.com/5fc0f648ba2981921764866b/a3d0b01f-8779-45ce-b7e0-e5b5bc5e7f2d)

inghylt:
Seems to be related to missing model name in url

mihai-sysbio:
A working url https://metabolicatlas.org/api/v2/compartments/inner_mitochondria?full=true&model=HumanGem

nanjiangshu:
It returned the same result with or without version number in the url 
https://metabolicatlas.org/api/v2/compartments/inner_mitochondria?full=true&model=HumanGem
https://metabolicatlas.org/api/v2/compartments/mitochondria?full=true&model=HumanGem&version=1_10_0

PR #901, opened 2022-03-09
Fix: description in the API page

### LINKED ISSUE #759: Description text in the API is out of date
**Describe the bug**
The description of the text in the API `/api/v2` is out of date in some places. 

**Screenshots**
For example, for Interaction Partners, it is valid not only for gene but also metabolite

**Additional context**
There could be places to be fixed at other places on the API page.

**Acceptance Criteria**
- [ ] Update `for a given gene` to `for a given gene or metabolite`
- [ ] Look for other related inconsistencies in the api description

nanjiangshu:
This PR closes #759

1. Fixed the outdated description for the interaction partners 
2. Fixed typos and grammatical errors
3. Fixed the outdated upper limit for related reactions

**Type of change**  

**Testing**  

http://localhost/api/v2/

**Further comments**  

The API for the random components does not return valid results if the optional field `componentTypes` is unfilled. 
that is the url
http://localhost/api/v2/random-components?model=HumanGem&componentTypes=%7B%7D returns a 400 error.
The url that returns a random sample of all components is either 
http://localhost/api/v2/random-components?model=HumanGem&componentTypes=
or 
http://localhost/api/v2/random-components?model=HumanGem

**Checklist**  

### DIFF
--- a/api/src/swagger/config.yaml
+++ b/api/src/swagger/config.yaml
@@ -189,7 +189,7 @@ paths:
       tags:
         - 'Metabolites'
       summary: 'Get related reactions for a given metabolite.'
-      description: 'This query retrives all the related reactions that involve the specified metabolite.'
+      description: 'This query retrieves all the related reactions that involve the specified metabolite.'
       operationId: 'metaboliteRelatedReactionInfo'
       produces:
         - 'application/json'
--- a/api/src/swagger/config.yaml
+++ b/api/src/swagger/config.yaml
@@ -348,7 +348,7 @@ paths:
     get:
       tags:
         - 'Interaction Partners'
-      summary: 'Get interaction partners for a given gene.'
+      summary: 'Get interaction partners for a given gene or metabolite.'
       description:
         'This query retrieves all first the order interaction partners for
         a given gene or metabolite.'
--- a/api/src/swagger/config.yaml
+++ b/api/src/swagger/config.yaml
@@ -113,7 +113,7 @@ paths:
     get:
       tags:
         - 'Genes'
-      summary: 'Get information on a specific gene.'
+      summary: 'Get information for a given gene.'
       description: 'This query retrieves all the information for a given gene.'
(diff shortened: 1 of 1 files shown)


inghylt (approved):
Nice work! I'm surprised to see that we had so many typo:s, good that you fixed them as well

mihai-sysbio (approved):
awesome 🦄

e0 (approved):
Looks really good!

Regarding random components, maybe we could have a list of default component types if the user does not provide any. Considering that we already provide the example for component types and the purpose of the endpoint, I think it's ok to consider this as something we "won't fix".

MalinAhlberg (approved):
Nice!

While you're at it, there's a typo at line 192 as well ("retrives"). I approve this anyway, in case you don't feel to fix it in this PR.

Another related thing, although it's out of scope for this PR: it would be nice to have some default values for the optional in the `config.yaml`, for example 'HumanGEM' for model.
swagger

nanjiangshu:
> Nice!
> 
> While you're at it, there's a typo at line 192 as well ("retrives"). I approve this anyway, in case you don't feel to fix it in this PR.
> 

Good catch. Updated in c94c0ea5
> Another related thing, although it's out of scope for this PR: it would be nice to have some default values for the optional in the `config.yaml`, for example 'HumanGEM' for model. 

We can probably create an issue for that.

inghylt:
@nanjiangshu When working with #819 I just saw that we say "_This query retrieves all the metabolites, genes, reactions and subsystems contained by the given compartment._" at http://localhost/api/v2/#/Compartments/compartmentInfo, but since we leave out the param `full=true` the api actually does not return the metabolites, genes or reactions. Maybe something we should fix quickly now since it seems very related to the work in this PR?

nanjiangshu:
Thanks for pointing out this problem @inghylt. I think it makes sense to fix this right now. I will create a new PR and attach to the issue #759

PR #902, opened 2022-03-09
feat: use kebab-case for documentation toc links

### LINKED ISSUE #844: inconsistent ToC links in Documentation
**Describe the bug**
The links are not consistently capitalized and they do not follow a consistent naming scheme.

**Steps to reproduce**
See the GEM Browser subpages.

**Expected behavior**
Consistent links

**Screenshots**

**Additional context**

MalinAhlberg:
This PR closes #844.

Use kebab-case for all links to the documentation.

**Type of change**  

- [X] Other
 
**Testing**  

Please test different links at the documentaiton page, and also at other places (CompareModels, DataOverlay...)

**Further comments**  

**Checklist**  

### DIFF
--- a/frontend/src/components/Documentation.vue
+++ b/frontend/src/components/Documentation.vue
@@ -469,7 +469,7 @@
           </p>
 
           <hr class="mt-6" />
-          <h4 id="search" class="is-size-4">
+          <h4 id="searching" class="is-size-4">
             <span class="icon pr-5 is-large has-text-info"><i class="fa fa-search"></i></span>
             Search
           </h4>
@@ -693,7 +693,7 @@ export default {
         {
           name: 'Search',
           icon: 'fa-search',
-          link: '#search',
+          link: '#searching',
           subsections: [
             { name: 'Quick search', link: '#quick-search' },
             { name: 'Global search', link: '#global-search' },


inghylt (approved):
Very nice! Not sure if that is something we should fix in this branch but the Resources section is a bit out of context now since we have moved that from a separate tab to under the About tab. Should we rename it to "Related Resources" and add a link or something?

nanjiangshu (approved):
Looks great!

nanjiangshu:
> Very nice! Not sure if that is something we should fix in this branch but the Resources section is a bit out of context now since we have moved that from a separate tab to under the About tab. Should we rename it to "Related Resources" in the Documentation and add a link to `http://localhost/about/resources` or something?

Good idea to change it to `Related Resources`

MalinAhlberg:
Good idea @inghylt ! New issue?

frontend/src/components/Documentation.vue
```diff
@@ -673,48 +673,48 @@ export default {
...
+            { name: 'Export graph', link: '#interaction-partners-export' },
+            { name: 'Highlights', link: '#interaction-partners-highlights' },
+            { name: 'Reactions table', link: '#interaction-partners-reactions' },
           ],
         },
         {
           name: 'Search',
           icon: 'fa-search',
-          link: '#Search',
+          link: '#search',
```
e0, line 696: For some reason, when I click on this from the ToC, it does not jump to the right section. It works correctly when refreshing the page though with `/documentation#search`. I tested on Chrome, Safari, and Firefox on macOS. Not a big problem though but it's an interesting behavior. All of the other links in the ToC seem to work. Wondering if it's something related specifically for the id `search`. The prod server with `#Search` also seems to work.

e0 (approved):
Great job!

Left one comment but not sure if it's something worth looking into.

frontend/src/components/Documentation.vue
nanjiangshu: Maybe because of caching? When I tested on all three Browsers it jumped to the right place.

e0: Follow up on this. It is because we already have an element that uses the same id.

image

MalinAhlberg: Well spotted! I renamed it to `searching` in e9a36db. Hope you all find that name ok.

e0: Looks great, thanks!

MalinAhlberg:
> Good idea @inghylt ! New issue?

Created #904 for this. Feel free to improve it.

PR #909, opened 2022-03-09
fix: remove padding to avoid code element overflow

### LINKED ISSUE #760: Error message text clashing
**Describe the bug**
Text of the error message when 2D maps are missing clashes

**To reproduce**
Visit `explore/Fruitfly-GEM/map-viewer/phenylalanine_tyrosine_and_tryptophan_biosynthesisxxx?dim=2d&panel=0&sel=&search=&coords=0,0,1,0,0,500&datatype=None&datasource=None&dataSet=None`

**Screenshots**

**Additional context**

The same behavior observed on Firefox, Safari and Chrome

inghylt:
This PR closes #760.

**Type of change**  

**List of changes made**  

Removed top and bottom padding to avoid overflow of code elements

**Screenshot of the fix**  

image

**Testing**  

Visit for example `/explore/Fruitfly-GEM/map-viewer/phenylalanine_tyrosine_and_tryptophan_biosynthesisxxx?dim=2d` or `/explore/Fruitfly-GEM/map-viewer/phenylalanine_tyrosine_and_tryptophan_biosynthesisxkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkxx?dim=2d`

**Further comments**  

- I chose not to use bulma `py` since it would need to be added for two `<code>` elements + they already had a style section so to me it made sense. Feel free to have feedback on this though
- Should I add a comment in the code regarding why the padding is 0 or is that superfluous?

**Checklist**  

### DIFF
--- a/frontend/src/components/NotFound.vue
+++ b/frontend/src/components/NotFound.vue
@@ -70,5 +70,7 @@ export default {
 <style lang="scss" scoped>
 code {
   font-size: 1em;
+  padding-top: 0px;
+  padding-bottom: 0px;
 }
 </style>


nanjiangshu (approved):
I like this fix with minimal changes. It makes sense to keep using css styling here.

mihai-sysbio (changes requested):
Could you use Bulma classes instead? I think they have something like `pb-0` and `pt-0`.

inghylt:
@mihai-sysbio 
>**Further comments**
I chose not to use bulma py since it would need to be added for two code elements + they already had a style section so to me it made sense. Feel free to have feedback on this though

Nanjiang seems to share my opinion on this https://github.com/MetabolicAtlas/MetabolicAtlas/pull/909#pullrequestreview-905514627

mihai-sysbio:
Thanks for pointing me to the explanantion @inghylt, sorry I missed it.

> I chose not to use bulma py since it would need to be added for two code elements + they already had a style section so to me it made sense. Feel free to have feedback on this though

To me, every custom styling is technical debt, since it requires manual maintenance/adjusting. So, in this case, I'd rather use Bulma classes for everything, including the font size (if it's even necessary). My long term view is that it will be easier to switch from one CSS framework to another by switching equivalent classes, rather than by going into the styling section. In short, I'm putting a lot of emphasis on the maintainability.

e0:
Wanted to chime in here on this topic. I think using Bulma classes is a step in the right direction towards better maintainability. It should also go hand in hand with keeping things DRY. So ideally we could have a custom component that is refactored out so that we would have two instances of this component instead of repeating the same Bulma classes for these two elements.

In the long term, this would mean having some kind of design system so that all the reusable components are their own individual Vue components and we would not use any `class` or custom CSS directly when building pages or bigger components.

As of now, the version in this PR is more DRY. I think one could argue both sides when it comes to which alternative is easier to improve and maintain in the future. Personally I favor the current solution with CSS because it is quite easy to identify for refactoring. It also feels less error prone if we were to change this in the future because we don't have to worry about changing Bulma classes in multiple instances. I'm happy to discuss this more in a meeting if there is interest.

mihai-sysbio (approved):

mihai-sysbio:
In the interest of merging this PR within the current sprint, I'm approving this PR. Even if changes would be required as a a follow-up of the discussion (not booked yet), we could have a separate issue for that.

PR #912, opened 2022-03-10
fix: add full=true to compartment swagger api

nanjiangshu:
This PR is a patch to the issue #759 

Add the param `full=true` to the compartment API endpoint in swagger so that it retrieves all metabolites, genes, reactions and subsystems contained by the given compartment as claimed in the description.

**Type of change**  

**Testing**  

http://localhost/api/v2/#/Compartments

**Further comments**  

**Checklist**  

### DIFF
--- a/api/src/swagger/config.yaml
+++ b/api/src/swagger/config.yaml
@@ -58,7 +58,7 @@ paths:
           description: 'Successful query'
         '404':
           description: 'Invalid input'
-  /compartments/{id}?model={model}:
+  /compartments/{id}?model={model}&full=true:
     get:
       tags:
         - 'Compartments'


api/src/swagger/config.yaml
```diff
@@ -58,7 +58,7 @@ paths:
           description: 'Successful query'
         '404':
           description: 'Invalid input'
-  /compartments/{id}?model={model}:
+  /compartments/{id}?model={model}&full=true:
```
MalinAhlberg, line 61: Wouldn't it make sense to add `full` to the list of parameters (line 71) as well?

MalinAhlberg:
Sorry for not noticing this in the last PR, but what about this [comment](https://github.com/MetabolicAtlas/MetabolicAtlas/issues/759#issuecomment-966261151)?

nanjiangshu:
> Sorry for not noticing this in the last PR, but what about this [comment](https://github.com/MetabolicAtlas/MetabolicAtlas/issues/759#issuecomment-966261151)?

I think we've decided not to include svg-thumbnail in the API page since it is made solely for the HPA.

inghylt:
@MalinAhlberg I think we decided that it was HPA specific and not a functionality for the average user?

MalinAhlberg:
Ah thanks, good that you remembered @inghylt and @nanjiangshu ! :)

api/src/swagger/config.yaml
nanjiangshu: Not quite sure if we should add this as an additional parameter. And if it is added, the description text should also be changed.

There's another thing which confused me is how does this parameter `full` work. 
It seems both of the following two urls 
`/api/v2/compartments/golgi_apparatus?model=HumanGem&full=false`
`/api/v2/compartments/golgi_apparatus?model=HumanGem&full=true"`
return the same result. 
while 
`/api/v2/compartments/golgi_apparatus?model=HumanGem&full` and 
`/api/v2/compartments/golgi_apparatus?model=HumanGem`
return a different result.
Any ideas from @e0 , @inghylt, @mihai-sysbio and @MalinAhlberg

inghylt: Good catch @MalinAhlberg! I think it makes sense to add it as a parameter in the list of `/compartments/{id}?model={model}` parameters since it is a parameter we provide to get the result specified in the description. Given @nanjiangshu discovery though, I'm not really sure how we would do it since it seems like you need to omit it completely to get change in result, a bit silly instruction IMO. Either we fix the neo4j code in this PR or create a new issue to work with this, not sure what is optimal.

nanjiangshu:
During the standup at 2022-03-11 we decided to create a new issue to fix the bug at the API with the parameter `full`. This PR will be closed without merging.

PR #920, opened 2022-03-16
feat: UI and UX improvements for Table of Contents

### LINKED ISSUE #864: make the Table of Contents component sticky

As a user, I want the Table of Contents to stick to the top so that it does not disappear as I scroll down the web page.

Make sure that the Table of Contents (currently for the Documentation and About pages) is always visible even if the user scrolls to the bottom of a long page that surpasses the height of Table of Contents.

**Additional context**

Two things to keep in mind for implementation:
- The mobile layout is different (ToC at top, content at bottom) and it should not be sticky depending on how things are laid out.
- If the ToC is longer than the visible window height, it should be possible to reach the bottom of the ToC as well.
- For the active link, nested visualization should be added similarly to example below.

image

e0:
This PR closes #864.

**Type of change**  

- [x] Other
 
**List of changes made**  

Changes for the ToC are made according to the issue description, including:
- Sticks to the top after scrolling the page down.
- Has max visible height of screen height and gets its own scroll section if it's longer than the screen height.
- Does not become sticky for mobile (smaller than `$tablet`) devices.
- Adds nested visualization for active link. The active color before `white-ter` is now used for the active parent link. The actual active link color is now `link-light`, which seems to work well with the look and feel of the rest of the ToC, for example the icon color.

**Screenshot of the fix**  

image
image
image
image

**Testing**  

- Instructions on how to test

Try clicking the ToC links as well as scrolling the `/documentation` and `/about/*` pages.

**Further comments**  

- Please let me know if you have any suggestions on the design choices made in this PR.
- I have tested using different screen sizes on Chrome, Firefox, and Safari on macOS.

### DIFF
--- a/frontend/src/components/shared/TableOfContents.vue
+++ b/frontend/src/components/shared/TableOfContents.vue
@@ -6,7 +6,8 @@
         <li v-for="l in links" :key="l.name">
           <router-link
             :to="l.routeName ? { name: l.routeName } : l.link"
-            active-class="has-background-white-ter"
+            :class="{ 'has-background-white-ter': hasActiveSubsection(l) }"
+            active-class="has-background-link-light"
             @click.native="isMobileMenu = false"
           >
             <span v-if="l.icon" class="icon pr-5 has-text-info">
@@ -18,7 +19,7 @@
             <li v-for="sub in l.subsections" :key="sub.name">
               <router-link
                 :to="sub.routeName ? { name: sub.routeName } : sub.link"
-                active-class="has-background-white-ter"
+                active-class="has-background-link-light"
                 @click.native="isMobileMenu = false"
               >
                 {{ sub.name }}
@@ -37,6 +38,11 @@ export default {
     links: {
       type: Array,
       default: () => [],
+    },
+  },
+  methods: {
+    hasActiveSubsection({ subsections }) {
+      return (subsections ?? []).map(s => s.link).includes(this.$route.hash);
     },
   },
 };
--- a/frontend/src/components/shared/TableOfContents.vue
+++ b/frontend/src/components/shared/TableOfContents.vue
@@ -1,5 +1,5 @@
 <template>
-  <div class="column is-narrow">
+  <div id="table-of-contents" class="column is-narrow">
     <aside class="menu">
       <p class="menu-label">Table of Contents</p>
       <ul class="menu-list">
@@ -43,6 +43,16 @@ export default {
 </script>
 
 <style lang="scss" scoped>
+@media screen and (min-width: $tablet) {
+  #table-of-contents {
+    position: sticky;
+    top: 0;
+    align-self: flex-start;
+    max-height: 100vh;
+    overflow: auto;
+  }
+}
+
 .menu-list {
   ul {
     margin-top: 0;


inghylt:
Really nice! I've tried it out and messed with screen sizes on firefox and chrome, worked well. Minor thought, not sure how common/pleasant it is to scroll the left side index(ToC) as you need at `/documentation`. Maybe we could decrease the space between each sub item in the ToC to make it fit and remove the need for scrolling?

e0:
I’m open to decreasing the spacing but just to clarify it won’t guarantee not needing to scroll since that depends on the user’s window height. I think this technique is quite common in documentation pages and also in our map viewer sidebar. They do usually have clearer visual separation from the rest of the page though and I think there is room for improvement there as well. A new issue could be created if we want to pursue it.

frontend/src/components/shared/TableOfContents.vue
```diff
@@ -39,10 +40,25 @@ export default {
...
+  },
 };
 </script>
 
 <style lang="scss" scoped>
+@media screen and (min-width: $tablet) {
+  #table-of-contents {
+    position: sticky;
+    top: 0;
+    align-self: flex-start;
```
inghylt, line 56: Is it the Bulma class `container` in `./frontend/src/layouts/AboutLayout.vue` and `./frontend/src/components/Documentation.vue` that makes this div a child to a flexible container?

inghylt (approved):
Great job!

frontend/src/components/shared/TableOfContents.vue
e0: Good question! I think it's the `columns` class that adds the `display: flex;` style. Really nice that you spotted this as it points to an opportunity to further refactor these similar layouts.

MalinAhlberg (approved):
Looks really nice, great work!

mihai-sysbio:
Looks neat! I'm partly having second thoughts when I'm seeing that vertical scrollbar appear for the ToC, but I guess it will be a downside only for a small fraction of users, whereas it will be a improvement for most 👍🏻 

Regarding the use of the new class `has-background-link-light` - I think this is a new color that we haven't used anywhere on the website, so my immediate reactions are:
- is there any existing class that could be used?
- personally I find the blue too intense: the other colors in the palette defined in `vars.css` are more pastel, especially when it comes to a wider area (surface) taken by the visual element (a highlight is much larger than an icon)

e0:
> Looks neat! I'm partly having second thoughts when I'm seeing that vertical scrollbar appear for the ToC, but I guess it will be a downside only for a small fraction of users, whereas it will be a improvement for most 👍🏻

Yea it's not ideal with the scrollbar and it's a trade off we have to settle with for now. In the future we could consider improving how it looks as mentioned [here](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/920#issuecomment-1070736463).

___

> Regarding the use of the new class `has-background-link-light` - I think this is a new color that we haven't used anywhere on the website, so my immediate reactions are:
> 
> * is there any existing class that could be used?
> * personally I find the blue too intense: the other colors in the palette defined in `vars.css` are more pastel, especially when it comes to a wider area (surface) taken by the visual element (a highlight is much larger than an icon)

The choice of a blueish color is so that it should play well with our existing color for the ToC icons, which currently has `has-text-info`. In the `vars.scss` we have actually defined `$link: #00549e;` and `$info: $link;`. I'm open to suggestions if you have any colors to recommend. Not sure exactly which adjustments to make since I actually think this shade of blue is more pastelly than the other colors in `vars.scss`.

nanjiangshu (approved):
Really nice implementation! I tested on different Browsers with different dimensions and it all worked as expected. 
There might still be room for improvement for instance when only part of the ToC is shown (with a larger font), it is not obvious that there are more content existing in the ToC when the scroll bar is hidden. Nevertheless, I think the current solution is good enough for this issue.

e0:
> There might still be room for improvement for instance when only part of the ToC is shown (with a larger font), it is not obvious that there are more content existing in the ToC when the scroll bar is hidden. Nevertheless, I think the current solution is good enough for this issue.

That's a good point @nanjiangshu. I think we could also consider having the option to collapse and expand the menu.

PR #921, opened 2022-03-16
feat: add icons to the about page ToC

### LINKED ISSUE #876: add images/logos to About sections

As a user, I want pretty visual things to support the text so that get a better visual experience.

Adding logos or images where possible. For example, `/about/introduction `might have the Metabolic Atlas logo

**Additional context**

e0:
This PR closes #876.

**Type of change**  

- [x] Other
 
**List of changes made**  

Creating a draft PR for now to get some discussion going. There are two screenshots included below. The 2nd one has a fallback icon for links that do not include an icon.

A bit of background info:

We are currently using font awesome icons 4.7.0, which was released in 2016 and has 675 free icons. In contrast, font awesome 6 was just released earlier this month (March, 2022) and has 2007 free icons as well as 16068 pro icons.

Web best practices are a bit different compared to before as well. For example we are currently using a CDN for font awesome icons but this is [no longer the recommended way](https://fontawesome.com/docs/web/setup/use-kit#ready-to-get-kitted-up), at least for font awesome. Instead they are using what is called a Kit, which lets us create a bundle of just the icons that we need. The exact implementation and usage is not clear to me yet and requires a bit more research.

One benefit of using a Kit is that it allows us to upload our own logos. For example it was suggested in the issue to use the Metabolic Atlas logo for `Introduction`, which is currently not possible without a bit of code changes.

Outside of font awesome there is also other open source options such as [heroicons](https://heroicons.com/) and [Feather](https://feathericons.com/). There is also alternative tools like [fontello](https://fontello.com/) that let's you select, upload, and generate icon fonts in a similar way to how the new font awesome Kits work.

In the long run I think it would be a good to adopt a more modern way of using icons. But before doing any of that it would be great to discuss this with everyone and decide how to move forward together. Please let me know which of the following options (ranked by least amount of effort to most) you prefer.

1. Do nothing.
2. Merge what is in this PR, with no changes to how we use font awesome. In that case, any suggestions on icon changes is appreciated.
3. Start looking into a different way of using icons that allows us to include custom icons.

**Screenshot of the fix**  

image
image

**Testing**  

- Instructions on how to test

Look at the ToC for the `/about/*` pages.

**Further comments**  


inghylt:
- I think the addition of icons are really nice and that the ones you have chosen are suitable. However, icons are missing for Introduction and Elixir, how come?
- `For example, /about/introduction might have the Metabolic Atlas logo` with this I meant that the section Introduction can have the Metabolic Atlas logo as an image below the text to make it more visually attractive, I get that it's not obvious though!
- I think it makes sense to do things in a more modern way and have access to more icons. If that's not straightforward and requires time, a new issue should be recreated IMO
-

e0:
> * I think the addition of icons are really nice and that the ones you have chosen are suitable. However, icons are missing for Introduction and Elixir, how come?

I didn’t find any that felt like a good match. I could have probably spent more time looking at all the icons. We did discuss about the problem of whether or not to enforce icons for all root level links so I specifically wanted to illustrate the problem. I thought that we agreed on that icons should not be enforced if there are no good matches but maybe there was a misunderstanding?

> * `For example, /about/introduction might have the Metabolic Atlas logo` with this I meant that the section Introduction can have the Metabolic Atlas logo as an image below the text to make it more visually attractive, I get that it's not obvious though!

Ah I see, I definitely misunderstood that bit. For this example my intuition is that having a logo on the page so close to the actual logo in the navbar might look awkward. Having said that, it’s worth trying out. I probably won’t be working too much on the project for the rest of this week and if you or anyone else wants to continue on this branch please go ahead!

inghylt:
I don't have a memory of agreeing on that, but either way IMO all or none should have icons :)

e0:
Sorry about that! I was not on the same page when it came to what the “where possible” was referring to. Do you have any thoughts on the option of using fallback icons in the second screenshot?

inghylt:
I think the fallback icons in the second screenshot works! But as you mentioned in the text there are many more options in the newer versions of `font awesome` so I'm up for creating an issue for that. Also would be nice to have custom logos, e.g. for Metabolic Atlas and Elixir

MalinAhlberg:
Looks nice!
As for not enforcing icons, that did sound good to me when we discussed this, but when looking at it now I agree with @inghylt that it should be all or none.
I also think it sounds good update, so that we don't get outdated and so that we get access to more icons. It would be very nice to have custom icons, for example for Elixir! If it's not complicated or time consuming, I think it should be done in this PR. If it is time consuming, I vote for option 2, but using another icon for Elixir. Not sure which is best, but it looks a bit weird to use the fallbacks and have the same for Introduction and Elixir. Suggestions (of some rather bland icons that could work):
- [`cogs`](https://fontawesome.com/v4/icon/cogs) (as for resources under Documentation)
- [`circle`](https://fontawesome.com/v4/icon/circle)
- [`certificate`](https://fontawesome.com/v4/icon/certificate)
- [`asterisk`](https://fontawesome.com/v4/icon/asterisk)

mihai-sysbio:
Nice work with the icons! and thanks for that good intro to the reasons why we should switch away from how we are doing this atm with Font Awesome. There is a very old issue created for this https://github.com/MetabolicAtlas/private-issues/issues/73 (it was in Jira before).

My vote would be to try to stick to the initial issue description by merging  the work that has been done (with @MalinAhlberg's suggestion above). Adopting a new way to use icons together with a new icon pack is to me much more work that was assigned in this issue, and also a bit on the side of the sprint theme, so should have a separate issue for it.

e0:
Thank you for the review and feedback!

Really nice of you to suggest the icons with links @MalinAhlberg! I tested them and thought the `asterisk` one looks the best and have added a new commit for it. I agree with all or nothing and I left the fallback icon implementation just in case we add any pages/links in the future and forget to set an icon.

nanjiangshu:
These icons look really nice. I think we'd better use these simple icons instead of the actual logos. For example, in the page `/about/elixir` there are many logos listed and I feel it a bit weird to have the header ELIXIR also be paired with its own logo. 
My only concern is whether we should add the icon besides the header text. In the Documentation page we do, but for the About page maybe it looks a bit redundant?

MalinAhlberg (approved):
I think this looks really nice, so I approve

I agree with you @nanjiangshu that having the icons beside the header text would be a bit redundant, but I think it's good that you pointed this out. Anyway I'm fine with either.

> My only concern is whether we should add the icon besides the header text. In the Documentation page we do, but for the About page maybe it looks a bit redundant?

inghylt (approved):
Very nice!

nanjiangshu (approved):
LGTM!

PR #923, opened 2022-03-20
2.4

nanjiangshu:
In this minor release we have focused on improving the search functionality. The improvement has been achieved for both the Gem search and Global search by a better scoring of searching items and a more logical presentation of found results. More specifically, in the Gem search, the category with the highest scoring hit is ranked at the top; and in the Global search, the default tab is set as the category with the highest scoring hit. Moreover, we have resolved a number of small bugs that affected user experience. In addition, we have improved our development routine by adopting a [Definition of Done](https://en.wikipedia.org/wiki/Scrum_(software_development)#Definition_of_done_(DoD)), and an Acceptance criteria section has been added to the issue templates.

**Improving searching**
* #765
* [Improve GemSearch (quick search) layout](https://app.zenhub.com/workspaces/metatlas-sprint-607745622d49260019ce115a/issues/metabolicatlas/private-issues/151)
* #705 
* [search pagination - research issue - decided to be implemented in the future](https://app.zenhub.com/workspaces/metatlas-sprint-607745622d49260019ce115a/issues/metabolicatlas/private-issues/101)
* #884 
* #710 
* #744 
* #686 
* #764
* #707 
* #771 
* #729 

**Bug fixing** 
* #760 
* #759 
* #893 
* #819 
* #775 
  
**Documentation**
* #894 

**Misc**
* #790 
* #844 
* #857

PR #928, opened 2022-03-22
chore: minor changes to the news 2.4 

nanjiangshu:
This PR adds minor changes to the news for the release 2.4 in responding to [this commit](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/923#issuecomment-1075245589)

I was trying to merge without opening a new PR but then realized it is forbidden.

### DIFF
--- a/frontend/src/content/news.json
+++ b/frontend/src/content/news.json
@@ -3,7 +3,7 @@
     {
       "date": "2022-03-21",
       "title": "Metabolic Atlas 2.4 improves search functionality",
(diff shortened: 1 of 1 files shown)


inghylt (approved):

MalinAhlberg (approved):
Nice!

A side note: [semantic line breaks](https://sembr.org/) were lifted in the dev channel the other day, I think it would make much sense for these types of texts.

nanjiangshu:
> Nice!
> 
> A side note: [semantic line breaks](https://sembr.org/) were lifted in the dev channel the other day, I think it would make much sense for these types of texts.

Good point @MalinAhlberg , we can discuss about that.

PR #932, opened 2022-03-23
fix: complete meta descriptions

### LINKED ISSUE #754: only one compartment listed in reaction description
**Describe the bug**
The meta information for reactions only contains one compartment even if the reaction belong to multiple. 

**To reproduce**
1. Visit for example `/explore/Human-GEM/gem-browser/reaction/MAR08363`
2. Inspect the page
3. Search for "description"
4. Note that Cytosol is not listed as a compartment in the content of `<meta data-vue-meta>`:
`<meta data-vue-meta="1" property="og:description" data-vmid="og:description" content="The reaction MAR08363 in Human-GEM (version 1.8.0) can be found in the Nucleus compartment and the Transport reactions subsystem.">`

**Expected behavior**
The description should be containing all compartments, e.g. "The reaction MAR08363 in Human-GEM (version 1.8.0) can be found in the Cytosol and Nucleus compartments and the Transport reactions subsystem."

- [ ] make sure this works well for both compartments and subsystems

**Additional context**
The same case for reaction at `/explore/Human-GEM/gem-browser/reaction/MAR12870`.

e0:
This PR closes #754.

**Type of change**  

**List of changes made**  

For reactions, metabolites, and genes, full names of compartments and subsystems are now included in the meta description.
For subsystems, full names of compartments are now included in the meta description.

**Screenshot of the fix**  

image
image
image
image

**Testing**  

- Instructions on how to test

Please look at the meta description for different reactions, metabolites, genes, and subsystems. The ones included in the screenshots are:

- `/explore/Human-GEM/gem-browser/reaction/MAR06245`

- `/explore/Human-GEM/gem-browser/metabolite/MAM02724m`

- `/explore/Human-GEM/gem-browser/gene/ENSG00000074416`

- `/explore/Human-GEM/gem-browser/subsystem/beta_alanine_metabolism`   

**Further comments**  

When there are two compartments or subsystems, they are joined with ` and `, otherwise they are joined with `, `. Please let me know if you have any alternative suggestions.

### DIFF
--- a/frontend/src/components/explorer/gemBrowser/Subsystem.vue
+++ b/frontend/src/components/explorer/gemBrowser/Subsystem.vue
@@ -116,7 +116,7 @@ export default {
     });
 
     const title = `${this.info.name}, Subsystem in ${this.model.short_name}`;
-    const description = `The subsystem ${this.info.name} in ${this.model.short_name} (version ${this.model.version}) can be found in the ${compartments} ${compartmentLabel}, and contains ${this.metabolites.length} metabolites and ${this.genes.length} genes.`;
+    const description = `The subsystem ${this.info.name} in ${this.model.short_name} (version ${this.model.version}) can be found in the ${compartmentLabel} ${compartments}; and contains ${this.metabolites.length} metabolites and ${this.genes.length} genes.`;
 
     return {
       title,
--- a/frontend/src/components/explorer/gemBrowser/Subsystem.vue
+++ b/frontend/src/components/explorer/gemBrowser/Subsystem.vue
@@ -83,7 +83,12 @@
 <script>
 import { mapGetters, mapState } from 'vuex';
 import ComponentLayout from '@/layouts/explorer/gemBrowser/ComponentLayout';
-import { buildCustomLink, generateSocialMetaTags, reformatTableKey } from '@/helpers/utils';
+import {
+  buildCustomLink,
+  generateSocialMetaTags,
+  reformatTableKey,
+  combineWords,
+} from '@/helpers/utils';
 
 export default {
   name: 'Subsystem',
@@ -105,8 +110,13 @@ export default {
       return {};
     }
 
+    const [compartments, compartmentLabel] = combineWords({
+      items: this.info.compartments.map(c => c.name),
+      itemType: 'compartment',
+    });
+
     const title = `${this.info.name}, Subsystem in ${this.model.short_name}`;
-    const description = `The subsystem ${this.info.name} in ${this.model.short_name} (version ${this.model.version}) can be found in ${this.info.compartments.length} compartments, and contains ${this.metabolites.length} metabolites and ${this.genes.length} genes.`;
+    const description = `The subsystem ${this.info.name} in ${this.model.short_name} (version ${this.model.version}) can be found in the ${compartments} ${compartmentLabel}, and contains ${this.metabolites.length} metabolites and ${this.genes.length} genes.`;
 
     return {
       title,
--- a/frontend/src/components/explorer/gemBrowser/Gene.vue
+++ b/frontend/src/components/explorer/gemBrowser/Gene.vue
@@ -83,7 +83,7 @@ export default {
     });
 
     const title = `${this.gene.geneName}, Gene in ${this.model.short_name}`;
-    const description = `The gene ${this.gene.geneName} in ${this.model.short_name} (version ${this.model.version}) can be found in the  ${compartments} ${compartmentLabel} and the ${subsystems} ${subsystemLabel}.`;
+    const description = `The gene ${this.gene.geneName} in ${this.model.short_name} (version ${this.model.version}) can be found in the ${compartmentLabel} ${compartments}; and the ${subsystemLabel} ${subsystems}.`;
 
     return {
       title,
--- a/frontend/src/components/explorer/gemBrowser/Gene.vue
+++ b/frontend/src/components/explorer/gemBrowser/Gene.vue
@@ -37,7 +37,7 @@
 <script>
 import { mapGetters, mapState } from 'vuex';
 import ComponentLayout from '@/layouts/explorer/gemBrowser/ComponentLayout';
-import { generateSocialMetaTags, reformatTableKey } from '@/helpers/utils';
+import { generateSocialMetaTags, reformatTableKey, combineWords } from '@/helpers/utils';
 
 export default {
   name: 'Gene',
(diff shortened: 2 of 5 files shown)


frontend/src/components/explorer/gemBrowser/Subsystem.vue
```diff
@@ -105,8 +110,13 @@ export default {
...
     }
 
+    const [compartments, compartmentLabel] = combineWords({
+      items: this.info.compartments.map(c => c.name),
+      itemType: 'compartment',
+    });
+
     const title = `${this.info.name}, Subsystem in ${this.model.short_name}`;
-    const description = `The subsystem ${this.info.name} in ${this.model.short_name} (version ${this.model.version}) can be found in ${this.info.compartments.length} compartments, and contains ${this.metabolites.length} metabolites and ${this.genes.length} genes.`;
+    const description = `The subsystem ${this.info.name} in ${this.model.short_name} (version ${this.model.version}) can be found in the ${compartments} ${compartmentLabel}, and contains ${this.metabolites.length} metabolites and ${this.genes.length} genes.`;
```
inghylt, line 119: We are only listing the amount of metabolites and genes in the meta data but at the interface, e.g. at `/explore/Human-GEM/gem-browser/subsystem/beta_alanine_metabolism`, you can see that there are also 17 reactions. Shouldn't we show that information as well?

inghylt (approved):
Really nice work! I approve of this since I think it is production ready. To me, it would be ok to fix [this comment](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/932/files#r834162595) in this PR, but maybe it makes more sense to create another issue for it.

Also, I wonder if one should add an between the last two compartments/subsystems if the are multiple. For example, `the Pool reactions, Acylglycerides metabolism, Transport reactions subsystems."` would be `the Pool reactions, Acylglycerides metabolism and Transport reactions subsystems."` Don't have a strong preference for this though and not even sure it is grammatically preferable but at least wanted to raise the thought

frontend/src/components/explorer/gemBrowser/Subsystem.vue
e0: We could list these too and I could add a commit if we want this info. I was worried that it would generate descriptions that are too long since some subsystems have many more components. For reactions specifically we rely on a potentially slow request and only return max 1000 so it could be incomplete.

e0:
> Also, I wonder if one should add an between the last two compartments/subsystems if the are multiple. For example, `the Pool reactions, Acylglycerides metabolism, Transport reactions subsystems."` would be `the Pool reactions, Acylglycerides metabolism and Transport reactions subsystems."` Don't have a strong preference for this though and not even sure it is grammatically preferable but at least wanted to raise the thought

I was pondering about this too! That would usually be my preference too. I think the grammatically correct version would be to have a comma before the last “and”. The reason I left it out was because we also use “and” to join the compartments part of the sentence with the subsystems part of the sentence and it feels kind of wrong to have the “, and” in front of the last compartment and an “and” before starting subsystems. We could also separate the two parts into their own sentences. Although I do like this format, especially for when there are fewer components.

frontend/src/components/explorer/gemBrowser/Subsystem.vue
nanjiangshu: Since these meta data is not supposed to be used by actual users, I think it makes sense to show only the count for genes, reactions and metabolites since otherwise the text can be rather tedious. 

However, for the compartments, I guess we can show all of them since there are not so many.

inghylt:
@e0 thanks for providing your reasoning! I agree with you, let's keep it as it is

nanjiangshu:
Really nice work @e0,  but I have a similar feeling as @inghylt regarding the grammar and the consistence of expression.  
1. The sentence is started with "the subsystems A..., the reaction A...", but then in the second part of the sentence, the expression is changed to `can be found in A, B and C subsystems`. that is, the order of `${subsystems}` and `${subsystemLabel}` is switched. I think it is better to be consistent. 
In addition, if you google it, the expression `the subsystems A, B and C` (9910 hits) is much more common than the expression `the A, B and C subsystems` (3 hits). The same is for compartments and metabolites. 

3. For the grammatical problem. I think for the one type, people usually expressed things like this 
* For the single item: `the subsystem A`
* For two items : `the subsystem A and B`
* For more than two: `the subsystem A, B and C`, `the subsystem A, B, C and D`

4. When there are multiple types, it is actually slightly more complicated for how comma is used with the word `and`. For example
* If both types have just one item, no comma is needed: `the subsystem A and the compartment M`
* If the first type has two items, a comma is needed before the second type: `the subsystems A and B, and the compartment M`
However, I guess it does not worth the time to do this kind of programming. So my suggestion would be to add comma in both cases, that is 
* `the subsystem A, and the compartment M`
* `the subsystems A and B, and the compartment M`

e0:
> 1. The sentence is started with "the subsystems A..., the reaction A...", but then in the second part of the sentence, the expression is changed to `can be found in A, B and C subsystems`. that is, the order of `${subsystems}` and `${subsystemLabel}` is switched. I think it is better to be consistent.
>    In addition, if you google it, the expression `the subsystems A, B and C` (9910 hits) is much more common than the expression `the A, B and C subsystems` (3 hits). The same is for compartments and metabolites.

I like this suggestion, especially for when there is two or more components. When there is one I think `the A subsystem` sounds better than `the subsystem A`, but it's a good tradeoff. I can make this change if everyone is on board.

___

> 2. For the grammatical problem. I think for the one type, people usually expressed things like this
> 
> * For the single item: `the subsystem A`
> * For two items : `the subsystem A and B`
> * For more than two: `the subsystem A, B and C`, `the subsystem A, B, C and D`

As I mentioned above, some advocate that the comma should be added to be in front of the `and`, so it would be `the subsystems A, B, and C`. I think the biggest advantage of this is that if the item itself has the word `and` in it. For example: `My favorite cartoons are The Simpsons, Family Guy, and Tom and Jerry.` In this case it is more clear that `Tom and Jerry` is a single item.  Reference: https://www.grammarflip.com/curriculum/commas-items-in-a-series

> 4. When there are multiple types, it is actually slightly more complicated for how comma is used with the word `and`. For example
> 
> * If both types have just one item, no comma is needed: `the subsystem A and the compartment M`
> * If the first type has two items, a comma is needed before the second type: `the subsystems A and B, and the compartment M`
>   However, I guess it does not worth the time to do this kind of programming. So my suggestion would be to add comma in both cases, that is
> * `the subsystem A, and the compartment M`
> * `the subsystems A and B, and the compartment M`

I like this clearer separation, maybe a `;` symbol would be even more appropriate here to clearly distinguish the two parts and stand out from the `,` and `and` instances.

___

Overall, I'm fine with any of these changes but also ok with leaving things as is. As long as we are on the same page I'm happy to make the changes.

e0:
Thanks for the nice meeting now @inghylt and @nanjiangshu! The conclusion is that since it should not take much time to implement, we should make the changes suggested by @nanjiangshu, including using a `;` to separate the compartments part and the subsystems part of the sentence.

e0:
I have added [a commit](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/932/commits/8c31f5222e58909c1a9cb651b3ec74225de0c9df) that improves the sentence structure as we agreed now. Please have another review when you have time. @inghylt @MalinAhlberg @nanjiangshu

MalinAhlberg (approved):
Nice!

nanjiangshu (approved):
Looks great!

inghylt:
LGTM!

Issue #896, opened 2022-03-07
Improve scoring


MalinAhlberg:
**Please describe the feature**

As a user I want the scoring from neo4j to be improved so that I'm always presented with the most relevant search result.

**Describe the solution you would like**

- When searching for `3(R)-hydroxy-(2S,6R,10)-trimethyl-hendecanoyl-CoA` you should get `MAM00685p` as the top result (currently gets score 0).
- ~When searching for `fatty` the top result should not be the gene `FASTK` as top result.~ Moved to new issue #978.

**Acceptance Criteria**
- [ ] Add tests related to search, covering all compartments and more than 1 model

**Additional context**

Related:
https://github.com/MetabolicAtlas/MetabolicAtlas/pull/886
https://github.com/MetabolicAtlas/MetabolicAtlas/pull/887

inghylt:
I've done some research this morning and I now know why this still is an issue despite #937

If you search for example `3(R)-hydroxy-(2S,6R,10)-trimethyl-hendecanoyl`, `MAM00685p` will not be part of the result returned at [this line](https://github.com/MetabolicAtlas/MetabolicAtlas/blob/1e665d2373d59ec92038bc324494608cdeba07b3/api/src/neo4j/queries/search/modelSearch.js#L202), but the object

`{
     score: 14.249983787536621,
     id: '3_r_hydroxy_2s_6r_10_trimethyl_hendecanoyl_coa',
     labels: [ 'Metabolite', 'HumanGem' ]
 },`
 
while the obj used in [this row](https://github.com/MetabolicAtlas/MetabolicAtlas/blob/1e665d2373d59ec92038bc324494608cdeba07b3/api/src/neo4j/queries/search/modelSearch.js#L259) will look like

`{
   name: '3(R)-hydroxy-(2S,6R,10)-trimethyl-hendecanoyl-CoA',
   compartment: 'Peroxisome',
   id: 'MAM00685p'
}`

[...]

PR #937, opened 2022-03-24
fix: model search bug

### LINKED ISSUE #936: Can't search by metabolite id
- When searching for metabolite id MAM02040c in the search bar of metabolic atlas with "Human-GEM" selected at the left gives no results, but it works fine when using reaction ids or the name of the metabolite (h2o).
Capture_a
Capture_b
Capture_c
Capture_d

- Also, the compartment 's' in the metabolic atlas seems to be different from the external compartment in models, which is 'e'

inghylt:
This PR closes #936 and was written by @e0 and @inghylt 

**Type of change**  

- [x] Other, refactoring to make to code more readable and thus less bug prone 
 
**List of changes made**  

- Resolved bug related to UNWIND with empty list. These changes are also made in global search
- Made sure that it is not the lowest scores that is saved if multiple nodes with the same id are in the fulltext search results + refactored the related code to make it more readable to avoid future bugs. These changes are also made in global search.
- Removed redundant code from the search store. Some of it was actually redundant before the changes introduced by this PR, but the main thing is that the API now always returns a score, i.e. there should be no `undefined` cases.

**Screenshot of the fix**  

image

**Testing**  

Search for `MAM02040c` but also other component ids and models

**Further comments**  

An unexpected discovery was made during the bug fixing. Apparently, there are sometimes duplicate components returned from the fulltext search, and the one that appears latest, i.e. the one with the lowest score, will be the one "stored". This is probably the reason behind the situations we have seen in the past with weird scores. 

### DIFF
--- a/api/src/neo4j/queries/search/globalSearch.js
+++ b/api/src/neo4j/queries/search/globalSearch.js
@@ -18,14 +18,22 @@ const fetchCompartmentalizedMetabolites = async ({
   if (viaMetabolties) {
     statement += `
 WITH ${JSON.stringify(ids)} as mids
-UNWIND mids as mid
+UNWIND
+  CASE
+      WHEN mids = [] THEN [null]
+      ELSE mids
+  END AS mid
 MATCH (:Metabolite:${model} {id:mid})-[${version}]-(cm:CompartmentalizedMetabolite)
 WITH DISTINCT(cm.id) as cmid
 `;
   } else {
     statement += `
 WITH ${JSON.stringify(ids)} as cmids
-UNWIND cmids as cmid
+UNWIND
+  CASE
+      WHEN cmids = [] THEN [null]
+      ELSE cmids
+  END AS cmid
 `;
   }
 
--- a/api/src/neo4j/queries/search/globalSearch.js
+++ b/api/src/neo4j/queries/search/globalSearch.js
@@ -313,21 +313,21 @@ LIMIT ${limit}
 
   const results = await queryListResult(statement);
 
-  const idsToScore = {};
-  const uniqueIds = results.reduce((o, r) => {
-    const c = intersect(COMPONENT_TYPES, r.labels);
-    if (!o[c]) {
-      o[c] = new Set();
+  const uniqueIds = {};
+  for (let [pos, node] of Object.entries(results)) {
+    if (!(node['id'] in uniqueIds)) {
+      uniqueIds[node['id']] = node;
     }
-    o[c].add(r.id);
-    idsToScore[r.id] = r.score;
-    return o;
-  }, {});
+  }
+  const groupedByComponents = {};
+  for (let [id, properties] of Object.entries(uniqueIds)) {
+    const c = intersect(COMPONENT_TYPES, properties.labels);
+    if (!groupedByComponents[c]) {
+      groupedByComponents[c] = [];
+    }
+    groupedByComponents[c].push(id);
+  }
 
-  const ids = Object.assign(
-    {},
-    ...Object.keys(uniqueIds).map(c => ({ [c]: Array.from(uniqueIds[c]) }))
-  );
   const [
     compartmentalizedMetabolites,
     metabolites,
@@ -337,33 +337,33 @@ LIMIT ${limit}
     compartments,
   ] = await Promise.all([
     fetchCompartmentalizedMetabolites({
-      ids: ids['CompartmentalizedMetabolite'],
+      ids: groupedByComponents['CompartmentalizedMetabolite'],
       model,
       version: v,
       limit,
     }),
     fetchCompartmentalizedMetabolites({
-      ids: ids['Metabolite'],
+      ids: groupedByComponents['Metabolite'],
       model,
       version: v,
       limit,
       viaMetabolties: true,
     }),
-    fetchGenes({ ids: ids['Gene'], model, version: v }),
+    fetchGenes({ ids: groupedByComponents['Gene'], model, version: v }),
     fetchReactions({
-      ids: ids['Reaction'],
+      ids: groupedByComponents['Reaction'],
       model,
       version: v,
       includeMetabolites: !!limit,
     }),
     fetchSubsystems({
-      ids: ids['Subsystem'],
+      ids: groupedByComponents['Subsystem'],
       model,
       version: v,
       includeCounts: true,
     }),
     fetchCompartments({
-      ids: ids['Compartment'],
+      ids: groupedByComponents['Compartment'],
       model,
       version: v,
       includeCounts: true,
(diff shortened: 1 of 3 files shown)


e0 (approved):
LGTM!

mihai-sysbio (approved):
That was quick @inghylt, nice! Looks good as far as I can tell.
- Did you notice any difference in performance?
- Does this PR also solve #896? edit: Not quite, from what I testef.

nanjiangshu (approved):
Great fix! I haven't experienced any performance downgrade although no benchmark has been done.

inghylt:
```
const idsToScore = {};
  const uniqueIds = results.reduce((o, r) => {
    const c = intersect(COMPONENT_TYPES, r.labels);
    if (!o[c]) {
      o[c] = new Set();
    }
    o[c].add(r.id);
    idsToScore[r.id] = r.score;
    // Sometimes there are duplicate components with different score, and the lower score
    // will then override the higher score without this if statement
    if (!idsToScore[r.id]) {
      o[c].add(r.id);
      idsToScore[r.id] = r.score;
    }
    return o;
  }, {});
```

I feel like this code can be rewritten. It's kind of messy and we are using Set but still have to make a check if r.id is present. Should I create a new issue for it? Or try to make it better in this PR?

mihai-sysbio:
> Should I create a new issue for it? Or try to make it better in this PR?

This branch has been deployed so there is no time-pressure anymore, if you feel like it makes sense to continue in this PR.

inghylt:
Pushed the changes I sat with a bit last week. I think the code was a bit messy and hard to understand (thus more bug prone) before, hope you guys think this is an improvement as well. @MalinAhlberg I realized now as well why some scores was undefined in #887, see the commit message dc44435. If you guys approve these changes, we could create an issue (or fix here directly) to remove the check at https://github.com/MetabolicAtlas/MetabolicAtlas/blob/f780f4b597bbf01c1721577c94a26802152fdb17/frontend/src/store/modules/search.js#L25

inghylt:
Just noticed now, the bugs fixed in this PR is still present in global search. Should we fix global search as well in this PR?

e0 (approved):
Good job on the new refactoring! I think it's more appropriate that we are using `for` loops now instead of using `reduce` and it's more readable.

Regarding global search, the bug is not as severe since the result is still included but ranked lower. I feel a bit conflicted about whether to fix it now or not. We mentioned before that we would like to tackle the global search pagination/redesign issue _soon_, which means that in the mean time it could require more time whenever we are fixing anything related to the search logic. What do you think?

MalinAhlberg:
> Pushed the changes I sat with a bit last week. I think the code was a bit messy and hard to understand (thus more bug prone) before, hope you guys think this is an improvement as well. @MalinAhlberg I realized now as well why some scores was undefined in #887, see the commit message [dc44435](https://github.com/MetabolicAtlas/MetabolicAtlas/commit/dc4443508adc676276322191d15b16e5a11bf565). If you guys approve these changes, we could create an issue (or fix here directly) to remove this check
> 

Nice, the `undefined`s actually seem to be gone :v:. I really like the refactoring too!
I vote for removing the line you mentioned above, and fixing the issue in global search before we re-test and merge this.

nanjiangshu (approved):
Really nice fix and refactoring. I feel the new code much easier to understand and better in dealing with duplicated IDs in the result result. 

1. For the global search, I agree with @e0 that the bug is less critical and I would vote for being solved in a separate issue. Mabye together with the major re-writing of the global search function. 
2. For checking `undefined`, it think it's more logical to define it in the API than to check it in the frontend. Strictly speaking, this line 
```javascript
score: obj.id in uniqueIds ? uniqueIds[obj.id]['score'] : 0,
```
may still result in an undefined score if `uniqueIds[obj.id]['score']` is undefined. But I guess this won't happen, otherwise it will be a bug with neo4j 😃

api/src/neo4j/queries/search/globalSearch.js
```diff
@@ -18,14 +18,22 @@ const fetchCompartmentalizedMetabolites = async ({
...
 `;
   } else {
     statement += `
 WITH ${JSON.stringify(ids)} as cmids
-UNWIND cmids as cmid
+UNWIND
+  CASE
+      WHEN cmids = [] THEN [null]
+      ELSE cmids
+  END AS cmid
```
e0, line 36: Did this cause any problems specifically for the global search? Just curious since it seemed like it wasn't needed for the model search.

inghylt:
(Hopefully) Final re-request! I have now removed the unnecessary if undefined statement and applied the fixes to global search as well. See the updated PR description and commits for more information.

Btw @nanjiangshu I tried out the examples in [this comment](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/938#pullrequestreview-925858090) and they all seem to work well

frontend/src/store/modules/search.js
```diff
@@ -21,16 +21,11 @@ const categorizeResults = results => {
...
-              d.score = 0;
-            }
             if (e.score > categoryScore || !categoryScore) {
               categoryScore = e.score;
             }
-            categoryScore = e.score > categoryScore ? e.score : categoryScore;
-            d.model = { id: model, name: resultsModel.name };
-            return d;
+            e.model = { id: model, name: resultsModel.name };
+            return e;
```
e0, line 28: While we do not have [this eslint rule](https://eslint.org/docs/rules/no-param-reassign) in place, it is considered good practice to not mutate function params in JavaScript. In this case I think the following should amount to the same value without mutating the function param.

```suggestion
            return { ...e, model: { id: model, name: resultsModel.name } };
```

e0 (approved):
Looks great! Nice that you took time to fix the global search as well. I left a couple of minor comments but they do not necessarily need to be addressed imo so approving this now.

api/src/neo4j/queries/search/globalSearch.js
inghylt: There was no error what I noticed in global search, but it felt safer to add this if (c)mids would be empty for some query.

inghylt: I don't think we tested anything in global search? Either way, I also agree it makes sense to keep it as is for now

e0: You are right! I assumed that it would work the same way.

MalinAhlberg (approved):
Good work!

nanjiangshu:
> (Hopefully) Final re-request! I have now removed the unnecessary if undefined statement and applied the fixes to global search as well. See the updated PR description and commits for more information.
> 
> Btw @nanjiangshu I tried out the examples in [this comment](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/938#pullrequestreview-925858090) and they all seem to work well

Yes. That comment is resolved. See the slack discussion at https://nbisweden.slack.com/archives/C01D13EKUV7/p1648642859897109

nanjiangshu (approved):
LGTM!

PR #938, opened 2022-03-28
fix: use parent node of pubmedreference

### LINKED ISSUE #889: Searching for PMID does not return relevant results
**Describe the bug**

Searching for PMIDs does not return relevant results. This affects both the GemSearch and GlobalSearch.

**Steps to reproduce**

e.g. `11579996`,  `16111821`,  `18406340`, `11368003`

**Expected behavior**

Searching of PMIDs return relevant reactions.

**Screenshots**

**Additional context**

This should work for both searches (GEMSearch and global).

MalinAhlberg:
This PR closes #889.

**Type of change**  

**List of changes made**  

Adapt case statement in neo4j query to use the parent node when encountering a PubmedReference node.

**Screenshot of the fix**  

pmid

**Testing**  

Try searching for PMIDs, eg `11579996`, `16111821`, `18406340`, `11368003`. The result should contain Reactions, where the PMIDs can be seen in th References table at the bottom of the page.

Also please verify that the results for other types of queries haven't changed.

### DIFF
--- a/api/src/neo4j/queries/search/globalSearch.js
+++ b/api/src/neo4j/queries/search/globalSearch.js
@@ -279,6 +279,7 @@ const search = async ({ searchTerm, model, version, limit, includeCounts }) => {
   const v = version ? `:V${version}` : '';
 
   const term = sanitizeSearchString(searchTerm, true);
+  const childLabels = "['ExternalDb', 'PubmedReference']"
 
   // The search term is used twice, once with exact match and once with
   // fuzzy match. This seems to produce optimal results.
@@ -290,7 +291,7 @@ OPTIONAL MATCH (node)-[${v}]-(parentNode:${model})
 WHERE node:${model} OR parentNode:${model}
 WITH DISTINCT(
 	CASE
-		WHEN EXISTS(node.id) AND NOT EXISTS(node.externalId) AND NOT 'PubmedReference' in LABELS(node)
+                WHEN EXISTS(node.id) AND NOT apoc.coll.intersection(${childLabels}, LABELS(node))
                 THEN { id: node.id, labels: labelList, score: score }
 		ELSE { id: parentNode.id, labels: LABELS(parentNode), score: score }
 	END
--- a/api/src/neo4j/queries/search/globalSearch.js
+++ b/api/src/neo4j/queries/search/globalSearch.js
@@ -1,6 +1,10 @@
 import queryListResult from 'neo4j/queryHandlers/list';
 import { sanitizeSearchString, intersect } from 'utils/utils';
-import { MODELS, COMPONENT_TYPES } from 'neo4j/queries/search/helper';
+import {
+  MODELS,
+  COMPONENT_TYPES,
+  CHILD_LABELS,
+} from 'neo4j/queries/search/helper';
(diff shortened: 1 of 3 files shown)


e0:
This works great @MalinAhlberg! Nice job using `labels` to filter out the nodes. I was thinking that we should have probably used this syntax initially for `ExternalDb` as well. So then I was wondering if the following would be easier to understand and maintain in the future.

```sql
WHEN EXISTS(node.id) AND NOT apoc.coll.intersection( ['ExternalDb', 'PubmedReference'], LABELS(node))
```

I think it's a bit clearer that it is _"all nodes that have an `id` but not belonging to any of these following types"_. What do you think?

MalinAhlberg:
> I think it's a bit clearer that it is _"all nodes that have an `id` but not belonging to any of these following types"_. What do you think?

Good idea @e0! I agree that this makes it more readable. I made a commit with this,  cc58b27. I also moved out the labels to a separate variable, as I think this makes it a bit easier to follow and maintain. Please let me know what you think, further feedback and ideas are welcome :)

e0:
Great that you added a variable @MalinAhlberg! I think it looks good. There is some room for further refactoring, for example defining a single instance of `childLabels` but there is good reason to refactor this entire bit of cypher instructions as well so I think it makes sense to wait for later. I will just test a bit more with the searches and then I think I can approve.

e0 (approved):

inghylt:
Nice work!

I think it makes sense to move `childLabels` to `neo4j/queries/search/helper` for DRY purposes, even if we do refactor this in the future.

MalinAhlberg:
> I think it makes sense to move childLabels to neo4j/queries/search/helper for DRY purposes, even if we do refactor this in the future.

Fixed in 7d0312a.

nanjiangshu (commented):
Nice fix @MalinAhlberg. Cool to use the label intersection in searching. However, when I tried to verify if the results for other search terms, I got different results.

For example, for the BiGG ID `ASNTYRTHRt`, the result is different compared to the previous version. Hits for reactions are slightly different but hits for metabolites are gone. The number of hits in the Global search seems to be the same, but I haven't checked if they are all identical. 

Screenshot 2022-03-30 at 11 39 43
Screenshot 2022-03-30 at 11 39 56

When I searched for the gene name `ring finger protein 40`. The results are again differrent
Screenshot 2022-03-30 at 11 45 09
Screenshot 2022-03-30 at 11 45 17
Hits for metabolites are gone while the rest seems to be the same. 
For the global search, the number of reactions is different. 

There is another problem I found which is not related to this issue. When searching for the term `ASNTYRTHRt`, the top ranked hits in the GemSearch is reactions, however, when clicking the global search, the default tab is `Metabolites`. I suppose it should be `Reactions` if the scores of the GemSearch and GlobalSearch are the same.

MalinAhlberg:
Thanks for the detailed tests, @nanjiangshu ! As discussed at stand-up today, the cases you found are related to PR #937 . I think your second point also might be related to that, since the metabolites were missing from the model search result?
Anyhow, it's very nice to know that this PR is well tested now :raised_hands:

PR #939, opened 2022-03-30
fix: make sure 404 is triggered when compartment or subsystem is not found

### LINKED ISSUE #800: missing error messages for invalid compartments and subsystems
**Describe the bug**
Navigating to subsystems or compartments with bogus IDs does not produce any error message.

**Steps to reproduce**
Here's a random compartment ID: `/explore/Yeast-GEM/gem-browser/compartment/zdfgdfsg`. The behaviour is the same for `/explore/Yeast-GEM/gem-browser/subsystem/scdfgdfgabolism`.

**Expected behavior**
The same error message as for the other GEM Browser pages (gene/metabolite/reaction)

**Screenshots**
image

versus the expected

image

**Additional context**

e0:
This PR closes #800.

**Type of change**  

**List of changes made**  

We have [these lines](https://github.com/MetabolicAtlas/MetabolicAtlas/blob/2d061fd4ff58e093a8748829ecdaccc8452ddae2/api/src/neo4j/queryHandlers/single.js#L20-L25) that trigger a `404` for whenever an object is empty. The problem before was that object at this point is not empty according to our definition because it has the following shape. Specifically due to the `"compartmentId": "typo"` part.

```json
{
  "metabolitesCount": 0,
  "reactionsCount": 0,
  "genesCount": 0,
  "compartmentSVGs": [
    {
      "compartmentId": "typo",
      "compartmentSVGs": []
    }
  ],
  "subsystems": [],
  "externalDbs": []
}
```

This PR fixes the cypher queries so that the object would have the following shape instead, which fits our definition of an empty object.

```json
{
  "metabolitesCount": 0,
  "reactionsCount": 0,
  "genesCount": 0,
  "compartmentSVGs": [],
  "subsystems": [],
  "externalDbs": []
}
```

**Screenshot of the fix**  

image
image

**Testing**  

1. Make sure the appropriate error message triggers when an incorrect id is provided for example: 
  - `/explore/Human-GEM/gem-browser/compartment/typo`
  - `/explore/Human-GEM/gem-browser/subsystem/typo`
2. Make sure the page loads as normal when a correct id is provided for example: 
  - `/explore/Human-GEM/gem-browser/compartment/cytosol`
  - `/explore/Human-GEM/gem-browser/subsystem/acylglycerides_metabolism`

### DIFF
--- a/api/src/neo4j/queries/compartment.js
+++ b/api/src/neo4j/queries/compartment.js
@@ -39,7 +39,12 @@ CALL apoc.cypher.run("
   UNION
   
   MATCH (:Compartment${m} {id: '${id}'})-[${v}]-(csvg:SvgMap)
-  WITH {compartmentId: '${id}', compartmentSVGs: COLLECT(DISTINCT(csvg {.*}))} as compartmentSVG
+  WITH  COLLECT(DISTINCT(csvg {.*})) as csvgs
+  WITH CASE
+    WHEN size(csvgs) > 0
+    THEN {compartmentId: '${id}', compartmentSVGs: csvgs}
+    ELSE null
+  END as compartmentSVG
   RETURN { compartmentSVGs: COLLECT(compartmentSVG) } as data
 ", {}) yield value
 RETURN apoc.map.mergeList(COLLECT(value.data)) as compartment
--- a/api/src/neo4j/queries/subsystem.js
+++ b/api/src/neo4j/queries/subsystem.js
@@ -41,7 +41,12 @@ CALL apoc.cypher.run("
   UNION
   
   MATCH (:Subsystem${m} {id: '${id}'})-[${v}]-(ssvg:SvgMap)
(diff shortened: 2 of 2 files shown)


MalinAhlberg (approved):
Nice fix!

I suppose there isn't any easy fix for stopping the content of the "normal" page (the tables etc that are shown when there's no error message) from showing for a second before the NotFound page is loaded? (Obviously not part of this issue, though.)

inghylt (approved):
Nice fix! I think what you raised @MalinAhlberg is really good to look at as well but I think it belongs in a separate issue

e0:
Thanks for the reviews! Sounds like something we could tackle in a new issue. I notice that on the production server this type of errors usually show up within 100ms. We could still prevent the empty table from showing but maybe it's not too much of an issue for now. If we ever add server side rendering this problem could be completely eliminated.

PR #960, opened 2022-04-20
fix: remove unwanted whitespaces

### LINKED ISSUE #870: Unwanted whitespace before comma
**Describe the bug**

Unwanted white space has been added before comma or point at the end of sentences, as highlighted by the screenshot. 

**Expected behavior**

Comma or point should directly follow the character before.

**Screenshots**

**Additional context**

The problem should be caused by non-properly configured `linter`.

MalinAhlberg:
This PR closes #870.

Remove unwanted spaces from html pages, set prettier to not mess with important whitespaces. See **Further comments** sections for more details and discussion.

**Type of change**  

**List of changes made**  

- Removed whitespaces from a bunch of .vue files
- Set `htmlWhitespaceSensitivity` to `css`.

**Screenshot of the fix**  

whitespaces

**Testing**  

Verify that no more unwanted whitespaces are around.
To run prettier (with `ma-exec frontend yarn run format`), make sure that `.prettierrc.json` is updated in the container.

**Further comments**  

`htmlWhitespaceSensitivity` was set to `ignore` in a21f080b80e8b59f661b064710407d0b5b947a34, to stop prettier from splitting html tags. However, this will make prettier treat all whitespaces as non-important. I can't find any way to stop prettier from both introducing whitespaces and splitting tags, it seems like we have to choose. Even though splitted tags are annoying, they are not as annoying as having unwanted whitespaces on the webpage, so I propose having `htmlWhitespaceSensitivity` set to `css`. For more info see [the documentation](https://prettier.io/docs/en/options.html#html-whitespace-sensitivity).

I tried setting `htmlWhitespaceSensitivity` to `strict`, also tried setting  `"bracketSameLine": true`, but no success.

### DIFF
--- a/frontend/src/components/Documentation.vue
+++ b/frontend/src/components/Documentation.vue
@@ -45,8 +45,8 @@
             is a set of dedicated pages for various components, i.e., reactions, metabolites, genes,
             subsystems and compartments of the selected model, which is indicated on the left side
             of the top navigation bar. Each page contains a button on the right to
-            <i>Report an issue</i>
-            . Click this button to report a problem or ask questions to the modellers.
+            <i>Report an issue</i>. Click this button to report a problem or ask questions to the
+            modellers.
           </p>
 
           <h5 id="reaction-page" class="is-size-5">Reaction page</h5>
@@ -86,18 +86,14 @@
             Below the
             <i>Cross references</i>
             table comes the
-            <i>Reactions</i>
-            table. It lists all reactions involving the current metabolite as a reactant or product.
-            The current metabolite is highlighted in black in the reaction equations. This table can
-            be exported by clicking the button
-            <i>Export to TSV</i>
-            . Since metabolites are specific to a cell compartment, only reactions involving the
-            metabolite in its specific compartment are displayed. To remove this restriction and
-            display additional reactions involving the metabolite in any compartment, click the
-            button to
-            <i>See reactions from all compartments</i>
-            . Note that the number of retrieved reactions is limited to 1000. It is recommended to
-            use the
+            <i>Reactions</i> table. It lists all reactions involving the current metabolite as a
+            reactant or product. The current metabolite is highlighted in black in the reaction
+            equations. This table can be exported by clicking the button <i>Export to TSV</i>. Since
+            metabolites are specific to a cell compartment, only reactions involving the metabolite
+            in its specific compartment are displayed. To remove this restriction and display
+            additional reactions involving the metabolite in any compartment, click the button to
+            <i>See reactions from all compartments</i>. Note that the number of retrieved reactions
+            is limited to 1000. It is recommended to use the
             <a href="#api">API</a>
             to retrieve all reactions.
           </p>
@@ -105,8 +101,8 @@
             On the top right of the page, the
             <a href="#interaction-partners">Interaction Partners</a>
             tool for that metabolite can be accessed by clicking the button
-            <i>Interaction Partners</i>
-            . Below this button there is a widget to access the corresponding 2D or 3D
+            <i>Interaction Partners</i>. Below this button there is a widget to access the
+            corresponding 2D or 3D
             <a href="#map-viewer">Map Viewer</a>
             pages for this metabolite.
           </p>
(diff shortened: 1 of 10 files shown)


nanjiangshu (approved):
Very nice fix! I think it makes sense to set `htmlWhitespaceSensitivity` to `css`. Sometimes we have to accept tradeoffs.

e0 (approved):
LGTM

Really appreciate you adding the context @MalinAhlberg! I agree with you and @nanjiangshu that this is a good trade off.

mihai-sysbio:
In this [old post](https://prettier.io/blog/2018/11/07/1.15.0.html#whitespace-sensitive-formatting) there is also another option of using `span` and `div` blocks for different results, so I wonder if it is still applicable.

MalinAhlberg:
> In this [old post](https://prettier.io/blog/2018/11/07/1.15.0.html#whitespace-sensitive-formatting) there is also another option of using `span` and `div` blocks for different results, so I wonder if it is still applicable.

Good point! However, it still only provides the two options of either introduced whitespaces or splitted tags (but allows for overriding the value given in the `rc` file).

frontend/src/components/Documentation.vue
```diff
@@ -310,9 +304,8 @@
                 <i>Data overlay</i>
                 sidebar allows uploading of the custom data in
                 <a target="_blank" href="https://en.wikipedia.org/wiki/Tab-separated_values">
-                  TSV format
-                </a>
-                . If the file is parsed correctly, the file name will be highlighted in green; in
+                  TSV format </a
```
inghylt, line 307: Should we remove the space between `TSV format` and the dot?

frontend/src/components/Documentation.vue
```diff
@@ -571,9 +551,8 @@
             older models that may no longer be maintained (for example HMR 2.0), and others that
             were recently published. The more recent GEMs can also be found in the
             <a href="https://www.github.com/SysBioChalmers/" target="_blank">
-              SysBioChalmers organization GitHub
-            </a>
-            .
+              SysBioChalmers organization GitHub </a
+            >.
```
inghylt, line 555: Unnecessary extra space

frontend/src/components/Documentation.vue
```diff
@@ -331,9 +324,8 @@
                   target="_blank"
                   href="https://github.com/MetabolicAtlas/data-files/blob/main/DATA_OVERLAY.md#data-source-file-requirements"
                 >
-                  here
-                </a>
-                . The headers of these data series will be shown automatically in the dropdown
+                  here </a
+                >. The headers of these data series will be shown automatically in the dropdown
```
inghylt, line 328: Unnecessary extra space?

frontend/src/components/Documentation.vue
```diff
@@ -632,10 +611,9 @@
           <p>
             Metabolic Atlas has a
             <a href="/api/v2/" target="_blank">
-              dedicated interface to facilite the use of the API
-            </a>
-            , with output provided in JSON format. When using this service, kindly use a limit of 10
```
inghylt, line 637: Unnecessary extra space?

frontend/src/components/Home.vue
```diff
@@ -66,9 +66,8 @@
                     Wang, H., et al, 2021.
                     <i>
                       Genome-scale metabolic network reconstruction of model animals as a platform
-                      for translational research
-                    </i>
-                    . PNAS 118, e2102344118
+                      for translational research </i
+                    >. PNAS 118, e2102344118
```
inghylt, line 70: Unnecessary extra space?

inghylt (commented):
Commented on the places where I found to me a redundant whitespace between a link and a dot. Otherwise it looks good!

MalinAhlberg:
> Commented on the places where I found to me a redundant whitespace between a link and a dot. Otherwise it looks good!

Great that you caught those, @inghylt! Fixed in da96ca5

inghylt (approved):
Nice work!

mihai-sysbio:
> > In this [old post](https://prettier.io/blog/2018/11/07/1.15.0.html#whitespace-sensitive-formatting) there is also another option of using `span` and `div` blocks for different results, so I wonder if it is still applicable.
> 
> Good point! However, it still only provides the two options of either introduced whitespaces or splitted tags (but allows for overriding the value given in the `rc` file).

Just describe more clarify, I was thinking that if you'd want to keep the previous version of the `rc` file
```
"htmlWhitespaceSensitivity": "ignore"
```
the other alternative would be to change the `<p>` tags to `<span>`. Don't let this comment stop the merge though 😊

MalinAhlberg:
> > > In this [old post](https://prettier.io/blog/2018/11/07/1.15.0.html#whitespace-sensitive-formatting) there is also another option of using `span` and `div` blocks for different results, so I wonder if it is still applicable.
> > 
> > 
> > Good point! However, it still only provides the two options of either introduced whitespaces or splitted tags (but allows for overriding the value given in the `rc` file).
> 
> Just describe more clarify, I was thinking that if you'd want to keep the previous version of the `rc` file
> 
> ```
> "htmlWhitespaceSensitivity": "ignore"
> ```
> 
> the other alternative would be to change the `<p>` tags to `<span>`. Don't let this comment stop the merge though blush

Ah, I see your point now. Let's keep that in mind for future modifications :)

PR #965, opened 2022-04-26
fix: search for metabolite names give score zero

### LINKED ISSUE #896: Improve scoring

MalinAhlberg:
This PR closes #896, created by @inghylt and @MalinAhlberg 

When searching for a metabolite name (eg "3(R)-hydroxy-(2S,6R,10)-trimethyl-hendecanoyl-CoA"), the score for the corresponding metabolite node (MAM00685p) will no longer be zero.

**Type of change**  

**List of changes made**  

- Keep the connection between metabolites and compartmentalized metabolites in the neo4j queries (global and model search), so that the score can be correctly set
- adds test for this bug fix 
- in commit 9257786, use `>=` when testing for `max 50`
 
**Screenshot of the fix**  

scorefix

**Testing**  

Test searching for metabolite names and make sure they show up as first result.
Also verify that other searches work as expected, regarding both components and models.

**Further comments**  

- This part of the issue description "When searching for fatty the top result should not be the gene FASTK as top result" is not solved. This problem turned out to be related only to the internal scoring of neo4j, and therefore ignored in this issue.

### DIFF
--- a/api/src/neo4j/queries/search/globalSearch.js
+++ b/api/src/neo4j/queries/search/globalSearch.js
@@ -4,6 +4,7 @@ import {
   MODELS,
   COMPONENT_TYPES,
   CHILD_LABELS,
+  getScore,
 } from 'neo4j/queries/search/helper';
 
 const fetchCompartmentalizedMetabolites = async ({
@@ -28,11 +29,11 @@ UNWIND
       ELSE mids
   END AS mid
 MATCH (:Metabolite:${model} {id:mid})-[${version}]-(cm:CompartmentalizedMetabolite)
-WITH DISTINCT(cm.id) as cmid
+WITH DISTINCT(cm.id) as cmid, mid
 `;
   } else {
     statement += `
-WITH ${JSON.stringify(ids)} as cmids
+WITH ${JSON.stringify(ids)} as cmids, null as mid
 UNWIND
   CASE
       WHEN cmids = [] THEN [null]
@@ -58,9 +59,10 @@ CALL apoc.cypher.run('
   MATCH (s)-[${version}]-(ss:SubsystemState)
   RETURN { id: $cmid, subsystem: COLLECT({id: s.id, name: ss.name}) } as data
 ', {cmid:cmid}) yield value
-RETURN apoc.map.mergeList(apoc.coll.flatten(
+WITH apoc.map.mergeList(apoc.coll.flatten(
 	apoc.map.values(apoc.map.groupByMulti(COLLECT(value.data), "id"), [value.data.id])
-)) as metabolites
+)) as metabolites, mid
+RETURN metabolites {.*, mid: mid} as metabolites
 `;
 
   if (limit) {
@@ -391,7 +393,7 @@ LIMIT ${limit}
     if (result) {
       resWithScore[component] = result.map(obj => ({
         ...obj,
-        score: obj.id in uniqueIds ? uniqueIds[obj.id]['score'] : 0,
+        score: getScore(obj, uniqueIds),
       }));
     } else {
       resWithScore[component] = [];
--- a/api/src/neo4j/queries/search/helper.js
+++ b/api/src/neo4j/queries/search/helper.js
@@ -16,4 +16,14 @@ const COMPONENT_TYPES = [
 
 const CHILD_LABELS = ['ExternalDb', 'PubmedReference'];
 
-export { MODELS, COMPONENT_TYPES, CHILD_LABELS };
+const getScore = (node, uniqueIds) => {
+  if (node.id in uniqueIds) {
+    return uniqueIds[node.id]['score'];
+  }
+  if (node.mid) {
+    return uniqueIds[node.mid]['score'];
+  }
+  return 0;
+};
+
+export { MODELS, COMPONENT_TYPES, CHILD_LABELS, getScore };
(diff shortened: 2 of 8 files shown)


api/src/neo4j/queries/search/globalSearch.js
```diff
@@ -28,11 +29,11 @@ UNWIND
       ELSE mids
   END AS mid
 MATCH (:Metabolite:${model} {id:mid})-[${version}]-(cm:CompartmentalizedMetabolite)
-WITH DISTINCT(cm.id) as cmid
+WITH DISTINCT(cm.id) as cmid, mid
```
nanjiangshu, line 32: Is `cmid` for CompartmentalizedMetaboliteIDs and `mid` for `MetaboliteIDs`? Is that because we have previously ignored the potential MetaboliteIDs (IDs without the ending compartment letter) that caused the problem with zero score hit?

api/test/search.test.js
```diff
@@ -17,7 +17,7 @@ describe('search', () => {
 
     const { metabolite } = data['Human-GEM'];
     expect(metabolite.length).toBeGreaterThan(0);
-    expect(metabolite.length).toBeLessThan(50);
+    expect(metabolite.length).toBeLessThan(51);
```
nanjiangshu: If I remember correctly, @e0 made it by purpose to demonstrate a failed test, since I was suggesting using `toBeLessThanOrEqual(50)`. Maybe we don't need this failed testing example anymore?

api/test/jest.setup.js
```diff
@@ -1,4 +1,6 @@
 global.beforeAll(() => {
   global.API_BASE = 'http://localhost:8081/api/v2';
   global.HUMAN_GEM_VERSION = '1_10_0';
```
nanjiangshu, line 4: Is that possible to define the VERSION for GEMs from e.g. `api/src/data/integratedModels.json`? Otherwise, we may forget to update them when we are updating the models.

nanjiangshu (commented):
Really nice work @inghylt and @MalinAhlberg! I tested with the examples you provided and it looks great. I also tried with a number of random examples for different models and no weird behavior was found. 

Regarding the searching result for e.g. `fatty`, I think it makes sense to leave it for now sine it is how neo4j searching works. 

Apart from that, I made a few minor comments.

api/src/neo4j/queries/search/globalSearch.js
MalinAhlberg: yes, exactly. we need to keep the metabolites id:s from the first part of the search since they are the ones that get the scores from neo4j.

api/test/jest.setup.js
MalinAhlberg: Good point! Added in afe2cc5

api/test/jest.setup.js
```diff
@@ -1,4 +1,27 @@
...
+    'gene',
+    'reaction',
+    'subsystem',
+    'compartment',
+  ];
 });
+
+function readModelInfo() {
+  let json = require('../src/data/integratedModels.json');
+  let result = {};
```
e0, line 22: Really minor detail here. `const` should be used instead of `let` when the values are not reassigned. Note that for the case of `result`, the object does can have its properties modified, a `let` should only be used if it is reassigned (`result = foo`). More info here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const

```suggestion
  const json = require('../src/data/integratedModels.json');
  const result = {};
```

e0 (approved):
Looks great! Left a couple of comments but please feel free to merge without addressing them.

api/test/search.test.js
inghylt: @e0 any comment on this?

api/test/jest.setup.js
inghylt: @MalinAhlberg I'm committing this, writing this comment so you can trace it

api/test/search.test.js
e0: Looks like this has been changed to `toBeLessThanOrEqual(50)` in the new files, which seems very appropriate.

api/src/neo4j/queries/search/helper.js
```diff
@@ -16,4 +16,14 @@ const COMPONENT_TYPES = [
...
-export { MODELS, COMPONENT_TYPES, CHILD_LABELS };
+const getScore = (node, uniqueIds) => {
+  if (node.id in uniqueIds) {
+    return uniqueIds[node.id]['score'];
+  }
+  if (node.mid) {
+    return uniqueIds[node.mid]['score'];
+  }
+  return 0;
+};
```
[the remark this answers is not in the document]
e0: That's a good idea to look into @MalinAhlberg ! Sorry for including the extra line! Nice that you spotted and fixed it.

MalinAhlberg:
> If we still want to improve the searching for e.g. fatty in the future, try to create a new issue for that.

#978 is created for that :)

Issue #789, opened 2021-11-30
Overlayed data missing when switching between maps


MalinAhlberg:
**Describe the bug**
When moving between maps with the data overlay sidebar open,  the selected tissue is still there but the transcriptomics data will not be overlayed on the 2D map

**Steps to reproduce**

- go to `/explore/Human-GEM/map-viewer/protein_modification?dim=2d&panel=1&sel=&search=&coords=-830.11,-1766.88,0.32,0,0,500&datatype=transcriptomics&datasource=hpaRna.tsv&dataSet=cerebral%20cortex` (protein modification with overlay for cerebral cortex chosen)

- change maps to eg `purine metabolism` and then change back to `protein modification` without closing the overlay sidebar

Cerebral cortex is still selected in the sidebar, and set in the url, but the colored overlay is not shown an the map.

**Expected behavior**
Either the overlay should be shown on the map, or removed from the url and deselected from the menu in the sidebar

PR #967, opened 2022-04-27
fix: overlay disappears on collapsing bar

### LINKED ISSUE #935: Data overlay automatically disappears upon hiding data overlay window
In looking at the 3D maps with overlaid data, the coloring disappears automatically upon hiding the data overlay window. This is not ideal as it hinders exploration of the map. Perhaps it would be desirable to reset the colors of the map by going back to the closed data overlay and selecting `None`.

Steps to reproduce the problem:
1. Go on 3D viewer and select mitochondria (I found the same behavior in cytoscol, have not tested others);
2. Click the green vertical "data overlay" tab;
3. Select levels from Liver;  
4. Close the green vertical "data overlay" tab;
5. Levels have now disappeared.

Desired results:
- Steps 1-4 are the same;  
- Step 5 retains levels;
- To reset levels, click green vertical tab and select "None" for the levels from HPA.

I have not tried this with my own uploaded data, but if the behavior is still there it may be a problem.

**Acceptance criteria**
- [ ] a collapsed bar doesn't reset the overlay

**Optional**
- [ ] a collapsed bar still indicates that overlay is applied ([example](https://github.com/CreativeBulma/bulma-badge/))-> conceptually makes more sense to have a new issue instead

MalinAhlberg:
This PR closes #935 and #789.

Changes of the overlay is kept even if the overlay bar is collapsed.
The levels / overlay is kept when switching between 2D/3D, when a new map is chosen, or if leaving the Maps and then returning.

**Type of change**  

**List of changes made**  

- [X] overlay data is not reset when collapsing the sidebar
- [X] overlay data is kept when reopening the sidebar
- [X] overlay is always applied when a 2D map is drawn (already present for 3D)
- [X] keep the values when (re-)entering the Map Viewer / switching maps.

As a consequence, issue #789 was solved.

**Testing**  

Navigate to the map viewer, open any map and then open the `Data Overlay` side bar. Add a Level.
Verify that the overlay is kept (on the map, in the dropdown, in the url) when collapsing and reopening the overlay side bar. 2D and 3D  should work the same way.
Also test that the overlay works as expected when moving around on the page.

**Further comments**  

I did not implement the optional acceptance criterion (adding some sort of indicator for when overlay is added), but I think it would make much sense to add it, especially since the overlay is kept when you leave and re-enter the map viewer. I will open a new issue for this.

### DIFF
--- a/frontend/src/components/explorer/MapViewer.vue
+++ b/frontend/src/components/explorer/MapViewer.vue
@@ -160,7 +160,7 @@
 </template>
 
 <script>
-import { mapGetters, mapState } from 'vuex';
+import { mapGetters, mapState, mapActions } from 'vuex';
 import { debounce } from 'vue-debounce';
 import DataOverlay from '@/components/explorer/mapViewer/DataOverlay.vue';
 import ErrorPanel from '@/components/shared/ErrorPanel';
@@ -249,6 +249,9 @@ export default {
     this.loadMapFromParams();
   },
   methods: {
+    ...mapActions({
+      resetOverlay: 'dataOverlay/resetValues',
+    }),
     handleSidebarScroll() {
       if (this.$refs.mapSidebar.scrollTop > 0) {
         this.sidebarLayoutReset = false;
@@ -301,6 +304,7 @@ export default {
       }
     },
     loadMapFromParams() {
+      this.resetOverlay();
       const id = this.$route.params.map_id;
       if (id) {
         const categories = Object.keys(this.mapsListing);
--- a/frontend/src/components/explorer/MapViewer.vue
+++ b/frontend/src/components/explorer/MapViewer.vue
@@ -160,7 +160,7 @@
 </template>
 
 <script>
-import { mapGetters, mapState, mapActions } from 'vuex';
+import { mapGetters, mapState } from 'vuex';
 import { debounce } from 'vue-debounce';
 import DataOverlay from '@/components/explorer/mapViewer/DataOverlay.vue';
 import ErrorPanel from '@/components/shared/ErrorPanel';
@@ -249,9 +249,6 @@ export default {
     this.loadMapFromParams();
   },
   methods: {
-    ...mapActions({
-      resetOverlay: 'dataOverlay/resetValues',
-    }),
     handleSidebarScroll() {
       if (this.$refs.mapSidebar.scrollTop > 0) {
         this.sidebarLayoutReset = false;
@@ -304,7 +301,6 @@ export default {
       }
     },
     loadMapFromParams() {
-      this.resetOverlay();
       const id = this.$route.params.map_id;
       if (id) {
         const categories = Object.keys(this.mapsListing);
--- a/frontend/src/components/explorer/mapViewer/DataOverlay.vue
+++ b/frontend/src/components/explorer/mapViewer/DataOverlay.vue
@@ -195,7 +195,7 @@ export default {
       filename: datasource,
       propagate: false,
     });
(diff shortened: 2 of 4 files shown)


e0:
This looks really good @MalinAhlberg!

I would like to understand a few things better before approving.

1. I think resetting the values is ok as a solution and it is clearer now that the levels dropdown is set to `None` after changing maps. It is also nice that the 3D one is consistent. This does mean that the 3D viewer no longer keeps the levels when changing maps. I'm not too familiar with how researchers would use this but it seems like there could be some value if we can keep the level when changing maps so that the user can quickly see how the same levels apply to different compartments/subsystems. Do you think it's worth adding a new issue to enable this for both 2D and 3D in the future?
2. I notice that when switching from 2D to 3D, the levels are kept and they apply to the 3D network. When switching from 3D to 2D though, the levels are still kept in the dropdown but they do not seem to be applied to the 2D map. Do you know if we have an issue for this or is it something that is closely enough related to #789?

nanjiangshu (commented):
Great work @MalinAhlberg 

Except for the comment which @e0 made (which I think they are features to be discussed rather than bugs), I have one more comment. For the 2D Map viewer, when one select a type for expression levels, the data is overlayed on the 2D map and also shown on the URL, however, when one make a hard refresh or open the URL in a new window, the overlayed data is gone. This is the same for the 3D map viewer. I think we have previously discussed that a user should get consistent result with the same URL (so that it can be shared) or makes a hard refresh.

MalinAhlberg:
Thanks @e0 and @nanjiangshu for your reviews, very valuable!

The issue @nanjiangshu is fixed in db088d0 (please verify this).

The second issue mentioned by @e0 actually seems to be due to a slightly unrelated bug. I will see if it can be solved easily, and  if so include it in this PR, otherwise create a new issue.

For the first issue from @e0 , I'm also not familiar with what researchers would like. But I could of course change the behavior in this PR so that the overlay is kept when switching maps. Any input from @mihai-sysbio? Otherwise we could create a new issue to ask an advisor about this.

MalinAhlberg:
> The second issue mentioned by @e0 actually seems to be due to a slightly unrelated bug. I will see if it can be solved easily, and if so include it in this PR, otherwise create a new issue.

I created #971 for this.

mihai-sysbio:
> overlay is kept when switching maps

That would be great to have - it's very useful to avoid unnecessary clicks, especially when having multiple data types overlaid (not there atm but in the future).

nanjiangshu:
> The issue @nanjiangshu is fixed in [db088d0](https://github.com/MetabolicAtlas/MetabolicAtlas/commit/db088d0f63e8533ad486de4b6f03ec2c5acdc9a8) (please verify this).

Nice fix! I can verify that the overlay is preserved with a hard refresh or opening the URL in another window. And also the overlay is preserved when switching maps. 

I think the second point @e0 mentioned is also solved, when I tested to switch between 2D/3D repeatedly, the highlights of the expression levels are preserved. However, the popup of the expression levels when mouse-on genes is not working after switching from 3D to 2D, I will create a bug report for that. In addition, mouse-on genes in 3D maps does not show the numeric expression level, I'm not quite sure if it is always behaving like this or a bug, maybe @e0, @mihai-sysbio, @inghylt have an idea of that.  

It's great that you created the issue #971 though I think it is a bug not closely related to this issue, see my comment at https://github.com/MetabolicAtlas/MetabolicAtlas/issues/971#issuecomment-1118358783

I think current fixings have addressed all points of this issue and I can approved it as is.

nanjiangshu (approved):

inghylt (approved):
Very nice work!

PR #972, opened 2022-05-04
feat: timeline gems visualization

### LINKED ISSUE #949: Timeline view for integrated GEMs

As a user, I want to visualize integrated GEMs in a timeline view so that I can understand their history of development.

**Acceptance Criteria**

- [ ] The documentation is updated accordingly (on the webpage)
- [ ] The changes do not reduce the performance (e.g. by adding a large library)
- [ ] The timeline is visually appealing and interactive (at least clicking on nodes for more details) based on the design from #929

**Additional context**

- If the related data collection issue is not completed at the start of the issue, use mock data.
- If the related data collection issue is not completed when a PR for this issue is ready to be made, please set the PR to draft mode.

e0:
This PR closes #949 .

**Type of change**  

**List of changes made**  

Add a timeline chart for history of integrated models for the `GEM Repository` page

**Screenshot of the fix**  

image
image

**Testing**  

- Instructions on how to test
  1. This PR requires the work from https://github.com/MetabolicAtlas/data-generation/pull/27. 
  2. Rebuild the project and start it: `build-stack && start-stack`.
  3. Visit the `/gems/repository` page and the timeline chart should appear below the list of integrated models.
  4. Click different versions (circles) and verify that the links work correctly.


frontend/src/components/Repository.vue
```diff
@@ -59,7 +59,8 @@
             </div>
           </div>
         </div>
-        <br />
+        <h4 id="integrated-models-history" class="title is-4">History of integrated models</h4>
+        <gem-history />
```
mihai-sysbio, line 63: I'd suggest adding some padding, either below `<gem-history>` or above the `<h3>` title below.

mihai-sysbio (changes requested):
This is a huge step for Metabolic Atlas 🚀  well done @e0 and the team for making this feature happen! We had intense discussions on everything from data format, data collection and mockup, but the efforts have paid off 🌟 

The only reason to request changes is that I'm not getting the pop-up to appear, nor do I get any errors in the console, so I'll approve this if I'm the only one with this problem.

Edit: my problem might have to do that all the bubbles are blue, whereas the screenshot above contains some green ones.

frontend/src/components/Repository.vue
inghylt: I would also suggest some extra space, but between this `<h4>` and the `<div>` above it that relates to `Integrated models`

inghylt:
Really nice work so far :clap: :medal_sports: :trophy: 

- The popup works for me if you replace line 134 in `frontend/src/components/GemHistory.vue` with `const c = e.target;`
- It would be nice to be able to see clearly that  Zebrafish-GEM stems from Human-GEM and not e.g. Rat-GEM which you might interpret from the current layout. Is there any way to make the branches interactive? For example if you hoover/click them they are hightlighted/marked in some way? If so this could be used to make the relationship clearer
- In the popup, having `Citations: Human-GEM-1.5.0` for the 1.0.0 of Mouse/Rat etc. is a bit confusing for me. A suggestion would be to rename this to for example `Template GEM:  Human-GEM-1.5.0`. Another suggestion could also be to have this information in the related branch and not the node, e.g. if you click or hover the branch a popup with this information is displayed.
- I wish I had a good suggestion regarding the non optimal Yeast GEM versions
  - 7.6.0 and 7.70 being released on the same day as mentioned [here](https://github.com/MetabolicAtlas/MetabolicAtlas/issues/949#issuecomment-1116008366)
  - 8.1.0 and 7.7.0 being so close in time so that the nodes overlap
 
   but I can't come up with something. I'm fine with keeping things as is if nobody comes up with a good solution since you can click the release notes to find out more about them

frontend/src/components/GemHistory.vue
```diff
@@ -0,0 +1,219 @@
+<template>
+  <div id="timeline-wrapper">
+    <div id="timeline-svg-wrapper" ref="wrapper">
+      <inline-svg
```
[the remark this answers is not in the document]
e0: For those SVG files, we are loading them from the server so we are not loading the `.svg` filetype directly in the frontend. In the server, we are using the file system (`fs`) to read in these files and returning them.

If we want to load the `.svg.` filetype directly in the frontend, we need some type of file loader. Alternatively we could add this SVG file to the `api` instead and load it from the frontend. That would be one less library to include but I'm not sure if it would be better in terms of developer (more code needed) and user (potentially longer page load time) experiences.

frontend/src/components/Repository.vue
e0: I have added some margins here: https://github.com/MetabolicAtlas/MetabolicAtlas/pull/972/commits/84ec2526c74c87dee796d92d2160a306e0fedf9f

Please let me know if you think they should be adjusted. 
image

e0:
> * The popup works for me if you replace line 134 in `frontend/src/components/GemHistory.vue` with `const c = e.target;`

I didn't get the problem with `e.path[0]` but `e.target` works for me too and is cleaner. Have addressed this in https://github.com/MetabolicAtlas/MetabolicAtlas/pull/972/commits/bd45a2a76bddedd3e6c9e0a3023b15ad8323a13e

___

> * It would be nice to be able to see clearly that  Zebrafish-GEM stems from Human-GEM and not e.g. Rat-GEM which you might interpret from the current layout. Is there any way to make the branches interactive? For example if you hoover/click them they are hightlighted/marked in some way? If so this could be used to make the relationship clearer

This is a really good idea! Currently there is no metadata added to the edges but I think we could highlight the parent circle. I will try to work on this.

___

> * In the popup, having `Citations: Human-GEM-1.5.0` for the 1.0.0 of Mouse/Rat etc. is a bit confusing for me. A suggestion would be to rename this to for example `Template GEM:  Human-GEM-1.5.0`. Another suggestion could also be to have this information in the related branch and not the node, e.g. if you click or hover the branch a popup with this information is displayed.

I'm happy to change this. Two notes:
1. Currently a list style is used so that it would work well in the case of multiple parent ids and citations. Should that be `Template GEMs` instead? Open to other suggestions here too.
2. I think the UX for clicking an edge could get complicated. I think it would help to increase the hover/click area of but we would have to make sure that the area doesn't interfere with other elements. I'm also a bit concerned that it would not be intuitive to know that edges can be clicked on.

inghylt:
@e0 Nice!

> Currently a list style is used so that it would work well in the case of multiple parent ids and citations. Should that be Template GEMs instead? Open to other suggestions here too.

Good thinking! Maybe add the extra `s` if the length of `externalParentIds` is larger than 1?

> I think the UX for clicking an edge could get complicated. I think it would help to increase the hover/click area of but we would have to make sure that the area doesn't interfere with other elements. I'm also a bit concerned that it would not be intuitive to know that edges can be clicked on.

Maybe one could color or somehow mark the special branches visually so you can see from the beginning that they are different? But I'm fine with keeping the branches non interactive as well, just an idea if it's easy to implement

e0:
Have added a commit that gives the parents the same `selected` color. Do you like this change?

image

frontend/src/components/Repository.vue
mihai-sysbio: My suggestion is 1,5-2x the space above `GEM Repository` (as shown in the figure above).

e0:
> For overlapping circles, as mentioned [here](https://github.com/MetabolicAtlas/data-generation/pull/27#pullrequestreview-962937853), one idea could be to temporarily add a very high z-index for the circle that is hovered, so that it is brought on top of the overlapping one.

This is a good suggestion but it doesn't complete solve the problem if two versions for the same model are released on the same date.

frontend/src/components/Repository.vue
mihai-sysbio: > increase the space between the chart and `GEM Repository`

Yes, by a lot :)

e0: Ok sure. FYI bulma runs out of helper numbers after `-6` (https://bulma.io/documentation/helpers/spacing-helpers/). The inconsistent scale also bothers me. @mihai-sysbio do you prefer using bulma helper for the top margin and custom css for the bottom margin? FWIW, I would use custom css for both in this case.

frontend/src/components/GemHistory.vue
```diff
@@ -0,0 +1,239 @@
...
+          }, 0);
+        });
+      });
+
+      this.$refs.inlineSvg.$el.addEventListener('click', e => {
+        const el = e.target;
+        if (el.tagName !== 'circle') {
+          this.clearSelection();
+        }
+      });
```
mihai-sysbio, line 178: This only works for dismissing the pop-up only when click some other place on the SVG. Intuitively I was expecting it to be dismissed when click anywhere. Any thoughts?

mihai-sysbio:
> Since Shan does not have time for this issue this week, I can help with the styling, such as font size or adding explanations.

@nanjiangshu for an easier continuation and review of the improvements please use #974 . My suggestion is to merge that in first, and then merge in the current PR.

frontend/src/components/GemHistory.vue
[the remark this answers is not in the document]
mihai-sysbio: Thanks for the explanation @e0. I guess one option would be to serve this file from the API as well, but I'm not sure that would be an improvement in any way.

e0:
Thanks again to everyone for taking time to review and provide feedback! I have created a new issue #976 for us to continue the discussion and improve the timeline chart. I have compiled the action points into the acceptance criteria.  Please let me know if I missed anything.

inghylt (approved):
Great work! Makes sense to have a new issue for improvements

mihai-sysbio (approved):

PR #974, opened 2022-05-05
style: improve timeline for gems repository

mihai-sysbio:
This PR is seen as an integral part of #972. It has been separated for a simpler review of the changes.

### DIFF
--- a/frontend/src/components/GemHistory.vue
+++ b/frontend/src/components/GemHistory.vue
@@ -13,6 +13,7 @@
       class="box"
       :style="{ top: popoverOffset.top, left: popoverOffset.left }"
     >
+      <h5 class="is-size-5 mt-0 mb-2">{{ selectedVersion.model }} {{ selectedVersion.version }}</h5>
       <div v-if="currentVersionData">
         <p>
           <router-link :to="{ name: 'browser', params: { model: selectedVersion.model } }">
@@ -35,10 +36,10 @@
         <p>Model: {{ selectedVersion.model }}</p>
         <p>Released: {{ selectedVersion.releaseDate }}</p>
         <div v-if="selectedVersion.externalParentIds.length > 0">
-          <p>Citations:</p>
+          <p>Based on:</p>
           <ul>
             <li v-for="{ id, citLink } in selectedVersion.externalParentIds" :key="id">
-              <a :href="citLink" target="_blank">{{ id }}</a>
+              {{ id }} (<a :href="citLink" target="_blank">source</a>)
             </li>
           </ul>
         </div>
--- a/frontend/src/components/GemHistory.vue
+++ b/frontend/src/components/GemHistory.vue
@@ -130,7 +130,7 @@ export default {
         const { model, version } = { ...circle.dataset };
         const matchingModel = this.getMatchingModel(model, version);
         if (matchingModel) {
-          circle.style.fill = '#25543c'; // eslint-disable-line no-param-reassign
+          circle.classList.add('integrated');
         }
       });
     },
@@ -190,16 +190,22 @@ export default {
     svg {
       .circle {
         cursor: pointer;
-        fill: $link;
+        fill: $primary-lighter;
 
         &:hover,
         &.selected {
-          fill: $icon-interaction-partner !important;
+          fill: $warning !important;
+        }
+
+        &.integrated {
+          stroke: $primary;
+          stroke-width: 3;
         }
       }
 
       .label {
         pointer-events: none;
+        fill: $primary !important;
       }
     }
   }
--- a/frontend/src/components/GemHistory.vue
+++ b/frontend/src/components/GemHistory.vue
@@ -149,13 +149,6 @@ export default {
             element: c,
           };
 
-          this.selectedVersion.externalParentIds.forEach(({ id }) => {
-            const p = Array.from(circles).find(
-              _c => id === `${_c.dataset.model}-${_c.dataset.version}`
-            );
-            p.classList.add('selected');
-          });
-
           // timeout is needed to make sure the popover element appears before calculation
           setTimeout(() => {
             this.popoverOffset = {
@@ -179,9 +172,7 @@ export default {
     },
     clearSelection() {
       if (this.selectedVersion) {
-        document.querySelectorAll('svg .selected').forEach(c => {
-          c.classList.remove('selected');
-        });
+        this.selectedVersion.element.classList.remove('selected');
         this.selectedVersion = null;
       }
     },


mihai-sysbio:
Here is a screenshot of the changes introduced by the commits above:
image

inghylt:
- I think it is good that we try to be consistent with the colors used on the web site. However, the current colors makes it much harder for me to read/interpret the graph. Might be because the lack of contrast with the background? Maybe the nodes could be green with white version text to improve readability.
- I think the `<h5>` title of the popup is very large compared to the rest of the text, the popup in #972 is more even and nice to look at in terms of different font sizes. I do like having the model and version as title though

mihai-sysbio:
> This looks pretty good @mihai-sysbio ! I like the color changes overall. Here are some thoughts. Please note that I understand these are difference in preferences so feel free to disregard them.
> 
>     1. I like the emphasis on the integrated versions. The stroke width seems just a tad heavy to me. Personally I might prefer `2` instead of `3`.

Very nice! Implemented in [36f54fd](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/974/commits/36f54fd8efd525aa46e2ba9e923a2b38b12d0cae).

>     2. I thought @inghylt made a good suggestion with highlighting the parent. It would have been better to highlight the branch/edge too but I thought highlighting the parent circle was still an improvement. How come this changes was reverted?

Apologies, I should have explained my thinking here:
- too many colors made the chart look distracting, even though it is as its "smallest" (the plan is to expand it)
- if we think of Mouse-GEM 1.1, it is a child of 1.0, and 1.0 has Human-GEM 1.5 as parent. However, Mouse-GEM 1.1 is still based on Human-GEM 1.5. I do realise that this highlighting was in line with the "Based on" text in the pop-up, but overall my feeling was that this introduced confusion simply because we were forcing this information into the user's attention.
- while highlighting something that is hovered doesn't need explanations, "remote" highlighting needs to be explained. If the edge would have been highlighted as well I think it would have helped, but then one could imagine highlighting the nodes when hovering over an edge. Personally, I felt like this highlighting was not intuitive enough in its current state to be deployed without an explanation.

>     3. Adding a `(source)` as a link instead seems unnecessary compared to before.

I see your point. My attempt was to resolve what I perceived as non-intuitive behavior when clicking a GEM name and not arriving at that GEM. But perhaps this is a non-problem.

I'm curious to see what others think about 2. and 3. above.

mihai-sysbio:
> * I think it is good that we try to be consistent with the colors used on the web site. However, the current colors makes it much harder for me to read/interpret the graph. Might be because the lack of contrast with the background? Maybe the nodes could be green with white version text to improve readability.

In the current implementation, the font color needs to always be the same, regardless if the circle is selected or not. Given the current color palette that has more hues of lighter colors, it was easier to choose a darker color for the font and have different colors for the selection. 

image

A disclaimer, I'm not picking up on the contrast issue. I've experimented with a `black` font below - @inghylt is the contrast better than in the original post?
image

>     * I think the `<h5>` title of the popup is very large compared to the rest of the text, the popup in [feat: timeline gems visualization #972](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/972) is more even and nice to look at in terms of different font sizes. I do like having the model and version as title though

I'm using a font size from another place on the website (aiming for consistency). My interpretation of @nanjiangshu's [feedback](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/972#issuecomment-1118607504) was that the font is too small in the pop-up as well.

e0:
> * too many colors made the chart look distracting, even though it is as its "smallest" (the plan is to expand it)
> * if we think of Mouse-GEM 1.1, it is a child of 1.0, and 1.0 has Human-GEM 1.5 as parent. However, Mouse-GEM 1.1 is still based on Human-GEM 1.5. I do realise that this highlighting was in line with the "Based on" text in the pop-up, but overall my feeling was that this introduced confusion simply because we were forcing this information into the user's attention.
> * while highlighting something that is hovered doesn't need explanations, "remote" highlighting needs to be explained. If the edge would have been highlighted as well I think it would have helped, but then one could imagine highlighting the nodes when hovering over an edge. Personally, I felt like this highlighting was not intuitive enough in its current state to be deployed without an explanation.

Thanks for taking time to explain. Of these three points, I can fully buy this: _"I felt like this highlighting was not intuitive enough in its current state to be deployed without an explanation._" I don't understand the "too many colors" argument since the parent had the same color as the selected one. Maybe you meant that it became "too colorful", which would make more sense to me but is still something I probably disagree with. In the future, I think it would be helpful to visually indicate the versions directly involved when a branching occurs, especially when they overlap like it currently does.

___


> I see your point. My attempt was to resolve what I perceived as non-intuitive behavior when clicking a GEM name and not arriving at that GEM. But perhaps this is a non-problem.

Ah that makes sense. I guess it was more self-explanatory before when the listing label was `Citations` instead of `Based on`. The `(source)` version has another benefit for the case if an `externalParent` does not have any citation link, in which case we can just keep the text and omit the `(source)` part.

mihai-sysbio:
> Thanks for taking time to explain. Of these three points, I can fully buy this: _"I felt like this highlighting was not intuitive enough in its current state to be deployed without an explanation._" I don't understand the "too many colors" argument since the parent had the same color as the selected one. Maybe you meant that it became "too colorful", which would make more sense to me but is still something I probably disagree with.

Thanks for the follow-up @e0. I would suggest to continue the discussion irl, but then I'm not sure if we can do that before the sprint ends. I'm open to suggestion regarding what should be presented in the sprint review, for example in the case of not merging in this PR.

> In the future, I think it would be helpful to visually indicate the versions directly involved when a branching occurs, especially when they overlap like it currently does.

Yes! I think @inghylt commented along the same lines, so I'm expecting to get this feedback from the advisors as well.

e0:
There seems to be a lot of differences in opinions. Here are my suggestions on what should be added to #972 so that we can close this issue ASAP and present at the review.

1. The changes in this PR for popovers, preferably with smaller font size for title.
2. Remove the parent highlighting as done in this PR.
3. Increase font sizes as mentioned by @nanjiangshu.

To be clear, I personally don’t have a problem with either the contrast in this PR nor the font sizes. But to me these are non-negotiable. If the visual impact is that noticeable to some of us, it most definitely will be perceived similarly by some other users. Then we can add a new issue to discuss and improve the things we currently disagree on.

MalinAhlberg:
General comment on this PR and on #972: this is so nice :tada:!
I prefer the colors and styles of this PR, but I think the others looks good too. For the highlighting of parents, the most intuitive solution to me would be to highlight the edges when hovering. So hovering any node will highlight it's entire line, in both directions.

Having that said, I'm totally fine with going for the suggestions @e0 mentioned above and close this / create new issues.

Issue #991, opened 2022-05-17
Refactor code for fetching and handling GEMs


e0:
**Please describe the feature**

As a developer, I want to refactor the code for fetching and handling GEMs so that the code is more consistent with how Vuex modules and API fetching works for other instances.

**Acceptance Criteria**

- [ ] In `frontend/src/api/gems.js`, the `fetchGems` function includes some mapping logic. An argument could be made that the code for mapping should be included in the corresponding `frontend/src/store/modules/gem.js` instead. Consider if this should be done and if so update the files.
- [ ] The changes are in alignment with the privacy policy
- [ ] The changes are tested on multiple models and browsers
- [ ] The API response is the same as before
- [ ] The changes do not affect the deployment
- [ ] The changes do not reduce the performance

**Additional context**

inghylt:
Seems like `frontend/src/components/explorer/InteractionPartners.vue` in commit 890235ed82cd32fc6d83a8f39de69164598badd8 had a `TODO` about Vuex as well, should it be included in this issue?

inghylt:
Ah sorry I got it mixed up yesterday, thank for pointing out the obvious @e0

Issue #812, opened 2021-12-20
refactor interaction partners


e0:
**Please describe the feature**

As a developer, I want to refactor the interaction partners page logic, so that it would be easier to work with in the future.

**Additional context**

Currently there is [a huge file](https://github.com/MetabolicAtlas/MetabolicAtlas/blob/develop/frontend/src/components/explorer/InteractionPartners.vue) that could be refactored into smaller, more modular components/files.

e0:
There is good opportunity to refactor code in the `load` and `loadExpansion` functions. Consider using Vuex where appropriate.

mihai-sysbio:
Decided to close this is as it will be baked into another issue.

PR #992, opened 2022-05-17
chore: remove TODO comments

e0:
This PR closes https://github.com/MetabolicAtlas/private-issues/issues/34.

**Type of change**  

- [x] Other
 
**List of changes made**  

- Removed `TODO` comments from code.
- Added [a comment](https://github.com/MetabolicAtlas/MetabolicAtlas/issues/812#issuecomment-1128633785) to #812
- Created https://github.com/MetabolicAtlas/MetabolicAtlas/issues/991

**Testing**  

Only comments have been removed in this PR so there should be no noticeable changes to the user. Some `TODO`s were removed without creating an issue. I found them to be non-issues but please let me know if you think differently and I'm happy to create new issues for them.

### DIFF
--- a/frontend/src/api/europepmc.js
+++ b/frontend/src/api/europepmc.js
@@ -34,9 +34,7 @@ const searchReferences = async queryIds => {
         refDetails.title
       }</i> ${refDetails.journal}`;
       refs[details.id] = refDetails;
-    } catch {
-      // TODO: handle exception
-    }
+    } catch {}
   });
 
   return refs;
--- a/frontend/src/api/europepmc.js
+++ b/frontend/src/api/europepmc.js
@@ -34,7 +34,7 @@ const searchReferences = async queryIds => {
         refDetails.title
       }</i> ${refDetails.journal}`;
       refs[details.id] = refDetails;
-    } catch {}
+    } catch {} // eslint-disable-line no-empty
   });
 
   return refs;
--- a/frontend/src/store/modules/maps.js
+++ b/frontend/src/store/modules/maps.js
@@ -109,7 +109,6 @@ const actions = {
         apiFunc = metabolitesApi.fetchMetaboliteData;
         break;
       default:
-        // TODO: handle unexpected type
         break;
     }
 
@@ -165,7 +164,6 @@ const actions = {
   },
 
   initFromQueryParams({ commit }, { dim, panel, coords, sel, search }) {
-    // TODO: handle errors
     commit('setShowing2D', dim !== '3d');
     commit('setDataOverlayPanelVisible', panel === '1');
     commit('setSelectedElementId', sel);
--- a/api/src/neo4j/queries/relatedReactions.js
+++ b/api/src/neo4j/queries/relatedReactions.js
@@ -9,7 +9,6 @@ const NODE_TYPES = {
   compartment: 'Compartment',
 };
 
-// TODO: add pagination and search
 const getRelatedReactions = async ({
   nodeType,
   id,
--- a/frontend/src/api/gems.js
+++ b/frontend/src/api/gems.js
@@ -2,7 +2,6 @@ import axios from 'axios';
 
 const fetchGems = async () => {
   const { data } = await axios.get('repository/models/');
-  // TODO consider moving this mapping logic into store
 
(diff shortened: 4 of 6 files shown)


frontend/src/api/europepmc.js
```diff
@@ -34,7 +34,7 @@ const searchReferences = async queryIds => {
         refDetails.title
       }</i> ${refDetails.journal}`;
       refs[details.id] = refDetails;
-    } catch {}
+    } catch {} // eslint-disable-line no-empty
```
inghylt, line 37: I wonder if there is a better way to solve this than having an empty catch like this and having to disable a warning? Shouldn't we try to catch and somehow react on an error with the get request?

inghylt:
The removal of TODO:s in `frontend/src/store/modules/maps.js`, we don't want to create an issue or something for those?

frontend/src/api/europepmc.js
e0: I agree that empty `catch` blocks are not a good practice. In this case, the `try` is needed so that no error messages are shown when it fails. Requests failing sometimes is expected here. I think failing and empty responses both lead to the following message that's displayed to the user.

image

inghylt: For empty responses I think the `There are no associated references` makes sense. If the requested failed due to for example a server problem, then that message is not necessarily true. It would be bad UX IMO to trigger an error message since this is not the core functionality of the gem browser, but maybe one could display something like `Not possible to access Europe PMC right now` or similar

inghylt:
@e0 Yeah the ways things work right now seems robust enough. I think it makes sense to wait and create issues once we have detected an actual problem

frontend/src/api/europepmc.js
mihai-sysbio: Good idea, my suggestion would be to have that as a new issue.

mihai-sysbio (approved):
👍🏻

frontend/src/api/europepmc.js
e0: I took another look at this implementation and realized that I was sloppy and misunderstood it before. I thought the data fetching was a part of the `try` block but the actual purpose seems to be to avoid the explicit handling of errors from data processing. For example, an error would be thrown if `details.authorList` would be `null` or `undefined`. This is bad practice indeed and I have created a new issue: https://github.com/MetabolicAtlas/MetabolicAtlas/issues/996

Sorry for not giving this more attention initially and thanks for the discussion!

frontend/src/store/modules/maps.js
```diff
@@ -109,7 +109,6 @@ const actions = {
         apiFunc = metabolitesApi.fetchMetaboliteData;
         break;
       default:
-        // TODO: handle unexpected type
         break;
```
nanjiangshu, line 112: What's the meaning to have a  `break` here when the default action is empty?
We can either delete
```
default:
  break;
```
or we assign the default value of apiFunc to e.g. `null`. 
But when `apiFunc` is undefined or null (which seems will never happen), the function `apiFunc` should not be called.

nanjiangshu (commented):
Looks great to me and I agree with comments from your previous discussions. Apart from that, I added a small comment about a probably unnecessary coding in a switch condition.

perjo (approved):

frontend/src/store/modules/maps.js
e0: You are right that a `break` is not needed here. Sometimes people leave it in the switch statements, such as [this example](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch#syntax).

The `apiFunc` variable must always be one of `gene`, `reaction`, or `metabolite` and it should always be called whenever the `getSelectedElement` function is called. There is some room for improvement here to make things clearer. Overall it felt like there are too many instances of this level of improvement in the project and this one in specific didn't warrant a dedicated issue.

nanjiangshu: Good to get verified that `apiFunc` will always be assigned a value. I'm ok to leave the code as is.

nanjiangshu (approved):

PR #1000, opened 2022-05-25
feat: display compartments as links for interaction partners

e0:
This PR closes https://github.com/MetabolicAtlas/private-issues/issues/109.

**Type of change**  

**List of changes made**  

- Refactor: extract out a component for rendering compartment links
- New functionality: display compartments as links for interaction partners table

**Screenshot of the fix**  

image

**Testing**  

- Instructions on how to test
  1. Visit an interaction partners page that contains a table of reactions at the bottom. For example: `/explore/Human-GEM/interaction-partners/MAM03177s`.
  2. Verify that the individual compartments are displayed as links and the links are working correctly.

**Further comments**  

The issue description states the following: "_As part of this work, it is expected that the final string gets constructed correctly from the beginning, instead of being fixed later. See https://github.com/MetabolicAtlas/MetabolicAtlas/pull/644 for more details._" I reviewed the comments from that issue and I'm not sure what this is referring to. Please let me know if anyone remembers and if anything should be changed.

### DIFF
--- a/api/src/neo4j/queries/interactionPartners.js
+++ b/api/src/neo4j/queries/interactionPartners.js
@@ -40,7 +40,9 @@ CALL apoc.cypher.mapParallel2("
  MATCH (reaction)-[cmE${v}]-(cm:CompartmentalizedMetabolite)
  WITH DISTINCT cm, cmE, reaction, compartments, subsystem, genes
  MATCH (c:Compartment)-[${v}]-(cm)-[${v}]-(:Metabolite)-[${v}]-(ms:MetaboliteState)
- RETURN {id: reaction.id, subsystem: subsystem, compartments: compartments, genes: genes, metabolites: COLLECT(DISTINCT({id: cm.id, name: ms.name,  compartmentId: c.id, outgoing: startnode(cmE)=cm})) } as data", {}, reactions, 50) YIELD value
+ WITH DISTINCT cm, cmE, reaction, compartments, subsystem, genes, c, ms
+ MATCH (reaction)-[${v}]-(rs:ReactionState)
+ RETURN {id: reaction.id, reversible: rs.reversible, subsystem: subsystem, compartments: compartments, genes: genes, metabolites: COLLECT(DISTINCT({id: cm.id, name: ms.name,  compartmentId: c.id, outgoing: startnode(cmE)=cm})) } as data", {}, reactions, 50) YIELD value
 
 WITH apoc.map.mergeList(apoc.coll.flatten(
   apoc.map.values(apoc.map.groupByMulti(COLLECT(value.data), "id"), [value.data.id])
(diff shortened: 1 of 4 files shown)


MalinAhlberg (approved):
I think this looks good, and it's nice with the code cleanup!

I'm also unsure what the section you mentioned in "Further comments" means (maybe @mihai-sysbio remembers?), but as this looks good to me as is, I approve.

nanjiangshu (commented):
It works like a charm and the code looks cleaner with refactoring.
I have just one point: why the order of the reactions are reversed compared to previous result? Is that any special reason for that?

mihai-sysbio:
@e0 the code changes in #644 are fixing the compartment names/links in `frontend`. The comment was meant to indicate that it would be more efficient to do this work in the backend/api, so that a more useful response can get cached by Nginx.

The issue also implied that we way links are constructed should be consistent in the different places in the website:
> I also want to see this because it would the same as the Reaction Table section in the GEM Browser.

I'm not sure if this was considered when working on this feature.

e0:
> It works like a charm and the code looks cleaner with refactoring.
> I have just one point: why the order of the reactions are reversed compared to previous result? Is that any special reason for that?

This was not an intended effect. I will need to have another look at this. Thanks for spotting it!

e0:
> @e0 the code changes in #644 are fixing the compartment names/links in `frontend`. The comment was meant to indicate that it would be more efficient to do this work in the backend/api, so that a more useful response can get cached by Nginx.

Can you please expand on exactly what you mean by this? The way I'm reading it it sounds like you want to construct the HTML with `<a>` tags in the backend. If that's the case I'm not seeing how it would be a more useful response. As I mentioned in this PR description, I have gone through all of the comments in #644 and didn't find anything that made me think of what you wrote here. Just a reminder that it would be nice in the future to try to include this type of info when creating new issues.

> The issue also implied that we way links are constructed should be consistent in the different places in the website:
> 
> > I also want to see this because it would the same as the Reaction Table section in the GEM Browser.
> 
> I'm not sure if this was considered when working on this feature.

This was indeed taken into consideration. It was a a strong motivation for the [refactoring work](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1000/commits/561d7e86a68cafc12fa99536e128c21e70332457). Basically it means this Interaction Partners table and the Reaction Table section in the GEM Browser are using the same component to render the compartment links. Is this not what you had in mind?

mihai-sysbio:
> The way I'm reading it it sounds like you want to construct the HTML with <a> tags in the backend. 

Kind of, yes - I imagine as much of the work to be done as early rationally possible so that it can be cached.

> Just a reminder that it would be nice in the future to try to include this type of info when creating new issues.

Definitely ➕ also a reminder for caveats when working on unrefined issues. Perhaps we should plan more backlog refinements 🤔 

> This was indeed taken into consideration. 

Good to know, I have not reviewed the work.

e0:
> I have just one point: why the order of the reactions are reversed compared to previous result? Is there any special reason for that?

I took a look at this now. For me the results are not reversed but they are different. I speculate this might be due to the fact that we have updated the models. The same reactions are still there so I don't think we have to look more into it for now. On a related note, I saw that `MAM03177s` is actually no longer available, `MAM03177c` works though and I should update the PR description with that.

nanjiangshu (approved):

Issue #1023, opened 2022-06-13
Deployment and database handling for EnzymeDB


nanjiangshu:
As a developer, I want the EnzymeDB to be deployed together with the rest of the website, using the same commands, so that I don't have to do anything extra (manually).

Acceptance criteria:
- [x] do research if anything is deployed according to (security) best practice
- [x] update the resource allocation in Docker compose files
- [x] double check database field types
- [x] when importing data, would it be possible to have something less hard-coded, eg '/input_data/enzymes/*.txt' ?

Ideas:
- use hashes to see if the database has changed and needs redeployment
- use a flag as an indicator for redeployment

If there is no apparent way of doing this automatically, it is okay to do this kinda manually. At a minimum, these should be a way to deploy this.

mihai-sysbio:
> For the remote system resources allocation, I assume that to start with we should assume that the total CPU and RAM remain the same right?

Indeed, we do not plan to add more resources.

> my local output for `docker stats`.

Please have a look in `prod` as well.

> taking half of the resource from `neo4j`

There are two types of resources allocated: `cpus` and `memory`. From what I see, `memory` is _over_-allocated for both `nginx` and `api`, so I propose to take a little bit from that as well. Regarding the `cpus` allocation, I guess most of it should come from `neo4j`, since many of the queries there will in the end be cached by `nginx`, but we can't expect that to be the case (to the same extent) for `pg`.

e0:
Thanks for the answer about the system resources @mihai-sysbio , it's very helpful.

I have been preparing the automatic database creation and testing it on my local machine. I think this setup should be enough for the MVP and I would like some feedback and discussion to validate that thought. This is the matching commit if you are interested: https://github.com/MetabolicAtlas/MetabolicAtlas/commit/19edf218dec51979e408573729040e0c935bb974

This is how it currently works:

1. During the build stage, copy the input data as well as the `init.sql` script to the docker image.
2. During deployment (or locally `start-stack`), the postgres data is mounted and persisted to a local volume if it currently isn't. 
3. After deployment, postgres checks the mounted volume and if the database has not been initialized, the `init.sql` script would run to initialize the database and import the data.

The advantage of this setup are:
- The build stage only takes marginally longer compared to before. The extra time is for copying the input data, which should be done within a minute given solid internet connections.
- Since the data import happens after all the containers have started, the rest of the stack runs normally as the data is importing.
 
[...]

PR #1016, opened 2022-06-08
feat: add PostgreSQL to the project for the enzyme database

e0:
**Related issue(s) and PR(s)**  
This PR closes #1008.

**Type of change**  

- [x] New feature (non-breaking change which adds functionality)
 
**List of changes made**  

- Add PostgreSQL to the project using docker.
- For the enzyme database, ddd scripts to initialize and seed the data.

**Testing**  

- A ~~rebuild~~restart (`start-stack`) is needed due to new dependencies
- Instructions on how to test
1. Update the `env-local.env` file and add the `POSTGRES_PASSWORD` environment variable. An example can be found in `env-local.env.sample`.
2. (Re)start the stack: `start-stack`. This should set up the new `pg` container
3. Verify that the PostgresQL container is up and running: `ma-exec pg psql -U postgres -c '\list'`.
4. Initialize the DB schema: `ma-exec pg psql -f scripts/init-schema.sql -U postgres`.
5. Add the input data files (currently incomplete). They were first posted [here](https://github.com/MetabolicAtlas/MetabolicAtlas/issues/1008#issuecomment-1144904853) and then [here](https://github.com/MetabolicAtlas/MetabolicAtlas/issues/1008#issuecomment-1148622936). Please use the `MainData` from the first link and the `supplOutput` data from the second link.
  i. Copy the contents of `MainData`, except for `README.txt`, into the `./pg/input_data/enzymes/` directory.
  ii. Copy the contents of `supplOutput` into the `./pg/input_data/supplementary/` directory. 
6. Seed the database using the input data: `ma-exec pg psql -f scripts/seed-data.sql -U postgres`. Edit: if this doesn't work please see [this comment](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1016#issuecomment-1149785950).
7. Verify that the database is setup correctly: `ma-exec pg psql -U postgres -c 'select * from enzymes limit 1'`. You can replace `enzymes` with `compounds`, `ec`, or `reactions` to test the other tables.

**Further comments**  

- A third script to generate random data for the `enzymes` table is available at `./pg/scripts/seed-mock-data.sql`. I used this to help test the performance of the database for larger dataset. More info can be found [here](https://github.com/MetabolicAtlas/MetabolicAtlas/issues/1008#issuecomment-1148729227).
- The input data currently available is a small subset of what the complete data should be. I think the following things would make more sense to include once the complete data is available.
[...]

PR #1030, opened 2022-06-13
feat: enzyme db details

### LINKED ISSUE #1014: implement page for displaying entity details and table for the enzyme database

As a user, I want to see relevant details for an entity (for example a reaction), so that I can have a quick overview of the entity and its related cross references.

As a user, I want to have access to a table of relevant fields associated with an entity (for example a reaction), so that I can browse through and understand which other entities and field values it is associated with.

**Acceptance Criteria**

- [x] If available, display the basic fields for an entity. This is the first table from the top in the attached UX mockup.
- [x] If available, display the cross references for an entity. This is the second table from the top in the attached UX mockup.
- [x] An interactive table should be displayed that lists all of of related entities and field values the entity is associated with.
  - [x] The table should be server-side paginated. Page size should be 50.
  - [x] Every column should be sortable.
  - [x] Every column, with the exception of `kcat`, should be filterable. Only filter on exact matches for now. Filtered results should still be server-side paginated.
- [x] Display an error message if no entity is found.
- [x] The URLs should follow the format: `/enzymedb/ENTITY_TYPE/ENTITY_ID`.
  - `enzymedb` is a working name, if should be updated if it gets changed by the researchers.
  - The exact entity types should be clear once the actual data files are ready.

**Additional context**

- This issue should follow the work from #1008.
- Mock data should be replaced with actual data once it's available.

UX mockup
image

e0:
This PR closes #1014 .

**Type of change**  

**List of changes made**  

- Update PostgreSQL schema.
- Add API endpoints for enzymeDB, `enzyme` as well as supplementary data `compound`, `ec`, `reaction`.
- Add page in frontend that displays the entity (`compound`, `ec`, or `reaction`) details as well as listing the related `enzymes`.
- Update the cross references page to accommodate for IDs that are not found in the `neo4j` database.

**Screenshot of the fix**  

image
image
image

**Testing**  

- Instructions on how to test

1. Visit an entity's details page. For example:
  - `/enzymedb/compound/C00002`
  - `/enzymedb/ec/2.5.1.18`
  - `/enzymedb/reaction/R00006`
2. Verify that the basic info and cross references (not available for `ec`) tables are displayed.
3. Verify that the enzymes table is displayed.
  - Please try out the filters (must be exact match), table sorting and pagination and verify that they work as expected.
4. Visit an entity's details page that does not exist (for example `/enzymedb/reaction/R000060`) and verify that and error pop-up is shown.

If you run into any problems with the database, please try resetting it. I also noticed that if you don't want any errors during the data import, it helps to delete the extra empty lines at the end of files. For me this was the `abac.txt` and `neq.txt` files.

To reset the database, please do a `git pull` and run the following commands.

```bash

ma-exec pg psql -f scripts/delete-db.sql -U postgres
ma-exec pg psql -f scripts/init-schema.sql -U postgres
ma-exec pg psql -f scripts/seed-data.sql -U postgres

```

**Further comments**  

### DIFF
--- a/api/src/endpoints/enzymeDb.js
+++ b/api/src/endpoints/enzymeDb.js
@@ -19,10 +19,10 @@ enzymeDbRoutes.get('/reactions/:id', async (req, res) => {
 });
 
 enzymeDbRoutes.get('/enzymes', async (req, res) => {
-  const { filters, page, pageSize } = req.query;
+  const { filters, pagination } = req.query;
 
   try {
-    const enzymes = await getEnzymes({ filters, page, pageSize });
+    const enzymes = await getEnzymes({ filters, pagination });
     if (enzymes) {
       res.json(enzymes);
     } else {
--- a/api/src/endpoints/enzymeDb.js
+++ b/api/src/endpoints/enzymeDb.js
@@ -1,0 +1,21 @@
+import express from 'express';
+import { getReaction } from 'enzymeDb/index';
+
+const enzymeDbRoutes = express.Router();
+
+enzymeDbRoutes.get('/reaction/:id', async (req, res) => {
+  const { id } = req.params;
+
+  try {
+    const reaction = await getReaction(id);
+    if (reaction) {
+      res.json(reaction);
+    } else {
+      res.sendStatus(404);
+    }
+  } catch (e) {
+    res.status(400).send(e.message);
+  }
+});
+
+export default enzymeDbRoutes;
--- a/api/src/endpoints/enzymeDb.js
+++ b/api/src/endpoints/enzymeDb.js
@@ -14,7 +14,7 @@ enzymeDbRoutes.get('/compounds/:id', async (req, res) => {
       res.sendStatus(404);
     }
   } catch (e) {
-    res.status(400).send(e.message);
+    res.status(400).send(escape(e.message));
   }
 });
 
@@ -29,7 +29,7 @@ enzymeDbRoutes.get('/ecs/:value', async (req, res) => {
       res.sendStatus(404);
     }
   } catch (e) {
-    res.status(400).send(e.message);
+    res.status(400).send(escape(e.message));
   }
 });
 
@@ -44,7 +44,7 @@ enzymeDbRoutes.get('/reactions/:id', async (req, res) => {
       res.sendStatus(404);
     }
   } catch (e) {
-    res.status(400).send(e.message);
+    res.status(400).send(escape(e.message));
   }
 });
 
@@ -59,7 +59,7 @@ enzymeDbRoutes.get('/enzymes', async (req, res) => {
       res.sendStatus(404);
     }
   } catch (e) {
-    res.status(400).send(e.message);
+    res.status(400).send(escape(e.message));
   }
 });
 
--- a/api/src/endpoints/enzymeDb.js
+++ b/api/src/endpoints/enzymeDb.js
@@ -14,7 +14,8 @@ enzymeDbRoutes.get('/compounds/:id', async (req, res) => {
       res.sendStatus(404);
     }
   } catch (e) {
-    res.status(400).send(escape(e.message));
+    console.error(e.message);
+    res.sendStatus(400);
   }
 });
 
@@ -29,7 +30,8 @@ enzymeDbRoutes.get('/ecs/:value', async (req, res) => {
       res.sendStatus(404);
     }
   } catch (e) {
-    res.status(400).send(escape(e.message));
+    console.error(e.message);
+    res.sendStatus(400);
   }
 });
 
@@ -44,7 +46,8 @@ enzymeDbRoutes.get('/reactions/:id', async (req, res) => {
       res.sendStatus(404);
     }
   } catch (e) {
-    res.status(400).send(escape(e.message));
+    console.error(e.message);
+    res.sendStatus(400);
   }
 });
 
@@ -59,7 +62,8 @@ enzymeDbRoutes.get('/enzymes', async (req, res) => {
       res.sendStatus(404);
     }
   } catch (e) {
-    res.status(400).send(escape(e.message));
+    console.error(e.message);
+    res.sendStatus(400);
   }
 });
 
--- a/api/src/endpoints/enzymeDb.js
+++ b/api/src/endpoints/enzymeDb.js
@@ -15,7 +15,7 @@ enzymeDbRoutes.get('/compounds/:id', async (req, res) => {
     }
   } catch (e) {
     console.error(e.message);
-    res.sendStatus(400);
+    res.sendStatus(500);
   }
 });
 
@@ -31,7 +31,7 @@ enzymeDbRoutes.get('/ecs/:value', async (req, res) => {
     }
   } catch (e) {
     console.error(e.message);
-    res.sendStatus(400);
+    res.sendStatus(500);
   }
 });
 
@@ -47,7 +47,7 @@ enzymeDbRoutes.get('/reactions/:id', async (req, res) => {
     }
   } catch (e) {
     console.error(e.message);
-    res.sendStatus(400);
+    res.sendStatus(500);
   }
 });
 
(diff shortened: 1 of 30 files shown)


mihai-sysbio:
Since I'm not seeing any explicit mention here, is it intended for this PR to also address [these comments](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1016#pullrequestreview-1000842668) or is that reserved for another PR? I think that was the plan discussed on Slack but now I'm not sure.

e0:
> Since I'm not seeing any explicit mention here, is it intended for this PR to also address [these comments](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1016#pullrequestreview-1000842668) or is that reserved for another PR? I think that was the plan discussed on Slack but now I'm not sure.

Thanks for the reminder @mihai-sysbio! I didn't plan on actually changing any files atm regarding those. I should have definitely addressed the comments though.

For the comment https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1016#discussion_r893161495 regarding remote docker compose config, I think it makes more sense to wait with this until the full dataset is provided and tested. Initially I thought maybe there was a chance that we could get the full dataset during the development of this issue. Now that we are wrapping this one up and there is #1023, I think it would make more sense to address it there.

For the comments https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1016#discussion_r893162619 and https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1016#discussion_r893162619 regarding the field types. I have adapted the fields to accommodate the current dataset that's available. We most certainly would need to change some of these later once the full dataset is provided. Please note that it seems that the reference fields, such as `rhea` and `chebi`, have values that are semicolon separated. These should not be capped in length.

For the comment https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1016#discussion_r893164801 regarding database import, I want to remember you mentioning that the full dataset would be provided in a single file as opposed to being divided up like they currently are. If that's the case I think it would make more sense to hard-code a single filename once that's available. If they end up being multiple files then your wildcard suggestion makes a lot of sense.

Overall, I think it's most efficient to wait with these changes until the full dataset is available so that hopefully we can handle them in a single iteration.

mihai-sysbio:
Thanks for the detailed response @e0, I am very happy with the suggested strategy.

MalinAhlberg:
Great work!
Some initial comments/questions:
- when visiting `/enzymedb/compound/C00003` and order by `ko`, columns move around a bit. Is it possible to make the width of the `Protein` column more fixed, so that the table doesn't change size as much? (maybe just me being annoyed by such things though)

- Kegg id `C00708` is present in eg `/Human-GEM/gem-browser/metabolite/MAM02174c`, but when visiting `/identifier/Kegg/C00708?referenceType=compound` it says "This database identifier is not used by any of the integrated models."

api/src/endpoints/enzymeDb.js
```diff
@@ -0,0 +1,70 @@
...
+  try {
+    const compound = await getCompound(id);
+    if (compound) {
+      res.json(compound);
+    } else {
+      res.sendStatus(404);
+    }
+  } catch (e) {
+    console.error(e.message);
+    res.sendStatus(400);
```
perjo, line 18: Seems strange to me to return 400 Bad Request here? The only thing passed in as id and if nothing is found using using that id it should be a 404 and if there are multiple ids for the id the data is bad, not the request? Seems more natural to me to return 500

e0:
I just added a new commit that requires resetting the database. Hopefully this helps solve the sorting problem on numeric values. I also noticed that if you don't want any errors during the data import, it helps to delete the extra empty lines at the end of files. For me this was the `abac.txt` and `neq.txt` files.

To reset the database, please do a `git pull` and run the following commands.

```bash

ma-exec pg psql -f scripts/delete-db.sql -U postgres
ma-exec pg psql -f scripts/init-schema.sql -U postgres
ma-exec pg psql -f scripts/seed-data.sql -U postgres

```

e0:
Just noticed that this latest version is buggy. I set this PR to draft now and will try to fix it before re-requesting. Sorry for the inconvenience.

api/src/endpoints/enzymeDb.js
e0: Really good point here @perjo. I think we have been a bit sloppy with the error codes. Do you think it's alright if we address this in a new issue? We are using `400` in some other places where changes could be needed too.

image

e0: I think `404` and `500` are really good and sensible suggestions. If we are changing though, I would still like to evaluate more possibilities of code (implementation) changes or alternative status codes. So I would prefer if we can have a bit more discussion about this before changing.

The main concern for me is to do what's convenient for collaborators. In general I would be thinking about the PR size and complexity. In this case since this issue is a blocker for a few others I think that's another good reason to continue this in another issue. Having said that, if the reviewers still think it's better to handle this here, or at least add the changes suggested by Pelle, that would be fine with me as well. Please let me know!

e0: Apologies again for not double checking earlier. I misunderstood the original situation and thought we only had `400` and that @perjo suggested to change the code to include both `404` and `500`. I took a look now and it seems that I had forgotten that we had already implemented the `404` bit, so since the only suggestion was to change `400` to `500`, that seems trivial enough to be included here.

I have added [a commit](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1030/commits/9ba0ffa31f8fe7e7d90882a6ed7c6f51cef6a016) for this now. It also includes a change for the `enzymes` endpoint. `404` does not make sense there since if there are 0 matches, it still returns a valid response that looks like `{"enzymes":[],"totalCount":"0"}`.

nanjiangshu (commented):
Very nice work @e0!

All pages and the API endpoints seem to work nicely when I tested. Below are some of my questions and comments

1. I have to manually create the psql database (with the scripts you provided) after `start-stack`. Is that only I got this problem or it is supposed to be like this since the database handling will be solved in the issue #1023 ? 

2. When filtering the table, I find that it is case sensitive, for example, when filtering the domain with the word `archaea`, the filtered table is empty. I'm not sure if this is ideal as a user. Maybe we should get some feedback from researchers.

3. When filtering the table, only the whole word matching will return some result, if we want to have this feature, I think a drop-down list would be more useful. Otherwise, it might be a bit frustrating when a user tried to filter a column but the table becomes empty once a letter is typed in. My impression is that most of these kind of sort tables filter by partial matching, that is, the size of the table reduces as more letters typed in. 

4. As I already mentioned at the walkthrough meeting, I feel the organism is redundant in the protein name since there is already a column with organism. Shall we create an issue to address this problem?

e0:
Thanks for the feedback @nanjiangshu!

> 1. I have to manually create the psql database (with the scripts you provided) after `start-stack`. Is that only I got this problem or it is supposed to be like this since the database handling will be solved in the issue #1023 ?

This needs to be done manually by everyone right now and I think #1023 is a good place to automate it. Sorry that I didn't make this clear.

___

> 2. When filtering the table, I find that it is case sensitive, for example, when filtering the domain with the word `archaea`, the filtered table is empty. I'm not sure if this is ideal as a user. Maybe we should get some feedback from researchers.

> 3. When filtering the table, only the whole word matching will return some result, if we want to have this feature, I think a drop-down list would be more useful. Otherwise, it might be a bit frustrating when a user tried to filter a column but the table becomes empty once a letter is typed in. My impression is that most of these kind of sort tables filter by partial matching, that is, the size of the table reduces as more letters typed in.

Great points. I think these could be included in one of the search related issues.

___

> 4. As I already mentioned at the walkthrough meeting, I feel the organism is redundant in the protein name since there is already a column with organism. Shall we create an issue to address this problem?

Agree and I think we had already expressed this in the initial discussion a couple of weeks ago. Would be nice to have this removed for the final data output.

mihai-sysbio:
> Would be nice to have this removed for the final data output.

The researchers are investigating if the org name can be safely removed from the protein ID.

inghylt (approved):
Very nice work! To me this is ready to be merged and improvements can be done in another issue

nanjiangshu (approved):
Looks good to me!

Issue #917, opened 2022-03-14
research `scoped` css


mihai-sysbio:
**Please describe the feature**

As a user, I want the interface to feel fast, so I don't get demotivated by using the website.
As a developer, I want to know what is technically the best choice when setting up styling for the project

**Describe the solution you would like**

Research should confirm if we need to remove all `scoped` CSS.

**Additional context**

The performance aspects might differ between Vue 2 and Vue 3.

e0:
According to [the official documentation](https://vuejs.org/api/sfc-css-features.html#scoped-css), there is no specific `scoped` styles performance penalties. The important takeaway is that `class` and `id` selectors are faster than standard HTML element selectors such as `div` or `p`. The `scoped` css does not add any `class` or `id` so we should keep in mind to always use these selectors. I found several instances of this (standard HTML element selectors in scoped style blocks) and have added a PR to address them.

mihai-sysbio:
> The important takeaway is that `class` and `id` selectors are faster than standard HTML element selectors such as `div` or `p`

Indeed, here's a direct link where this is [documented](https://vue-loader.vuejs.org/guide/scoped-css.html#deep-selectors):

> Scoped styles do not eliminate the need for classes. Due to the way browsers render various CSS selectors, `p { color: red }` will be many times slower when scoped (i.e. when combined with an attribute selector). If you use classes or ids instead, such as in `.example { color: red }`, then you virtually eliminate that performance hit.

PR #1040, opened 2022-06-21
feat: got enzymes range filters

### LINKED ISSUE #1026: Filter on kcat values of the EnzymeDB table
As a user, I want  to filter on the kcat column.

Acceptance criteria:
- [ ] min value filter
- [ ] max value filter

e0:
This PR closes #1026.

**Type of change**  

**List of changes made**  

- Add a range filter component an use it as a custom filter for `kcat` values in the enzymes table.

**Screenshot of the fix**  

image
image

**Testing**  

- Instructions on how to test
1. Go to a details page, for example `/gotenzymes/ec/2.5.1.19`.
2. Try different values for the `min` and `max` inputs for the `kcat` column.
3. Verify that the table is updated accordingly. This should also work in combination with other filters and column sorting.

**Further comments**  


mihai-sysbio:
I've tested and it seems to work as intended.

The input type doesn't seem to apply to Firefox, but it does work in Chrome. Not a deal breaker, but worth looking into if there is a Firefox-specific option.

Also, giving our international userbase, we should consider the case when people type in different decimal separators, `,` instead of `.`. One way to deal with this could be to apply an `is-warning` class when the typed value is not a finite number. Any thoughts?

e0:
> The input type doesn't seem to apply to Firefox, but it does work in Chrome. Not a deal breaker, but worth looking into if there is a Firefox-specific option.

Do you mean the little arrows to increment or decrement. If so, it seems to be working for me with Firefox on macOS.

image

___ 

> Also, giving our international userbase, we should consider the case when people type in different decimal separators, `,` instead of `.`. One way to deal with this could be to apply an `is-warning` class when the typed value is not a finite number. Any thoughts?

The number input does not deal with this well. I looked this up briefly now and I think I would need to do more extensive research and testing if we want to address this. Not only do we have to make sure that the localization is handled in all browsers, we also have to make sure to convert it back to a `.` as separator for running the SQL query. Would you like us to continue with in the current issue or create a new issue for it?

mihai-sysbio:
> Do you mean the little arrows to increment or decrement. If so, it seems to be working for me with Firefox on macOS.

Sorry for not being clear, the input wasn't restricted to `number` for me - I could type letters as well.

> The number input does not deal with this well. I looked this up briefly now and I think I would need to do more extensive research and testing if we want to address this. Not only do we have to make sure that the localization is handled in all browsers, we also have to make sure to convert it back to a `.` as separator for running the SQL query. Would you like us to continue with in the current issue or create a new issue for it?

To solve this at the extent you're suggesting sounds indeed quite intensive. Is there an easy enough way to indicate to the user that the input is "wrong" (not interpreted as a valid number)? Otherwise I suggest simply dropping this aspect completely until it resurfaces again, and then we will create an issue.

e0:
> Is there an easy enough way to indicate to the user that the input is "wrong" (not interpreted as a valid number)? 

This should be relatively easy to handle as long as we define what is a valid number. Is this regex ok? `/^\-?\d+\.*\d*$/`

Examples:
image

mihai-sysbio:
That looks good to me @e0, and it addresses my test cases in FF (below) _and_ Safari where one can type letters.

https://user-images.githubusercontent.com/23480589/175477479-6dd26c76-edcf-4a7b-b8a6-5fb78c806617.mp4

.

MalinAhlberg:
Good work!

I experience the same problem as Mihail, but that should be fixed by your suggested fix @e0 . But why not `/^\-?\d+\.?\d*$/` (`\.?` instead of `\.*`) so that we don't allow `1....1`?

e0:
> I experience the same problem as Mihail, but that should be fixed by your suggested fix @e0 . But why not `/^\-?\d+\.?\d*$/` (`\.?` instead of `\.*`) so that we don't allow `1....1`?

Thanks @MalinAhlberg! That makes sense and was something I overlooked.

frontend/src/components/shared/RangeFilter.vue
```diff
@@ -0,0 +1,87 @@
...
+    },
+    max(newValue, oldValue) {
+      if (newValue !== oldValue) {
+        this.handleUpdate(this.rangePayload);
+      }
+    },
+  },
+  methods: {
+    inputValid(maybeNumber) {
+      if (maybeNumber === null) {
```
MalinAhlberg, line 72:
```suggestion
      if (maybeNumber === null || maybeNumber === "") {
```
For me, in Firefox, this is needed to avoid the field getting marked in red when deleting the current input

MalinAhlberg (approved):
Great!

I added one more comment, but nothing stopping my approval :rocket:

frontend/src/components/shared/RangeFilter.vue
```diff
@@ -0,0 +1,87 @@
...
+      }
+
+      const numberRegex = /^-?\d+\.?\d*$/;
+      return maybeNumber.toString().match(numberRegex);
+    },
+  },
+};
+</script>
+
+<style lang="scss" scoped>
```
mihai-sysbio, line 83: A minor note about `scoped`, which might not be advised for future use, depending on the research in #917 .

mihai-sysbio (approved):
Looks and works great!

frontend/src/components/shared/RangeFilter.vue
mihai-sysbio: Ah right, I intended to leave the resolving to you, but I should have mentioned that.

e0: Thanks! Yea that would be nice in the future as by default it's the reviewer's responsibility.

Issue #1034, opened 2022-06-16
Update About the project section with missing information


inghylt:
Acceptance criteria:

- [x] update text on the homepage
- [x] update table header to: k<sub>cat</sub>[1/s]

mihai-sysbio:
Note: it was not possible to have the subscript in the column header, not even with `html: true` as a column property.

PR #1043, opened 2022-06-22
feat: add more details pages for GotEnzymes

MalinAhlberg:
**Related issue(s) and PR(s)**  
This PR closes #1032.

Joint work with @e0.

**Type of change**  

- [x] New feature (non-breaking change which adds functionality)
 
**List of changes made**  

- Add details pages for `protein`, `organism`, and `domain`.
- Update landing page to accommodate for the new details pages (search and examples).
- Hide column of current page's entity type in the enzymes table.
- Turn table cells from columns/fields that have details pages into links.

**Screenshot of the fix**  

image
image

**Testing**  

- Relative urls that can be reused both for production and local testing
- Instructions on how to test
1. Test handling of new details pages from the landing page `/gotenzymes`. Please try some searches as well as clicking on the provided example links.
2. Verify that the page loads without problems for pages that do not have basic info or cross references, for example `/gotenzymes/organism/hsa`.
3. Verify that the column matching the current details page, for example the `Organism` column on `/gotenzymes/organism/hsa`, is not shown.
4. Verify that the links from the enzymes tables are working correctly. Please try all 6 types.

**Further comments**  

**Definition of Done checklist**  
- [ ] My code meets the Acceptance Criteria (no new tests added)
- [X] My code follows the [NBIS style guidelines](https://github.com/NBISweden/development-guidelines)
- [X] I have performed a self-review of my own code and commented any hard-to-understand areas
- [X] Tests and lint/format validations are passing
- [X] My changes generate no new warnings

PR #1031, opened 2022-06-14
feat: enzyme landing page

MalinAhlberg:
**Related issue(s) and PR(s)**  
This PR closes #1015 and relates to #1014.

This is joint work with @e0 .

**Type of change**  

- [X] New feature (non-breaking change which adds functionality)

 
**List of changes made**  

- Added the page `/gotenzymes`
- Added "GotEnzymes" to the navbar
- Added search field and example search strings
- Added about text

**Screenshot of the fix**  

image

**Testing**  

- Navigate to `/gotenzymes` and try out the provided example searches
- Please also try other searches. You may search for EC, reaction and compound ids. Examples: "C00002", "2.5.1.18", "R00006"
- Searching for non-valid strings:
  -  reaction like, compound like or ec like strings (eg "RRR", "CCC", "2222.2.2.2") should send you to the respective page, with an error message saying that nothing was found.
  - for other search strings (eg "xxxx"), "No matches for xxxx" will be shown"
- Verify that the about text section looks ok and that the links work.

**Further comments**  

This will be undrafted when #1030  is merged.

**Definition of Done checklist**  
- [x] My code meets the Acceptance Criteria
- [x] My code follows the [NBIS style guidelines](https://github.com/NBISweden/development-guidelines)
- [x] I have performed a self-review of my own code and commented any hard-to-understand areas
- [x] Tests and lint/format validations are passing
- [x] My changes generate no new warnings

PR #1045, opened 2022-06-23
chore: update documentation for GotEnzymes

mihai-sysbio:
This PR updates the documentation for GotEnzymes thus closing #1034.

edit: I also spotted a tiny bug on the layout of the _Landing_ page, it had to do with a missing class, so I opted to commit it to the same PR rather than having a separate one. Since the page is copied from `Home.vue`, it was easy to figure out the missing piece.

**Type of change**  

- [x] Other: documentation

**Testing**  

- Instructions on how to test: go to the landing page

**Further comments**  

### DIFF
--- a/frontend/src/components/gotEnzymes/Landing.vue
+++ b/frontend/src/components/gotEnzymes/Landing.vue
@@ -1,5 +1,5 @@
 <template>
-  <div>
+  <div class="extended-section">
     <section class="hero is-primary is-bold py-6">
       <div class="hero-body has-text-centered">
         <p class="is-size-1 title">GotEnzymes</p>
--- a/frontend/src/components/gotEnzymes/Landing.vue
+++ b/frontend/src/components/gotEnzymes/Landing.vue
@@ -84,7 +84,8 @@
               <hr class="mt-6" />
               <h4 id="how-to-use-it" class="is-info is-size-4">How to use it</h4>
               <p>
-                We currently support searches by reaction, compound, and EC number. Each predicted
+                We currently support fuzzy searches by reaction, compound, and EC number, organism
+                and domain, and exact searches for KEGG protein identifiers. Each predicted
                 parameter is filterable. Feel free to contact us if you find any issue. To report a
                 technical error on the webpage please email
                 <a href="mailto:contact@metabolicatlas.org">contact [at] metabolicatlas [dot] org</a
--- a/frontend/src/components/gotEnzymes/Landing.vue
+++ b/frontend/src/components/gotEnzymes/Landing.vue
@@ -102,7 +102,7 @@
                   href="https://doi.org/10.1038/s41929-022-00798-z"
                   >doi:10.1038/s41929-022-00798-z</a
                 >). Our vision for this database is to store every possible predicted parameter for
-                any available enzyme to facilitate to facilitate computational applications such as
+                any available enzyme to facilitate computational applications such as
                 flux simulations, and for use in metabolic engineering.
               </p>
 
--- a/frontend/src/components/gotEnzymes/Landing.vue
+++ b/frontend/src/components/gotEnzymes/Landing.vue
@@ -106,7 +106,7 @@
                 >). Our vision for GotEnzymes is to facilitate computational applications, such as
                 flux simulations, and to improve the Design-Build-Test-Learn cycle in metabolic
                 engineering, by suggesting candidate alternative enzymes. To this end, we aim to
-                store every possible predicted parameter for all available enzyme in a single
+                store every possible predicted parameter for all available enzymes in a single
                 database that is equally easy to use manually and programatically.
               </p>
 
--- a/frontend/src/components/gotEnzymes/Landing.vue
+++ b/frontend/src/components/gotEnzymes/Landing.vue
@@ -47,7 +47,7 @@
       <div class="columns is-centered pb-6">
         <div class="column is-three-fifths-desktop is-three-quarters-tablet is-fullwidth-mobile">
           <div class="content">
-            <h5 class="title is-5">With KEGG ids or EC codes, you can search for a:</h5>
+            <h5 class="title is-5">One can search for a:</h5>
             <ul>
               <li>
                 reaction (e.g.
--- a/frontend/src/components/gotEnzymes/Landing.vue
+++ b/frontend/src/components/gotEnzymes/Landing.vue
@@ -84,10 +84,11 @@
               <hr class="mt-6" />
               <h4 id="how-to-use-it" class="is-info is-size-4">How to use it</h4>
               <p>
-                We currently support fuzzy searches by reaction, compound, EC number, organism
-                and domain, and exact searches for KEGG protein identifiers. Each predicted
-                parameter is filterable. Feel free to contact us if you find any issue. To report a
-                technical error on the webpage please email
+                We currently support fuzzy searches by reaction, compound, EC number, organism and
+                domain, and exact searches for KEGG gene identifiers. All table columns are
+                filterable and sortable, including the predicted numberical values. Feel free to
+                contact us if you find any issue. To report a technical error on the webpage please
+                email
                 <a href="mailto:contact@metabolicatlas.org">contact [at] metabolicatlas [dot] org</a
                 >. For questions about the scientific methods, please reach out to
                 <a href="mailto:gotenzymes@metabolicatlas.org"
@@ -97,19 +98,23 @@
               <hr class="mt-6" />
               <h4 id="how-it-was-made" class="is-info is-size-4">How it was made</h4>
               <p>
-                Data in this database are predicted from available advanced deep learning tools:
-                k<sub>cat</sub> published in <i>Nature Catalysis</i> in 2022 (<a
+                The data in this database are predicted with available advanced deep learning tools,
+                such as <i>DLKcat</i> for k<sub>cat</sub> prediction published in
+                <i>Nature Catalysis</i> in 2022 (<a
                   href="https://doi.org/10.1038/s41929-022-00798-z"
                   >doi:10.1038/s41929-022-00798-z</a
-                >). Our vision for this database is to store every possible predicted parameter for
-                any available enzyme to facilitate computational applications such as
-                flux simulations, and for use in metabolic engineering.
+                >). Our vision for GotEnzymes is to facilitate computational applications, such as
+                flux simulations, and to improve the Design-Build-Test-Learn cycle in metabolic
+                engineering, by suggesting candidate alternative enzymes. To this end, we aim to
+                store every possible predicted parameter for all available enzyme in a single
+                database that is equally easy to use manually and programatically.
               </p>
 
               <hr class="mt-6" />
               <h4 id="citation" class="is-info is-size-4">Citation</h4>
               <p>
-                <i>Manuscript under consideration</i>. For more information about the k<sub>cat</sub
+                <i>Manuscript under consideration</i>. For more information about the k<sub
+                  >cat</sub
                 >
                 prediction methods, please have a look at
                 <i
--- a/frontend/src/components/gotEnzymes/Landing.vue
+++ b/frontend/src/components/gotEnzymes/Landing.vue
@@ -77,7 +77,7 @@
             <div id="intro" class="column content has-text-justified">
               <p>
                 GotEnzymes provides open access to over 25,794,195 million predicted k<sub>cat</sub>
-                for 5,825,213 unique sequences and 4,147 compounds from 8,099s species with the
+                for 5,825,213 unique sequences and 4,147 compounds from 8,099 species with the
                 cutting-edge artificial intelligence tools.
               </p>
 
--- a/frontend/src/components/gotEnzymes/Landing.vue
+++ b/frontend/src/components/gotEnzymes/Landing.vue
@@ -47,7 +47,7 @@
       <div class="columns is-centered pb-6">
         <div class="column is-three-fifths-desktop is-three-quarters-tablet is-fullwidth-mobile">
           <div class="content">
-            <h5 class="title is-5">You can search for a:</h5>
+            <h5 class="title is-5">With KEGG ids or EC codes, you can search for a:</h5>
             <ul>
               <li>
                 reaction (e.g.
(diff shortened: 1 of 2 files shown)


frontend/src/components/gotEnzymes/Landing.vue
```diff
@@ -47,7 +47,7 @@
       <div class="columns is-centered pb-6">
         <div class="column is-three-fifths-desktop is-three-quarters-tablet is-fullwidth-mobile">
           <div class="content">
-            <h5 class="title is-5">You can search for a:</h5>
+            <h5 class="title is-5">With KEGG ids or EC codes, you can search for a:</h5>
```
e0, line 50: Is the `organism` value a KEGG id? For example on [this page](https://www.genome.jp/entry/gn:T03489), the value we have stored is `bmae` and the **Entry** listed on the linked page is actually `T03489`. It seems that the **Entry** field is the actual KEGG id, like on [this reaction page](https://www.genome.jp/entry/R03050).

Similarly for `domain`, it doesn't look like a KEGG id either.

frontend/src/components/gotEnzymes/Landing.vue
```diff
@@ -76,18 +76,15 @@
             <TableOfContents :links="tocLinks" />
             <div id="intro" class="column content has-text-justified">
               <p>
-                GotEnzymes provides open access to over XYZ million predicted kinetic data points
-                for XYZ unique sequences and XYZ compounds from XYZ species with the cutting-edge
-                artificial intelligence tools. The values for kcat [reference] and KM [reference] in
-                this database are predicted from available advanced deep learning tools. Our vision
-                for this database is to store every possible predicted parameter for any available
-                enzyme to facilitate better science.
+                GotEnzymes provides open access to over 25,794,195 million predicted k<sub>cat</sub>
```
e0, line 79: A bit curious about this since we have >33 million rows in the `enzymes` table. If I count unique values (`select count(*) from enzymes group by kcat_values;`), we have ~1.8 million. It would be interesting to get some background about where this listed number comes from.

frontend/src/components/gotEnzymes/Landing.vue
```diff
@@ -76,18 +76,15 @@
...
+                for 5,825,213 unique sequences and 4,147 compounds from 8,099s species with the
+                cutting-edge artificial intelligence tools.
               </p>
 
               <hr class="mt-6" />
               <h4 id="how-to-use-it" class="is-info is-size-4">How to use it</h4>
               <p>
-                We currently support searches by reaction, compound, EC number. Each predicted
+                We currently support searches by reaction, compound, and EC number. Each predicted
```
e0, line 87: Just a note that we should update #1043 with changing this line. Either that or we can update directly in this PR, depending on which one gets merged first.

e0 (commented):
Looks good overall. Unsure if anything should be changed (more details in inline comments) so not approving just yet.

Minor general feedback, I think we usually would use the `docs` prefix for the branch and commits for things like this.

frontend/src/components/gotEnzymes/Landing.vue
feiranl: I counted the unique proteinID&compund pairs. They may contain identical values, which is due to that protein sequences from different organisms are identifical.

frontend/src/components/gotEnzymes/Landing.vue
e0: Thanks for the additional context @Yu-sysbio. My comment was mostly about the wording of the title `With KEGG ids or EC codes`. Just wondering if it should be changed since technically it's not just KEGG id and EC codes that we can search for.

frontend/src/components/gotEnzymes/Landing.vue
e0: Ah I see. Thank you for taking time to explain!

frontend/src/components/gotEnzymes/Landing.vue
Yu-sysbio: It is true that `KEGG id` (e.g. T03489) seems technically better than `Org code` (e.g. bmae) to search for an organism, but currently let us just not change since we are using `Org code` in our database. Normally, users would most likely search the full `Name` (e.g. Burkholderia mallei 6) for an organism, which we expect to be a future feature.

e0: Sorry that I was still unclear about this. I'm not suggesting to change anything in the data, I'm only talking about changing this line:

```html
<h5 class="title is-5">With KEGG ids or EC codes, you can search for a:</h5>
```

mihai-sysbio: There is an issue to improve the search, which would make this reference to KEGG ids obsolete, so I'm proposing to go back to the previous line:
```suggestion
            <h5 class="title is-5">One can search for a:</h5>
```

e0 (approved):

frontend/src/components/gotEnzymes/Landing.vue
```diff
@@ -99,24 +96,27 @@
...
-                Finibus Bonorum et Malorum" (The Extremes of Good and Evil) by Cicero, written in 45
-                BC. This book is a treatise on the theory of ethics, very popular during the
-                Renaissance. The first line of Lorem Ipsum, "Lorem ipsum dolor sit amet..", comes
-                from a line in section 1.10.32.
+                Data in this database are predicted from available advanced deep learning tools:
+                k<sub>cat</sub> published in <i>Nature Catalysis</i> in 2022 (<a
+                  href="https://doi.org/10.1038/s41929-022-00798-z"
+                  >doi:10.1038/s41929-022-00798-z</a
+                >). Our vision for this database is to store every possible predicted parameter for
+                any available enzyme to facilitate
```
MalinAhlberg, line 104: Should this sentence really end here..?

MalinAhlberg (commented):
Nice! I added one question, apart from that I think everything looks great.

frontend/src/components/gotEnzymes/Landing.vue
```diff
@@ -76,18 +76,15 @@
...
             <div id="intro" class="column content has-text-justified">
               <p>
-                GotEnzymes provides open access to over XYZ million predicted kinetic data points
-                for XYZ unique sequences and XYZ compounds from XYZ species with the cutting-edge
-                artificial intelligence tools. The values for kcat [reference] and KM [reference] in
-                this database are predicted from available advanced deep learning tools. Our vision
-                for this database is to store every possible predicted parameter for any available
-                enzyme to facilitate better science.
+                GotEnzymes provides open access to over 25,794,195 million predicted k<sub>cat</sub>
+                for 5,825,213 unique sequences and 4,147 compounds from 8,099s species with the
```
MalinAhlberg, line 80:
```suggestion
                for 5,825,213 unique sequences and 4,147 compounds from 8,099 species with the
```

frontend/src/components/gotEnzymes/Landing.vue
mihai-sysbio: Great that you caught this @MalinAhlberg. I think the original was `to facilitate better science`, but I think it's time to address @inghylt's previous comment [here](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1031#discussion_r898955659).

```suggestion
                any available enzyme to facilitate improved predictions.
```

What do you think?

MalinAhlberg: The new version sounds really nice! The old one is also fine to me.

feiranl: It is not improved predictions. We can skip the part 'to facilitate...', just `our vision is to store every possible predicted parameter for any available enzyme`

frontend/src/components/gotEnzymes/Landing.vue
mihai-sysbio: The suggestion above was committed in [512597d](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1045/commits/512597d229f070d351b68b95b19246a8de4baa48).

MalinAhlberg (approved):
Looks nice!
I left a comment about some duplicated words, apart from this I think it's all good!

mihai-sysbio:
> I left a comment about some duplicated words, apart from this I think it's all good!

😓 sorry for being so sloppy with this, thanks for the approval

frontend/src/components/gotEnzymes/Landing.vue
```diff
@@ -76,18 +76,16 @@
...
+                GotEnzymes provides open access to over 25,794,195 million predicted k<sub>cat</sub>
+                for 5,825,213 unique sequences and 4,147 compounds from 8,099 species with the
+                cutting-edge artificial intelligence tools.
               </p>
 
               <hr class="mt-6" />
               <h4 id="how-to-use-it" class="is-info is-size-4">How to use it</h4>
               <p>
-                We currently support searches by reaction, compound, EC number. Each predicted
+                We currently support fuzzy searches by reaction, compound, and EC number, organism
```
nanjiangshu, line 87:
```suggestion
                We currently support fuzzy searches by reaction, compound,  EC number, organism
```
The `and` before `EC number` should be removed after addition of `organism and domain`

frontend/src/components/gotEnzymes/Landing.vue
```diff
@@ -76,18 +76,16 @@
...
+                and domain, and exact searches for KEGG protein identifiers. Each predicted
```
nanjiangshu, line 88: All parameters are filterable, not only predicted ones. I suggest to remove the word `predicted`.

frontend/src/components/gotEnzymes/Landing.vue
```diff
@@ -99,24 +97,28 @@
...
-                piece of classical Latin literature from 45 BC, making it over 2000 years old.
-                Richard McClintock, a Latin professor at Hampden-Sydney College in Virginia, looked
-                up one of the more obscure Latin words, consectetur, from a Lorem Ipsum passage, and
-                going through the cites of the word in classical literature, discovered the
-                undoubtable source. Lorem Ipsum comes from sections 1.10.32 and 1.10.33 of "de
-                Finibus Bonorum et Malorum" (The Extremes of Good and Evil) by Cicero, written in 45
-                BC. This book is a treatise on the theory of ethics, very popular during the
-                Renaissance. The first line of Lorem Ipsum, "Lorem ipsum dolor sit amet..", comes
-                from a line in section 1.10.32.
+                Data in this database are predicted from available advanced deep learning tools:
```
nanjiangshu, line 100: This sounds like the method is called k<sub>cat</sub> but it's not. What about we add a citation like format with link to the publication?
e.g. 
                Data in this database are predicted from available advanced deep learning tools ([Li et al. Nature Catalysis](https://doi.org/10.1038/s41929-022-00798-z))

frontend/src/components/gotEnzymes/Landing.vue
```diff
@@ -99,24 +97,28 @@
...
+                k<sub>cat</sub> published in <i>Nature Catalysis</i> in 2022 (<a
+                  href="https://doi.org/10.1038/s41929-022-00798-z"
+                  >doi:10.1038/s41929-022-00798-z</a
+                >). Our vision for this database is to store every possible predicted parameter for
+                any available enzyme to facilitate computational applications such as
+                flux simulations, and for use in metabolic engineering.
```
nanjiangshu, line 106:
```suggestion
                flux simulations and metabolic engineering.
```

frontend/src/components/gotEnzymes/Landing.vue
```diff
@@ -99,24 +97,28 @@
...
               </p>
 
               <hr class="mt-6" />
               <h4 id="citation" class="is-info is-size-4">Citation</h4>
               <p>
-                To cite GotEnzymes please use this [link]. For more information about the prediction
-                methods, have a look at reference title [link] for kcat prediction and KM
-                prediction.
+                <i>Manuscript under consideration</i>. For more information about the k<sub>cat</sub
+                >
```
nanjiangshu, line 113: I fee it's a bit duplicated to use this publication again since it has been referred in the previous paragraph already. Moreover, this section is about `citation` and I guess it should be the manuscript to NAR that should be cited for GotEnzymes, right?

frontend/src/components/gotEnzymes/Landing.vue
mihai-sysbio: Good point @nanjiangshu, I just had to remove the extra space as well
```suggestion
                We currently support fuzzy searches by reaction, compound, EC number, organism
```

frontend/src/components/gotEnzymes/Landing.vue
mihai-sysbio: To people in the field, k<sub>cat</sub> is a known term. It's also easily googleable. I'd rather suggest something along the lines of `prediction of k<sub>cat</sub> with _DLKcat_ published in ...`.

frontend/src/components/gotEnzymes/Landing.vue
mihai-sysbio: The above suggestion can generate more confusion, since metabolic engineering is not a type of computational application; perhaps it would be clearer to add an `also` after `and`.

frontend/src/components/gotEnzymes/Landing.vue
mihai-sysbio: Let's see where this gets published first, and then we can remove the duplicated DLKcat reference here.

frontend/src/components/gotEnzymes/Landing.vue
nanjiangshu: Then probably separate them in two sentences.

frontend/src/components/gotEnzymes/Landing.vue
nanjiangshu: That sounds much better.

frontend/src/components/gotEnzymes/Landing.vue
mihai-sysbio: Is this what you mean @nanjiangshu ?
```
Our vision for GotEnzymes is to facilitate computational applications, such as
flux simulations, and to improve the Design-Build-Test-Learn cycle in metabolic engineering,
by suggesting candidate alternative enzymes. To this end, we aim to store every
possible predicted parameter for all available enzyme in a single database that is
equally easy to use manually and programatically.
```

Also pinging @feiranl and @Yu-sysbio .

nanjiangshu: Looks much better now.

frontend/src/components/gotEnzymes/Landing.vue
e0: Should this be `gene` identifiers since it has been changed to that everywhere else?

mihai-sysbio: > Should this be `gene` identifiers since it has been changed to that everywhere else?

Absolutely!

frontend/src/components/gotEnzymes/Landing.vue
mihai-sysbio: The above suggestion was applied in [ebcff15](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1045/commits/ebcff150ae5c301f9d16157ee98aba7d6c0a80cd).

mihai-sysbio:
Thanks for the detailed review @nanjiangshu. I hope to have addressed all the comments.

frontend/src/components/gotEnzymes/Landing.vue
```diff
@@ -99,24 +98,32 @@
...
-                from a line in section 1.10.32.
+                The data in this database are predicted with available advanced deep learning tools,
+                such as <i>DLKcat</i> for k<sub>cat</sub> prediction published in
+                <i>Nature Catalysis</i> in 2022 (<a
+                  href="https://doi.org/10.1038/s41929-022-00798-z"
+                  >doi:10.1038/s41929-022-00798-z</a
+                >). Our vision for GotEnzymes is to facilitate computational applications, such as
+                flux simulations, and to improve the Design-Build-Test-Learn cycle in metabolic
+                engineering, by suggesting candidate alternative enzymes. To this end, we aim to
+                store every possible predicted parameter for all available enzyme in a single
```
nanjiangshu, line 109:
```suggestion
                store every possible predicted parameter for all available enzymes in a single
```

nanjiangshu (commented):
Looks great except for a single missing plural 😄

PR #1051, opened 2022-06-28
feat: quick search for GotEnzymes

### LINKED ISSUE #1027: Implement Quick Search for the EnzymeDB landing page

As a user, I want the same search functionality as in the GEM browser so that I can see suggestions even for partial strings.

**Acceptance Criteria**

- [x] 5-10 suggestions maximum
- [x] Should be reasonably fast
- [x] The changes are in alignment with the privacy policy
- [x] The changes are tested on multiple models and browsers
- [x] The API response is the same as before
- [x] The changes do not reduce the performance
- [x] The web page documentation is updated accordingly
- [x] The changes do not affect the deployment

**Additional context**

Re-use as much as possible from the existing quick search.
Does not have to be grouped by type.
Only one character is needed to search for domains, maybe remove the restriction of 2 characters totally.

If possible, this could solve issue #1028.

e0:
This PR closes #1027.

**Type of change**  

**List of changes made**  

- Add quick search to the GotEnzymes landing page, including a new search endpoint in the `api`.
- Update the database initialization script.
  - Add new indexes and views to enable fast fuzzy searching.
  - Optimize data import speed (faster to add data before creating indexes).

**Screenshot of the fix**  

image

**Testing**  

- Instructions on how to test
1. Reconstruct the database: `ma-exec pg psql -f scripts/init.sql -U postgres`. (This should be slightly faster than before, took about 7.5 min vs ~10min from before for me.)
2. Visit the GotEnzymes landing page: `/gotenzymes`.
3. Try different searches and try to return results for each of the page types.
  - Verify that clicking a result row goes to the details page.
  - Verify that a "No matches ..." message is displayed when there are no matching results for the search term, for example: `qpowqej`.

**Further comments**  

There weren't any sketches for this so I took some liberty in design choices, such as including a search results highlighter. Please let me know if you have any feedback on what could be changed.

### DIFF
--- a/frontend/src/helpers/utils.js
+++ b/frontend/src/helpers/utils.js
@@ -215,14 +215,6 @@ export function sanitizeSearchString(term, isAddBackSlash = true) {
   return newTerm;
 }
 
-export const sanitizeSearchTerm = term => {
-  if (!term) {
-    return '';
-  }
-
-  return sanitizeSearchString(term).replace(/#/g, '');
-};
-
 export const combineWords = ({ items, itemType }) => {
   let combined;
 
--- a/frontend/src/store/modules/gotEnzymes.js
+++ b/frontend/src/store/modules/gotEnzymes.js
@@ -1,5 +1,4 @@
 import gotEnzymesApi from '@/api/gotEnzymes';
-import { sanitizeSearchTerm } from '@/helpers/utils';
 
 const data = {
   info: {},
@@ -38,7 +37,8 @@ const actions = {
     commit('setCrossReferences', []);
   },
   async search({ commit }, searchTerm) {
-    const term = sanitizeSearchTerm(searchTerm);
+    const term = encodeURIComponent(searchTerm);
+
     if (term.length === 0) {
       commit('setSearchResults', []);
       return;


MalinAhlberg:
This is really nice, well done @e0!
I think the design is looks good and it's extremely useful. I ran into some problems when searching:
- Searching for some genes, eg `100000859`, really takes long time (unless cached).
- Searching for `R0` gives no results, but if you type one more letter (`R01`) you get results. Just giving one letter (`R`) also gives results.
- Searching for the gene `ZZ6_1708` gives no results, although it exist (see https://dev-metatlas.csbi.chalmers.se/gotenzymes/gene/ZZ6_1708 and https://dev-metatlas.csbi.chalmers.se/gotenzymes/organism/zmb). Maybe there's a timeout?

MalinAhlberg:
> For the `R0` case I guess that's a case where things are working "correctly" but may seem weird.

But shouldn't `R0` gives suggestions of for example R01701, R01301,R01201..?

e0:
> * Searching for the gene `ZZ6_1708` gives no results, although it exist (see https://dev-metatlas.csbi.chalmers.se/gotenzymes/gene/ZZ6_1708 and https://dev-metatlas.csbi.chalmers.se/gotenzymes/organism/zmb). Maybe there's a timeout?

This has been addressed in https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1051/commits/1d942e689b36fd1690f595523b1baf4f5338adbc now.

e0:
> > For the `R0` case I guess that's a case where things are working "correctly" but may seem weird.
> 
> But shouldn't `R0` gives suggestions of for example R01701, R01301,R01201..?

Yea sorry I should have said that by the way `pg_trgm` and fuzzy search (using `%`) are currently setup, by default it doesn't seem to return matches that are not close enough to the result. So I guess they classify `R0` as being too dissimilar to `R01701`. I will see if there is someway to customize this though as it could make more sense for our users to display some results.

e0:
@MalinAhlberg `R0` should show results now after https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1051/commits/33828926ad69da2a234f7ba062ec7bc868b04b95.

Regarding the `100000859` issue, I couldn't replicate the problem after a fresh deploy (should be uncached). When deploying to the dev server, I ran into some problems and saw that the disk space was running out. I'm not sure if this is related to some thing being slow before. In general this type of problem happens when things are not indexed correctly. I suspect that this could be a hardware resources related issue but I'm not certain. If we encounter more cases like this, I think we could try assigning the `pg` container more resources. Perhaps it's something we could look into in a new issue unless the problem is happening too frequently atm.

mihai-sysbio:
> I ran into some problems and saw that the disk space was running out.

Running `df -h` should show if any of the meaningful locations, such as `/var` or `/usr/local`, is running out of space. Docker is often the main culprit so cleaning it up tends to resolve the issue.

MalinAhlberg (approved):
:star: 

All cases in my previous comment are working fine now!

frontend/src/components/shared/SearchHighlighter.vue
```diff
@@ -0,0 +1,25 @@
+<template>
+  <span>
+    <span v-for="(char, j) in matchTerm.split('')" :key="j">
+      <mark v-if="isMatch(char)" class="has-background-warning">{{ char }}</mark>
```
mihai-sysbio, line 4: The `warning` is a bit too strong for me. I'm suggesting to use `bold` as in the Quick Search, and maybe also a pale grey. What do you think?

Come to think of it, it also looks like there is an extra `<span>` holding stuff together, perhaps that can be removed.

```suggestion
      <mark v-if="isMatch(char)" class="has-text-weight-bold has-background-grey-lighter">{{ char }}</mark>
```

mihai-sysbio (approved):
🦄 awesome work!
thanks for the fixes to the previous comments from Malin

frontend/src/components/shared/SearchHighlighter.vue
e0: Thanks for the suggestion! I like the bold effect but don't like the grey background as much. How about having no background instead? Please see attached images for comparison. Note that we would probably switch from `mark` to `span` if we go with no background since `mark` is supposed to be for highlighted text.

We need the extra `span` as well since the inner `span` is with a `v-for` and Vue 2 does not allow multiple root elements.

image

image

nanjiangshu (commented):
Great work @e0! this quick search is responsive and and intuitive! 
I've found only two small issues
1. When type in searchTerm with special characters, such as "#swf", an error message pops up. as shown in the screenshot below. Maybe the searchTerm can be sanitized by `sanitizeSearchString`
Screenshot 2022-06-29 at 13 56 55

2. When tying in the searchTerm quickly, it seems the message `No matches found` was shown first before showing the result as shown in the recording below. Maybe it's better just show spinning circle indicating searching in progress. 

https://user-images.githubusercontent.com/5833722/176433810-38f515a9-cba5-4e91-b4c9-da98e88b0e67.mov

e0:
Thank you for the feedback and suggestions @nanjiangshu! Both cases have been handled now in this commit: https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1051/commits/1ea3711bf634bf2143ceb376b7e87cb82ea7d567


> 1. When type in searchTerm with special characters, such as "#swf", an error message pops up. as shown in the screenshot below. Maybe the searchTerm can be sanitized by `sanitizeSearchString`

I went with a modified version of this and left the original function untouched so it doesn't disrupt other usages.

___

> 2. When tying in the searchTerm quickly, it seems the message `No matches found` was shown first before showing the result as shown in the recording below. Maybe it's better just show spinning circle indicating searching in progress.

This was actually something that we had fixed but reintroduced in commit: https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1051/commits/2e2ba9e2774b1f2d7d58229854bbd0af6cae0bb0

frontend/src/components/shared/SearchHighlighter.vue
e0: Nesting `template` doesn't work either since it gets removed after processing so it still ends up with multiple root elements.

e0: Style change is handled here: https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1051/commits/067c4b2f227c322ab8b7054da0666a7f59fec08e

nanjiangshu:
Cool with `encodeURIComponent`!

PR #1062, opened 2022-07-27
feat: expand the supplementary info in GotEnzymes

### LINKED ISSUE #1061: update GotEnzymes

As a stakeholder, I want the latest generated data provided to be available on the website, so that the end users find the website more useful.

**Acceptance Criteria**

- [x] Integrate the latest data from the researchers by adding new columns in the database where applicable
- [x] The web page documentation is updated accordingly
- [x] The changes are in alignment with the privacy policy
- [x] The changes are tested on multiple models and browsers
- [x] The API response is the same as before - the API response has been expanded but keeping the same format
- [x] The changes do not affect the deployment

**Additional context**

The supplementary data has been enhanced, so the table definition and indexing used in the quick search needs to be updated as well.

mihai-sysbio:
This PR closes #1061.

**Type of change**  

**List of changes made**  

The details pages now contain more information, according to the input files. The indexing has also been updated, going for the basic table indexing instead of the views.
The landing page has also been updated to indicate that text searches based on names is also possible.
The script to seed mock data has been removed, since it is no longer planned to be used.

**Testing**  

- To test, fetch the latest GotEnzymes from OneDrive and unzip the `input_data`. It contains the latest files which are expected by the schema changes in this PR.

**Further comments**  

~In my local testing, the database load times have been improved from over 10 minutes before to about half that. The search feels about the same as before.~ Please do test these aspects on your local setup to verify that the changes don't affect performance. So far the Enzymes tables seems slower, e.g. when sorting and maybe even filtering, but it's not clear (at least to me) where this slowness would come from.

_edit:_ It seems the slow speed is due to the delayed indexing - I think the import finishes but the indexing still continues in the background. This seems to have gone away by simply waiting a bit after the local database re-import and deploying on the dev server. 
The second aspect seems to be related to the RAM and CPU limitations: RAM seems to sit at 75-85%, but the CPU seems to ramp quite easily to the limit of 0.5. This should be addressed when switching to the new production VM.


api/src/gotEnzymes/crossReferencesMapping.js
```diff
@@ -40,7 +40,7 @@ export default {
     dbPrefix: 'sabiork',
     compoundSuffix: '.compound',
     reactionSuffix: '.reaction',
-  },
+  },
```
e0, line 43:
```suggestion
  },
```

api/src/gotEnzymes/crossReferencesMapping.js
```diff
@@ -51,4 +51,12 @@ export default {
...
     compoundSuffix: '',
   },
+  ncbigene: {
+    db: 'NCBI Gene',
+    dbPrefix: 'ncbigene',
+  },
+  uniprot: {
+    db: 'UniProtKB',
+    dbPrefix: 'uniprot',
+  },
```
e0, line 61: It seems like the `/identifier/...` pages are not working correctly for these new reference types. For example:
- `/identifier/NCBI%20Gene/BAA98448?referenceType=gene`
- `/identifier/UniProtKB/Q9Z8U6?referenceType=gene`

Edit: Actually this seems to be for all the new reference types, for example `/identifier/SABIO-RK/321?referenceType=domain` does not work either. This is because we have the following lines in the `idInModels.js` file.

```javascript
    const VALID_REFERENCE_TYPES = ['compound', 'reaction'];
    if (!VALID_REFERENCE_TYPES.includes(referenceType)) {
      throw e;
    }
```

pg/scripts/init.sql
```diff
@@ -52,61 +49,62 @@ create table reactions (
...
-create index on enzymes("domain");
-create index on enzymes("reaction_id");
-create index on enzymes("compound");
+create index on reactions using gist (name, kegg);
+create index on compounds using gist (kegg);
+create index on compounds using gin (string_to_array(name, ';'));
+create index on ec using gist (name, ec);
+create index on organisms (kegg);
+create index on organisms using gist (name);
+create index on domains (abbreviation);
```
e0, line 87: Is this needed since it's a primary key? I noticed that the other similar indexes have been removed, which is correct I think since these are indexed by default.

frontend/src/store/modules/gotEnzymes.js
```diff
@@ -9,33 +9,51 @@ const data = {
...
     commit('setInfo', info);
     commit('setCrossReferences', crossReferences);
   },
-  async getCompoundData({ commit }, id) {
-    const { info, crossReferences } = await gotEnzymesApi.fetchCompound(id);
 
+  async getDomainData({ commit }, id) {
+    const { info } = await gotEnzymesApi.fetchDomain(id);
     commit('setInfo', info);
-    commit('setCrossReferences', crossReferences);
```
e0, line 22: This is needed or else the the domain page can incorrectly display cross references from another component. For example: `/gotenzymes` -> Click the link for `R01677` -> Click back -> Click the link for `eukaryotes`. Similarly this needs to be added for `getOrganismData`. This is not the most DRY solution so happy for other suggestions too.

```suggestion
    commit('setCrossReferences', []);
```

pg/scripts/init.sql
```diff
@@ -52,61 +49,62 @@ create table reactions (
...
+    taxa text
+);
+
+create table domains (
+    abbreviation text primary key not null,
+    name text
+);
 
+create table genes (
+    kegg text,
```
e0, line 65: How come this one is not a `primary key`? It seems like there is no indexing at all now for the `genes` table. I feel like that would negatively impact the performance.

pg/scripts/init.sql
```diff
@@ -52,61 +49,62 @@ create table reactions (
...
 
 -- create lookup indexes
-create index on enzymes("gene");
-create index on enzymes("organism");
-create index on enzymes("domain");
-create index on enzymes("reaction_id");
-create index on enzymes("compound");
+create index on reactions using gist (name, kegg);
+create index on compounds using gist (kegg);
+create index on compounds using gin (string_to_array(name, ';'));
```
e0, line 83: Is this needed? I didn't find any instances where the `name` contains `;`. The query I used: `select count(*) from compounds where name like '%' || ';' || '%';`.

frontend/src/components/gotEnzymes/DetailsPage.vue
```diff
@@ -74,10 +72,6 @@ export default {
   },
   methods: {
     async setup() {
-      if (!SUPPLEMENTARY_INFO_TYPES.includes(this.componentType)) {
-        this.$store.dispatch(`gotEnzymes/resetInfoAndCrossReferences`);
```
MalinAhlberg, line 78: I suspect that this line is still needed, but maybe without the`return`?

If you click gene `zwf` from `gotenzymes/`, go back andt then visit domain `eukaryotes`, the cross references from the gene will still be shown. The provided links give blank pages / error messages.

pg/scripts/init.sql
```diff
@@ -52,61 +49,62 @@ create table reactions (
...
+select text 'ec' as type, ec as id, name as match from ec
 union all
-select text 'reaction' as type, reaction_id as id, reaction_id as match from enzymes_reaction_ids
+select text 'ec' as type, ec as id, ec as match from ec
 union all
-select text 'ec' as type, ec_number as id, ec_number as match from enzymes_ec_numbers
+select text 'organism' as type, kegg as id, name as match from organisms
 union all
-select text 'compound' as type, compound as id, compound as match from enzymes_compounds;
+select text 'domain' as type, abbreviation as id, name as match from domains
```
e0, line 110: We didn't include the `gene` here before creating the gist index for it was taking too long. I wonder if it would make sense to include it as a part of `multi_search` now since it has its own table.

e0 (changes requested):
Good job @mihai-sysbio! I think there is some issues that we should address and I have added inline comments.

MalinAhlberg (changes requested):
Good job! It works fast and well on my machine.

I noted that the cross references on some detail pages does not work as expected, however, see my other comment.

frontend/src/components/gotEnzymes/DetailsPage.vue
MalinAhlberg: (this problem might already be covered by the comments from @e0 above)

e0:
Regarding the search speed, I only tested this briefly but it seems to be slower, especially on the dev server. For example here are the response times for two searches I did on the dev server:

image

I think we can do some more thorough testing after the other indexing related issues are addressed.

api/src/gotEnzymes/crossReferencesMapping.js
mihai-sysbio: Nice catch. I'm not sure why these are needed - is it perhaps because of the Neo4j query that follows it?

pg/scripts/init.sql
mihai-sysbio: Ah this is a tricky one, I forgot to document it.

Making `genes` a primary key would also make them unique. Currently this is a problem because of the `supplementary/gene/` files, since they end with something that is _like_ an empty line (say in VSCcode) but _not_ an empty line that can be replaced with `gsed` or `sed`. Even if those lines are removed, then the `cat` command joins the last line from the first file with the first one from the second. And for some reason `cat -s` didn't work either.

e0: I think there are some other problems with the `genes` data too. It seems that the `kegg` value have duplicates. For example:

image

When searching this looks like:

image

And then the details page of course wouldn't load because we throw an error if the count is not 1.

I think we would want to consolidate these so all of the `ncbigene` and `uniprot` values are semicolon separated. Do you think that's reasonable or is there something else wrong or we should consider here?

pg/scripts/init.sql
mihai-sysbio: Wouldn't that impact the performance negatively?

pg/scripts/init.sql
mihai-sysbio: I haven't checked if it's _needed_, just intuitively thought it would faster if indexed.

api/src/gotEnzymes/crossReferencesMapping.js
e0: We are using the `referenceType` to construct the outgoing URL to the external sites. These need to be manually defined in the `crossReferencesMapping.js` file. Currently we have `compoundSuffix` and `reactionSuffix` so I think we need to add all of the other types too.

```javascript
    const dbSuffix = dbMapping[`${referenceType}Suffix`];
...
        url: `https://identifiers.org/${dbPrefix}${dbSuffix}:${externalId}`,
```

pg/scripts/init.sql
e0: I think that's a valid assumption, but then I don't understand why the other ones are removed.

api/src/gotEnzymes/crossReferencesMapping.js
mihai-sysbio: But that would lead to a bad URL to `identifiers.org`, which I don't think is the problem here.

pg/scripts/init.sql
e0: I just tried this now and it was still taking too long. To answer your question, it would impact the performance of the indexing process negatively but the hope is that it would improve the search performance for the users. If we can get this down to a reasonable time I think it is worth trying. If we can fix the duplicates problems for `genes` maybe we can try this again.

api/src/gotEnzymes/crossReferencesMapping.js
e0: Can you please specify what the problem you are referring to is? The problem I want to solve is that these pages are not loading because we throw an error as mentioned in the initial comment because we have provided an invalid reference type by our definition of `VALID_REFERENCE_TYPES`.

pg/scripts/init.sql
mihai-sysbio: Totally my bad here - there was some merging being done before, for example in the first screenshot above - but that was done _per organism_, leading to the noticed duplicates. Consolidating would probably mean that ta new column for organisms would be added.

Since this is not straightforward to solve, I'm up for creating a different issue, to also handle the UI aspects of the refactoring (such as the links to the organism pages).

pg/scripts/init.sql
mihai-sysbio: Sorry for not being able to follow, I'm not sure what specifically you're thinking of as removed?

For more context, the main change here is indexing on the separate tables rather than the materializes views of the `enzymes` table.

pg/scripts/init.sql
mihai-sysbio, line 83:
```suggestion
create index on compounds using gist (name, kegg);
```

mihai-sysbio: Indeed, that's what the input data has now. I'm sure I saw this somewhere though.. anyway, this can be switched back easily enough, see the suggestion below.

pg/scripts/init.sql
e0: Ah sorry! I was looking at the `enzymes` lines that were removed and thought they were for manually creating the primary index.

e0: Sorry again. Unresolving this because I forgot to ask about why line 87 `create index on domains (abbreviation);` does not use the gist index.

pg/scripts/init.sql
mihai-sysbio: Sounds good!

frontend/src/store/modules/gotEnzymes.js
mihai-sysbio: Instead of doing this in the store, I'm more in favour of returning `[]` from the API, making the API response more consistent.

pg/scripts/init.sql
mihai-sysbio: Th `domain abbreviation` is a single capital letter, so I thought a gist index would not be necessary.

api/src/gotEnzymes/crossReferencesMapping.js
MalinAhlberg: @mihai-sysbio: Me and @inghylt playde around with this. We made these changes https://github.com/MetabolicAtlas/MetabolicAtlas/commit/263550d2ad6567f16e095ce074f2ced90b7ae8b3
which fixes the problem mentioned above and in my comment below.
For ncbi genes, we noted that the identifiers often seem to refer to protein structures. For example, zwf links to BAA98582 which can be found at  https://identifiers.org/ncbiprotein:BAA98582 but not at  https://identifiers.org/ncbigene:BAA98582 . 

Is this expected? As far as we have seen, this change does not break anything.

Which databases are expected to have genes? We only find examples of NCBI, Uniprot and  MetaNetX

frontend/src/store/modules/gotEnzymes.js
e0: Addressed in: https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1062/commits/42ce1b74bae43c2253e4497d9c7fea20ea136089

pg/scripts/init.sql
mihai-sysbio: I thought the index of the `abbreviation` might be relevant when navigating to routes such as `/gotenzymes/domain/E`, but I have nothing against removing it.

api/src/gotEnzymes/crossReferencesMapping.js
e0: @MalinAhlberg and @inghylt Those gene related changes look good I think! For the other new types, I think we also need to take care of `compound`. `domain` and `organism` don't have any cross references data so they don't need to be addressed now.

The other changes in the linked PR, removing the white space and resetting cross references, have been addressed in other commits in this current PR.

mihai-sysbio: The first time around I had added these as `ncbigene` but then I was corrected that the mapping provided by the researchers is from GotEnzymes Gene to NCBI Protein. Even in the links provided above, only the link to the protein works. I have already asked the researchers to improve the Gene cross-references, and they did. So there's little to improve in this regard.

Thanks for fixing the code checking for the `VALID_REFERENCE_TYPES`. I'm not sure why the `if` is needed there though, since I don't think this is user-reachable, so I would remove it.
```
    const VALID_REFERENCE_TYPES = ['compound', 'reaction'];
    if (!VALID_REFERENCE_TYPES.includes(referenceType)) {
      throw e;
    }
 ```

File: pg/scripts/init.sql (hunk 6 above)
Comment (e0) [reply pg/scripts/init.sql]: Addressed in: https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1062/commits/8547e6ebf003bd817bf407c1696299f931ade869

File: pg/scripts/init.sql (hunk 3 above)
Comment (e0) [reply pg/scripts/init.sql]: A primary key automatically gets a `btree` index so this would just be duplicating it. That was a main source of confusion for me before as well. It has been removed in https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1062/commits/8547e6ebf003bd817bf407c1696299f931ade869

File: frontend/src/components/gotEnzymes/DetailsPage.vue (hunk 7 above)
Comment (mihai-sysbio) [reply frontend/src/components/gotEnzymes/DetailsPage.vue:78]: I'm quite sure @e0's commit [42ce1b7](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1062/commits/42ce1b74bae43c2253e4497d9c7fea20ea136089) resolves this.

File: api/src/gotEnzymes/crossReferencesMapping.js (hunk 2 above)
Comment (MalinAhlberg) [reply api/src/gotEnzymes/crossReferencesMapping.js:63]: Commit c6d86b6 fixes the cross references for genes. Compounds already work.

Comment (e0) [reply api/src/gotEnzymes/crossReferencesMapping.js:63]: Nice! I was confused about `compound` before as I thought it was a new type we added.

About the `if` block, if we remove it then it would cause `dbSuffix` to be `undefined`, which would result in a success (`200`) response with an `externalDb` object that has an invalid `url`. Technically this should not be a problem for users of our website but imo this is good practice for API design to return an error response when appropriate instead of a success response with erroneous data. Considering that we would like to expose API endpoints for `GotEnzymes`, this is especially important to get right.

File: api/src/gotEnzymes/crossReferencesMapping.js [hunk 9]
```diff
@@ -51,4 +51,14 @@ export default {
     dbPrefix: 'reactome',
     compoundSuffix: '',
   },
+  ncbigene: {
+    db: 'NCBI Gene',
```
mihai-sysbio, line 55: This has to be:

```suggestion
    db: 'NCBI Protein',
```

pg/scripts/init.sql
e0: It seems that there is not a significant difference between the current number of rows and unique genes `40 016 606` vs `40 002 023` (`14 583` less) so I doubt it would have any significant impact on the indexing. As of result we probably won't be pursuing this in the near future.

api/src/gotEnzymes/crossReferencesMapping.js
mihai-sysbio: Maybe there's something I'm missing.

Given that the `dbSuffix` is simply appended (`${dbPrefix}${dbSuffix}`), the `url` to Identifiers.org could be fixed by changing the `dbPrefix` to include the suffix. Also, in [`genes.js`](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1062/files#diff-097c60247e7d6d4d84a07c1519e1a14279ef626455f51adadbc30607e976a5c8) the suffix is doesn't seem to be used.

pg/scripts/init.sql
e0: Created a [new issue (bug)](https://github.com/MetabolicAtlas/MetabolicAtlas/issues/1067) where we can look more into this.

api/src/gotEnzymes/crossReferencesMapping.js
inghylt: image

inghylt: This is how that would look, legit?

e0: Sorry @inghylt! committed just before I saw your message. I think we are on the same page though.

e0 (approved):

MalinAhlberg (approved):

nanjiangshu (commented):
The new feature on the landing page works great when I tested. However, I did experience a significant slow down when searching as compared to the `main branch`. I have waited for a few hours to avoid the potential problem indexing in progress in the background but it's still slow. As you can see from the video recording

**video recording searching in the new landing page**

https://user-images.githubusercontent.com/5833722/183517266-36fa33dd-d3c3-4903-9fa0-e4f5c680c3c2.mov



**video recording searching in the old landing page** (main branch)


https://user-images.githubusercontent.com/5833722/183516877-cfeb3c0f-4050-4b56-986a-5631fb1b5f5b.mov

I also noted that for the new landing page, postgres often used almost 100% of all 8 cores on my Linux box while in the old landing page, the stress on the CPU is much lower, probably only 10%. I'm not quite sure which queries are causing such an increase in searching complexity. On the other hand, if such searches are needed, maybe we should consider add more CPUs to the server.

api/src/gotEnzymes/crossReferencesMapping.js
mihai-sysbio: I would fetch the `reaction` or `metabolite` (or `gene`) from the tag of the node that is associated with the external ID in Neo4j. And if there is no component in the Neo4j database linked to that external identifier, then the information that needs to be shown is much shorter, so for these cases I would skip mentioning the `referenceType` entirely.
By the way, this refactoring sounds like a new issue to my ears.

mihai-sysbio:
Great analysis @nanjiangshu! Based on @e0's previous comment on the default indexing of indexes being `btree`, my hypothesis is that now searches are more "flexible" because of indexing with `gist` on the primary keys of the table. I guess a way to test this would be to split the indexing into the default (btree) for the primary keys (generally KEGG ids), and gist for the _name_ columns.

api/src/gotEnzymes/crossReferencesMapping.js
mihai-sysbio: Thanks for bringing this up @inghylt.

It's certainly confusing to have protein and gene next together but "NCBI Protein" is what it should say on the GotEnzymes page for genes. Something will need to be done about this identifier page, but this may be merged as is, and any feedback may be addressed separated.

In [another comment](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1062#discussion_r940920432) I'm proposing to forego the `referenceType` altogether, which would remove the word "gene" from this screenshot, and thus revolve the confusion.

I'll have a chat with @feiranl and @Yu-sysbio about this.

api/src/gotEnzymes/crossReferencesMapping.js
e0: For this conversation, I think we can focus on the components that do no exist in the neo4j database because that's where [the relevant code](https://github.com/MetabolicAtlas/MetabolicAtlas/blob/main/api/src/neo4j/queries/idInModels.js#L50-L76) is.

The question is: Given that a user is looking for component `X` in database `Y`, how do you construct the `identifiers.org` URL? More specifically, how do the following get mapped if we want to skip mentioning of the `referenceType` entirely?

| frontend URL | identifiers.org URL |
| --- | --- |
| `/identifier/KEGG/C00003` | `https://identifiers.org/kegg.compound:C00003` |
| `/identifier/MetaNetX/MNXM22270` | `https://identifiers.org/metanetx.chemical:MNXM22270` |
| `/identifier/ChEBI/18050` | `https://identifiers.org/CHEBI:18050` |

e0:
> I guess a way to test this would be to split the indexing into the default (btree) for the primary keys (generally KEGG ids), and gist for the _name_ columns.

For fuzzy search to work, we need to be using `gist` index. I'm wondering if we might be using the `gist` index incorrectly here. When I run `select count(*) from multi_search` I only get 93669. Will try to look more into this now.

e0:
After discussion with @mihai-sysbio, we decided that it is better to merge this PR as is since the performance research may simplify be due to an increased amount of things being indexed. The attempt to improve the performance may also be unsuccessful so it feels like it makes more sense to merge this now instead of waiting.

nanjiangshu:
> After discussion with @mihai-sysbio, we decided that it is better to merge this PR as is since the performance research may simplify be due to an increased amount of things being indexed. The attempt to improve the performance may also be unsuccessful so it feels like it makes more sense to merge this now instead of waiting.

Make sense!

nanjiangshu (approved):
All look good. I agree that we should leave the performance problem to a new issue.

e0:
Just a follow up on the research. I'm 99% sure the poor search performance is because we do not have an index on `genes`. There is so little data in the other tables that it would not make sense otherwise. I have also verified locally to comment away the `gene` related stuff from the search logic and can confirm that everything is much faster. We already have a [new issue for handling duplicate genes](https://github.com/MetabolicAtlas/MetabolicAtlas/issues/1067) so no new research issue is needed.

api/src/gotEnzymes/crossReferencesMapping.js
mihai-sysbio: Had a chat with @e0 which really helped. The key message is that some databases such as `KEGG` or `MetaNetX` are namespaces which require specifying somehow if it's a `compound`, because there are others also such as `reaction` or `gene`. This is not needed for other databases such as `ChEBI`, and it's not needed for Neo4j queries since there we know what the cross-reference is mapped to in the model.

This is worth clearing up in a separate issue.

pg/scripts/init.sql
```diff
@@ -52,61 +49,65 @@ create table reactions (
...
+create index on organisms using gist (name, kegg);
+create index on domains using gist (name);
+
+-- create indexes for the enzymes table
 create index on enzymes("gene");
 create index on enzymes("organism");
 create index on enzymes("domain");
 create index on enzymes("reaction_id");
 create index on enzymes("compound");
 create index on enzymes using gin (string_to_array(ec_number, ';'));
```
mihai-sysbio, line 93: Upon reading up on [multicolumn indexes in Postgres](https://www.postgresql.org/docs/current/indexes-multicolumn.html) and research into the performance decrease by @e0 documented by @nanjiangshu, I propose to revert to single-column indexes everywhere. I'm also taking the opportunity to remove the `"` around column names.

```suggestion
create index on reactions using gist (name);
create index on reactions using gist (kegg);
create index on compounds using gist (name);
create index on compounds using gist (kegg);
create index on ec using gist (name);
create index on ec using gist (ec);
create index on organisms using gist (name);
create index on organisms using gist (kegg);
create index on domains using gist (name);

-- create indexes for the enzymes table
create index on enzymes(gene);
create index on enzymes(organism);
create index on enzymes(domain);
create index on enzymes(reaction_id);
create index on enzymes(compound);
create index on enzymes using gin (string_to_array(ec_number, ';'));
```

mihai-sysbio: > For the purpose of fuzzy search, I think using a compound column is more appropriate.

I'm not sure I follow. The way I see it, for any of the (reaction, compound, ec, organism), it will rarely be the case  that both _name_ and _kegg_ will be typed in the search. Since people are more likely to search after column decoupled from the other, I expect separate indexes (as they were initially set up I think) to be faster for fuzzy searches.

e0: I tried four variations. The one with your suggestion, comma separated (before suggestion), comma separated but primary id first (based on the page you linked to regarding multicolumn gist indexing), and index on compound column (`create index on reactions using gist ((kegg || ' ' || name));`). The performance seems to be about the same for all four actually, they all take around `500ms`. I don't think we have enough data to test this properly. I did some more reading into different implementations and here is a case that's similar to what we had before (comma separated): https://stackoverflow.com/a/29414489

Given the inconclusive results, I can't say any of these are better or worse for our case. I'm fine with merging the suggestion and we can revisit this in the future if needed. Does that sounds good @mihai-sysbio?

mihai-sysbio: Many thanks for the super thorough testing @e0. Not sure if you also looked at the CPU usage, if there was any difference.

I'm going to follow the above and commit the suggestion and merge the PR.

PR #1070, opened 2022-08-09
fix: improve ec search

### LINKED ISSUE #1060: cannot search for some EC codes in GotEnzymes
**Describe the bug**

When searching for some valid EC codes, they don't show up as suggestion so it appears they are missing, even if they are not.

**Steps to reproduce**

Navigate to the landing page of GotEnzymes and search for a valid EC code such as `7.1.1.1`.

**Expected behavior**

- [ ] when searching for a valid EC code, the first search suggestion should be that EC code

**Screenshots**

**Additional context**

It might have to do with the `.` being removed by the API or Postgres itself, since it is not removed by the frontend.

MalinAhlberg:
This PR closes #1060. Joint work with @e0.

The `.` in the EC codes are replaced by a standard character (`a`) in the psql search, so that the standard similarity score works as expected.

**Type of change**  

**List of changes made**  

- Use replacement of special character to improve search results for ECs.

**Screenshot of the fix**  

2022-08-10-174151_973x509_scrot

**Testing**  

- Try searching for example `7.1.1.1` and make sure the first hit is the exact match
- Search for `1.1.3` and see that the results are relevant
- Also try other types of searches to make sure the work as expected and that performance is not affected.

### DIFF
--- a/api/src/gotEnzymes/search.js
+++ b/api/src/gotEnzymes/search.js
@@ -16,13 +16,20 @@ const search = async searchTerm => {
     where kegg = ${searchTerm}
   `;
 
+  const ecMatchQuery = sql`
+    select 'ec' as type, ec as id, ec as match, ${sql`similarity(ec, ${searchTerm})`} as score
+    from ec
+    where ec LIKE ${searchTerm} || '%'
+  `;
+
   await sql`set pg_trgm.similarity_threshold = 0.25`; // this allows more matches, default is 0.3
-  const [fuzzyResults, geneResults] = await Promise.all([
+  const [fuzzyResults, geneResults, ecResults] = await Promise.all([
     fuzzyQuery,
     geneMatchQuery,
+    ecMatchQuery,
   ]);
 
-  return [...geneResults, ...fuzzyResults].slice(0, 10);
+  return [...geneResults, ...ecResults, ...fuzzyResults].slice(0, 10);
 };
 
 export default search;
--- a/api/src/gotEnzymes/search.js
+++ b/api/src/gotEnzymes/search.js
@@ -3,7 +3,14 @@ import sql from 'gotEnzymes/db';
 const search = async searchTerm => {
   const term = searchTerm.toLowerCase();
   const fuzzyQuery = sql`
-    select type, id, match, ${sql`similarity(match, ${term})`} as score
+    select
+      type,
+      id,
+      match,
+      ${sql`similarity(
+        ${sql`replace(match, '.', 'a')`},
+        ${sql`replace(${term}, '.', 'a')`}
+      )`} as score
     from multi_search
     where match % ${term}
     order by score desc
@@ -16,20 +23,13 @@ const search = async searchTerm => {
     where kegg = ${searchTerm}
   `;
 
-  const ecMatchQuery = sql`
-    select 'ec' as type, ec as id, ec as match, ${sql`similarity(ec, ${searchTerm})`} as score
-    from ec
-    where ec LIKE ${searchTerm} || '%'
-  `;
-
   await sql`set pg_trgm.similarity_threshold = 0.25`; // this allows more matches, default is 0.3
-  const [fuzzyResults, geneResults, ecResults] = await Promise.all([
+  const [fuzzyResults, geneResults] = await Promise.all([
     fuzzyQuery,
     geneMatchQuery,
-    ecMatchQuery,
   ]);
 
-  return [...geneResults, ...ecResults, ...fuzzyResults].slice(0, 10);
+  return [...geneResults, ...fuzzyResults].slice(0, 10);
 };
 
 export default search;
--- a/api/src/gotEnzymes/index.js
+++ b/api/src/gotEnzymes/index.js
@@ -7,4 +7,13 @@ import getOrganism from 'gotEnzymes/organism';
 import getReaction from 'gotEnzymes/reaction';
 import search from 'gotEnzymes/search';
 
-export { getCompound, getDomain, getEc, getEnzymes, getGene, getReaction, getOrganism, search };
+export {
+  getCompound,
+  getDomain,
+  getEc,
+  getEnzymes,
+  getGene,
+  getReaction,
+  getOrganism,
+  search,
+};


e0:
When checking both the dev server and locally, it seems that searching for `7.1.1.1` does show up but it may not be ranked first. This is due to the fact that they all get the `similarity` score of `1`.

image

Following is what I see from the dev server. As you see it is listed as the 4th item.

image

So I think we probably do not need to add an extra query to search for `ec`. We could add some custom logic to move the exact match to first in the list and that can be done in JavaScript after all the calls are made I think.

mihai-sysbio:
> This is due to the fact that they all get the `similarity` score of `1`.

Isn't that a bit frustrating though? `7.1.1.1` is an _exact_ match, whereas `1.7.1.1` is not. 

For some  ec codes, the exact match might not even be listed: 
image

e0:
> Isn't that a bit frustrating though? `7.1.1.1` is an _exact_ match, whereas `1.7.1.1` is not.

> For some ec codes, the exact match might not even be listed:

Frustrating indeed. Especially if there is possibility of the exact match not showing in the top 10, which requires further action.

Ideally we would want to see if there is a solution for configuring the fuzzy search to be order sensitive so that when searching for `123`, `123` gets a better score than `231`.

The fast solution (without further research) is to modify the `ecMatchQuery` in this PR to be more similar to the `geneMatchQuery`. So that we look for a perfect match instead. Then we need to manually remove duplicates (if any) before slicing out the top 10 results.

MalinAhlberg:
Thanks for the input @e0 and @mihai-sysbio!
It's a very valid idea to reorder the results in js! Though personally I do appreciate that the result, when searching for eg `7.1.1.`, contains all ids that starts with that exact string, as in my first picture. If that's something we want, I think doing the extra db search for prefixes makes sense. Doing the search for an exact match would in my opinion be a bit frustrating for the user, since you need to type so much. But I'm not a real user of this, so I'm very open for other ideas.

e0:
> It's a very valid idea to reorder the results in js! Though personally I do appreciate that the result, when searching for eg `7.1.1.`, contains all ids that starts with that exact string, as in my first picture.

That's a really good point @MalinAhlberg! I think the new `ecMatchQuery ` is superior in some cases while worse in others. For example it seems to work well for `7.1.1.`. Results below `ecMatchQuery` first, then `multi_search`.

image
image

However if we search for `7.1.11`, we get the following instead. Results below `ecMatchQuery` first, then `multi_search`.

image
image

I think it's worth testing both a bit more and see what the best way to combine them is.

MalinAhlberg:
Interesting @e0! That's a very good point! Let's continue the investigation :)

mihai-sysbio:
I'm adding here another test case `3.1.3.-` is not listing the match. My suspicion is that somehow non-letters have practically no weight in the fuzzy matching.

MalinAhlberg:
> I'm adding here another test case: `3.1.3.-` is not listing the match. My suspicion is that somehow non-letters have practically no weight in the fuzzy matching.



Is there a match for that exact ec? I can't find it at `gotenzymes/ec/3.1.3.-` either?

mihai-sysbio:
> Is there a match for that exact ec? I can't find it at `gotenzymes/ec/3.1.3.-` either?

That's a (different) bug then:
image

e0:
I think that's a known issue from before. There is no such entry in the `ec` table.

image

e0:
I did some more reading and experimenting on this. It turns out that the reason for this case is quite simple as `pg_trgrm` is meant to work with alphanumerical characters. So I thought in the `similarity` function, maybe we can replace the `.` with an accepted character, for example the following:

```sql
  const fuzzyQuery = sql`
    select
      type,
      id,
      match,
      ${sql`similarity(
        ${sql`replace(match, '.', 'a')`},
        ${sql`replace(${term}, '.', 'a')`}
      )`} as score
    from multi_search
    where match % ${term}
    order by score desc
    limit 10
  `;
```

Here is an example of this when run.

image

This looks pretty promising and maybe we can do some more tests on this and see how it behaves.

e0:
> That's an interesting approach, @e0! Does it also work for `-`, as in `3.1.3.-`?

In theory we could replace that too. At the moment we don't have these entries with `-` in the `ec` table. They are only found in the `enzymes` table, which we are not using in `multi_search`.

MalinAhlberg:
I just tested your latest verison @e0 and I think it behaves really nice, almost always. There are some cases that are a bit weird, like this
2022-08-10-154840_728x497_scrot
where you might expect more hits starting with `1.9` being ranked higher.

All in all, I think it's a really nice approach though, as we don't need extra searching or reordering.

e0:
That sounds good @MalinAhlberg and I have similar sentiment after testing! 

This is an interesting example and looking at the results it kind of makes sense. I guess I can understand how the algorithm thinks `1.1.1.9` is most similar to `1.9`.

image

I think this is still a small improvement over what's currently on the dev server now. For example it includes `1.1.5.9`, which does not have the exact sequence of `1.9`.

image

Would you like to commit the changes and un-draft the PR?

e0 (approved):
LGTM! Already tested this rather extensively so will just approve it now.

api/src/gotEnzymes/search.js
```diff
@@ -3,7 +3,14 @@ import sql from 'gotEnzymes/db';
...
   const term = searchTerm.toLowerCase();
   const fuzzyQuery = sql`
-    select type, id, match, ${sql`similarity(match, ${term})`} as score
+    select
+      type,
+      id,
+      match,
+      ${sql`similarity(
+        ${sql`replace(match, '.', 'a')`},
+        ${sql`replace(${term}, '.', 'a')`}
```
mihai-sysbio, line 12: I'm wondering if where this replacement is best done, i.e. would it be better if done in the `frontend` or the `api`, rather than the `pg` container?

inghylt: I have no problem of keeping it here. It gives a strong connection between the replacement and where it is actually used

inghylt (approved):
Have tested searching for different ec codes and other components and everything seems fine, good job :+1:

api/src/gotEnzymes/search.js
nanjiangshu: I also think it's good to place the replacement here. I was a bit confused by the commit message at the beginning though. Sounds like the search term after replacement is used in sql querying but actually the modified strings are used for similarity calculation.

nanjiangshu (approved):
Great fix and very smart way to solve the scoring problem!

PR #1072, opened 2022-08-10
feat: update doc and api

### LINKED ISSUE #1065: expose API endpoints for GotEnzymes

As a user, I want to be able to use GotEnzymes via the API, so that I can built it into my workflows,

**Acceptance Criteria**

- [ ] expose API for individual components (gene, compound, reaction and organism - not domain)
- [ ] The web page documentation is updated accordingly

**Additional context**
There is a built-in limit to the amount of entries that can be fetched from the database. For now, the purpose is to be able to at least use the API. Ideally, one should be able to fetch all entries for a given organism, but that is optional for the current issue - if it's too much work now then a seperate issue may be opened instead.

inghylt:
This PR closes #1065. It was co-created by @inghylt and @MalinAhlberg 

**Type of change**  

- [x] Other
 
**List of changes made**  

- Add examples on how to use the API for gotEnzymes queries in swagger
- Add information about and link to  gotEnzymes in Documentation 

**Screenshot of the fix**  

**Testing**  

- Instructions on how to test
- Visit `/api/v2/`, make sure everything is working and looks good
- Visit `/documentation#api` and make sure everything looks good

**Further comments**  


e0:
Based on the discussion in #1065, I got the impression that the Swagger doc should be updated to include endpoints for individual components. It seems like only the `/gotenzymes/enzymes` endpoint is covered in this PR. Is there something I'm missing?

inghylt:
@e0 That was unfortunately not possible due to the square brackets as mentioned in Slack yesterday. But if you have a look at the result from the changes made, all different type of searches (reactions, genes, organisms etc.) are covered

image

e0:
@inghylt I mean the endpoints mentioned [here](https://github.com/MetabolicAtlas/MetabolicAtlas/issues/1065#issuecomment-1209073317), not the `enzymes` filter related ones.

inghylt:
@e0 Pushed the addition of the links! Please review and see if everything looks good. Chose to keep domain in /gotenzymes/enzymes as some users might want to filter on it, but not exposing a individual url for it, e.g.  /gotenzymes/domains/{someDomain}

api/src/swagger/config.yaml
```diff
@@ -420,3 +421,164 @@ paths:
           description: 'Successful query'
         '404':
           description: 'Invalid input'
+  /gotenzymes/reactions/{reactionId}:
```
e0, line 424: Do we want to rename this to `keggId`? This also applies for `compound`, `organism` and `gene`.

e0 (approved):
This looks really good! All of the examples you provided seem to work just fine. Especially appreciate the extensive work that was done for the endpoint used for the enzymes table.

api/src/swagger/config.yaml
mihai-sysbio: For clarity, that would be of help, but it would also highlight the bug for the genes where our URL expects only a subpart of the gene id.

inghylt: I really don't have a strong opinion of this, feel free to commit changes as you see fit

e0: I'm fine with how things are and sounds like it's not clear that it's better so I will resolve the conversation for now but feel free to un-resolve if you disagree.

MalinAhlberg (approved):
Really nice!

I realized I was wrong about the `default` value, which should (of course) only be used for real defaults, so added suggestions for fixing that.
What I was thinking of was `example`, which I think is nice to provide as they can be used by other tools, so my suggestions include having these instead.

I'm fine with not having examples as well, so I approve

mihai-sysbio:
Having defaults that auto-fill when pressing `try it out` is a great usability improvement, so let's keep that in.

inghylt:
@MalinAhlberg we discussed the suggestions after standup and the conclusion was to keep things as is

inghylt:
Implemented changes suggested at standup and fixed a typo. Merging this as it is approved

PR #1095, opened 2022-08-15
fix: gotenzymes correct error code

### LINKED ISSUE #1074: 500 returned instead of 404 in gotEnzymes queries
**Describe the bug**

If you provide invalid input to gotEnzymes api queries, e.g. `/api/v2/gotenzymes/reactions/R01677xxxxxx`, the server returnes 500 instead of 404

**Steps to reproduce**

Visit `/api/v2/gotenzymes/reactions/R01677xxxxxx`, note the error in `Inspect/Network`

**Expected behavior**

- [ ] 404 and reasonable error message returned

**Screenshots**

**Additional context**

MalinAhlberg:
This PR closes #1074.

Return `404` instead of `500` when no matches are found. Compare to `api/v2/genes/unknown?model=HumanGem`

**Type of change**  

**List of changes made**  

- Return `404` when no (or multiple!) matches are found.
- The swagger doc is updated accordingly.

**Testing**  

- Try out the different api routes under `GotEnzymes` (see `api/v2/#/GotEnzymes`) with made-up ids (eg `
api/v2/gotenzymes/reactions/ueau`) and make sure that they return `404` and not `500`. This should work for  `gotenzymes/reactions`, `gotenzymes/genes`, `gotenzymes/ecs`, `gotenzymes/compounds` and `gotenzymes/organisms`.
- Also test clicking around the GotEnzymes pages (`/gotenzymes`) to make sure everything works as expected.

**Further comments**  

When multiple matches are found, the return code will still be `404`, as it was `500` before. See for example [here](https://github.com/MetabolicAtlas/MetabolicAtlas/blob/main/api/src/gotEnzymes/compound.js#L10). Maybe we should instead return the first hit and log a warning?

### DIFF
--- a/api/src/gotEnzymes/gene.js
+++ b/api/src/gotEnzymes/gene.js
@@ -7,8 +7,8 @@ const getGene = async value => {
     where kegg = ${value.toString()}
   `;
 
-  if (genes.length !== 1) {
-    throw new Error(`Gene ${value} returned ${genes.length} results.`);
+  if (!genes.length) {
+    throw new Error(`Gene ${value} returned ${gene.length} results.`);
   }
 
   const gene = genes[0];
--- a/api/src/gotEnzymes/gene.js
+++ b/api/src/gotEnzymes/gene.js
@@ -8,7 +8,7 @@ const getGene = async value => {
   `;
 
   if (genes.length !== 1) {
-    throw new Error(`Gene ${value} returned ${gene.length} results.`);
+    throw new Error(`Gene ${value} returned ${genes.length} results.`);
   }
 
   const gene = genes[0];
--- a/api/src/gotEnzymes/gene.js
+++ b/api/src/gotEnzymes/gene.js
@@ -8,7 +8,7 @@ const getGene = async value => {
   `;
 
   if (!genes.length) {
-    throw new Error(`Gene ${value} returned ${gene.length} results.`);
+    throw new Error(`Gene ${value} returned ${genes.length} results.`);
   }
 
   const gene = genes[0];
--- a/api/src/swagger/config.yaml
+++ b/api/src/swagger/config.yaml
@@ -438,6 +438,8 @@ paths:
       responses:
         '200':
           description: 'Successful query'
+        '404':
+          description: 'Not found'
         '500':
           description: 'Invalid input'
   /gotenzymes/genes/{geneId}:
@@ -457,6 +459,8 @@ paths:
       responses:
         '200':
           description: 'Successful query'
+        '404':
+          description: 'Not found'
         '500':
           description: 'Invalid input'
   /gotenzymes/ecs/{ecNumber}:
@@ -476,6 +480,8 @@ paths:
       responses:
         '200':
           description: 'Successful query'
+        '404':
+          description: 'Not found'
         '500':
           description: 'Invalid input'
   /gotenzymes/compounds/{compoundId}:
@@ -495,6 +501,8 @@ paths:
       responses:
         '200':
           description: 'Successful query'
+        '404':
+          description: 'Not found'
         '500':
           description: 'Invalid input'
   /gotenzymes/organisms/{organismId}:
@@ -514,6 +522,8 @@ paths:
       responses:
         '200':
           description: 'Successful query'
+        '404':
+          description: 'Not found'
         '500':
           description: 'Invalid input'
   /gotenzymes/enzymes?:
--- a/api/src/swagger/config.yaml
+++ b/api/src/swagger/config.yaml
@@ -582,3 +582,5 @@ paths:
       responses:
         '200':
           description: 'Successful query'
+        '400':
+          description: 'Invalid input'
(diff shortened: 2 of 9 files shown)


api/src/swagger/config.yaml
```diff
@@ -438,6 +438,8 @@ paths:
       responses:
         '200':
           description: 'Successful query'
+        '404':
+          description: 'Not found'
         '500':
```
inghylt, line 443: Shouldn't we remove all 500 in swagger now? I don't think it's possible to trigger that code any longer?

MalinAhlberg: I guess it could be triggered if there is an unexpected error during the code execution, for example if the db fails?
See https://github.com/MetabolicAtlas/MetabolicAtlas/blob/main/api/src/endpoints/gotEnzymes.js#L41 . Maybe the explicit returns of `500`s should be removed, though? I can't find similar cases for the other endpoints.

inghylt: Yeah I think so, since it's unlikely that 500 will be triggered from the swagger api examples

api/src/gotEnzymes/gene.js
```diff
@@ -8,7 +8,8 @@ const getGene = async value => {
   `;
 
   if (genes.length !== 1) {
-    throw new Error(`Gene ${value} returned ${gene.length} results.`);
+    console.log(`Gene ${value} returned ${genes.length} results.`);
```
nanjiangshu, line 11: Shouldn't `console.log` be `console.error`?

nanjiangshu (commented):
Good fix @MalinAhlberg, the endpoint returned the correct status code after the change. However, the reason why `throw new Error` returned the status `500` for `/api/v2/gotenzymes` but not for other API endpoints is because of the status code settings at https://github.com/MetabolicAtlas/MetabolicAtlas/blob/main/api/src/endpoints/gotEnzymes.js#L41-L43. `gotEnzymes.js`  is the only place where the status is set to 500 with caught exceptions. I'm not sure if we need to change them for the consistence of the code. Any opinions from others?

It is similar for the Swagger configuration, where `Invalid input` is set as 500 for GotEnzymes but 404 for the rest of API endpoints.

api/src/gotEnzymes/gene.js
MalinAhlberg: Yes, you're right. Fixed in 3fc0eaf.

api/src/swagger/config.yaml
MalinAhlberg: Removed in 6a297d9

MalinAhlberg:
Thanks @inghylt and @nanjiangshu ! I have refactored a bit, so it would be great if you could retest. No `500`s are left. I also fixed a bug in the swagger doc (2fc2020) so that the different gotenzymes sections can be folded/unfolded separetely

e0 (approved):
LGTM, nice that you fixed the swagger operation ids as well!

MalinAhlberg:
@e0 sorry for pushing new changes after your approval, I just wanted to fix the issue with matches discussed on slack.

e0:
> @e0 sorry for pushing new changes after your approval, I just wanted to fix the issue with matches discussed on slack.

Np! I think [this line](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1095/commits/85bb7e3daf0f9269ecd360e91b0406b778873373#diff-097c60247e7d6d4d84a07c1519e1a14279ef626455f51adadbc30607e976a5c8R11) might have been reverted to a previous  typo. It should be `genes` right?

MalinAhlberg:
> Np! I think [this line](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1095/commits/85bb7e3daf0f9269ecd360e91b0406b778873373#diff-097c60247e7d6d4d84a07c1519e1a14279ef626455f51adadbc30607e976a5c8R11) might have been reverted to a previous typo. It should be `genes` right?

Thanks! Fixed in 07b585

inghylt:
@MalinAhlberg I discovered that we missed replacing example with default for compound so fixed that in 
4dd01f9cc46a7f3b953cfa0ece8ca79a0f0ba817.

Discovered another thing:

`api/v2/gotenzymes/enzymes?filters%5Breaction_id%5D=R00835&pagination%5BpageSize%5D=50&pagination%5Bpage%5D=1&pagination%5Bcolumn%5D=geneTriggerError&pagination%5BisAscending%5D=true`

If you set column to a weird value you will get 500. Should we do something about it or not?

MalinAhlberg:
Thanks for the fix @inghylt and also well spotted with the error. It's fixed in 9d58040 and f7b94a1 (please test).

nanjiangshu (approved):
Nice refactoring and all look good to me!

inghylt (changes requested):
The latest commits has broken the api, we need to fix that

image

MalinAhlberg:
> The latest commits has broken the api, we need to fix that


Sorry for not testing properly... :( . It should work now, after f243576. And yes, testing would definitely be nice and beneficial.

inghylt (approved):
Nice! I think everything looks good now. I added 5c1589864be634ba90786f13454b0dbbc485b9a9 but it's not major, keep it if you like

PR #1098, opened 2022-08-16
fix: numbers on gotenzymes landing page

mihai-sysbio:
This is a tiny PR fixing the landing page for GotEnzymes, so that it is consistent with the _News_ and release text, thus resolving  #1100.

### DIFF
--- a/frontend/src/components/gotEnzymes/Landing.vue
+++ b/frontend/src/components/gotEnzymes/Landing.vue
@@ -116,9 +116,9 @@
             <TableOfContents :links="tocLinks" />
             <div id="intro" class="column content has-text-justified">
               <p>
-                GotEnzymes provides open access to over 25,794,195 million predicted k<sub>cat</sub>
-                for 5,825,213 unique sequences and 4,147 compounds from 8,099 species with the
-                cutting-edge artificial intelligence tools.
+                GotEnzymes provides open access to over 25.7 million predicted k<sub>cat</sub>
+                for +5.8 million unique sequences and 4147 compounds, across 8099 species, predicted
+                with cutting-edge artificial intelligence tools.
               </p>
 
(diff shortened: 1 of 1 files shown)


MalinAhlberg (approved):
Looks good!

frontend/src/components/gotEnzymes/Landing.vue
```diff
@@ -116,9 +116,9 @@
             <TableOfContents :links="tocLinks" />
             <div id="intro" class="column content has-text-justified">
               <p>
-                GotEnzymes provides open access to over 25,794,195 million predicted k<sub>cat</sub>
-                for 5,825,213 unique sequences and 4,147 compounds from 8,099 species with the
-                cutting-edge artificial intelligence tools.
+                GotEnzymes provides open access to over 25.7 million predicted k<sub>cat</sub>
+                for +5.8 million unique sequences and 4147 compounds, across 8099 species, predicted
```
e0, line 120: Minor thing that bothers me here, can we use `over` instead of the `+` prefix to be more consistent with the previous part of the sentence? Not a show-stopper for me so feel free to resolve this conversation without change if you think `+5.8` is fine.

```suggestion
                for over 5.8 million unique sequences and 4147 compounds, across 8099 species, predicted
```

e0 (approved):

frontend/src/components/gotEnzymes/Landing.vue
mihai-sysbio: A change here would still not resolve all instances where this occurs (_News_ page, GitHub releases, Zenodo), so although it's a valid point I hope it's okay to postpone further improvements until the paper gets accepted - since that will for sure lead to some changes in the text in the landing page.

PR #1101, opened 2022-08-17
feat: gotenzymes loader

### LINKED ISSUE #1075: GotEnzymes: Add loading visualization

As a user, I want some kind of indication of when data is loading, so I know that the table I'm looking at will be filled eventually

**Acceptance Criteria**

- [ ] Add loading visualization to the enzymes prediction table (bottom table)
- [ ] add the loader in a way that is also applicable when filters lead to new queries and new response data
- [ ]  when seeing "no data is available" it should only be if that's the case in the database, not that the data is just loading

**Additional context**

A loading animation is extensively used in the GEM Browser - check for potential to reuse.

MalinAhlberg:
This PR closes #1075.

Adds loader for the enzymes table.

**Type of change**  

**List of changes made**  

The loader is shown when:
- a new table is being loaded
- filtering is done
- sorting is done
- next/previous page is clicked
The download button is disabled when the table is loading

**Screenshot of the fix**  

2022-08-17-153812_781x344_scrot

**Testing**  

- To see better how the loader looks and behaves, you can add these lines
```
  const sleep = ms => new Promise(r => setTimeout(r, ms));
  await sleep(1000);
```
to `api/src/endpoints/gotEnzymes.js`, line 113
- Go to eg `gotenzymes/compound/C00003` and try reloading the page, sorting and filtering the columns, moving to another enzymes page etc.

**Further comments**  

### DIFF
--- a/frontend/src/components/gotEnzymes/EnzymesTable.vue
+++ b/frontend/src/components/gotEnzymes/EnzymesTable.vue
@@ -16,11 +16,17 @@
       :rows="enzymes"
       :sort-options="{ enabled: true }"
       :pagination-options="tablePaginationOptions"
+      :is-loading="showEnzymesLoader"
       :total-rows="totalRows"
       @on-page-change="onPageChange"
       @on-sort-change="onSortChange"
       @on-column-filter="onColumnFilter"
     >
+      <div slot="emptystate">No enzymes found</div>
+      <template slot="loadingContent">
+        <div>
+          <loader /></div
+      ></template>
       <template slot="table-row" slot-scope="props">
         <template v-if="linkableFields.includes(props.column.field)">
           <template v-if="props.column.field === 'ec_number'">
@@ -55,6 +61,7 @@
 
 <script>
 import { mapState } from 'vuex';
+import Loader from '@/components/Loader';
 import { VueGoodTable } from 'vue-good-table';
 import ExportTSV from '@/components/shared/ExportTSV';
 import RangeFilter from '@/components/shared/RangeFilter';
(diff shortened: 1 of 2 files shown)


inghylt:
Nice work!

To see how the loading and resulting table will look when there are no matches, try searching for something silly in the filtering:

image

I'm thinking that maybe we would like to keep the styling, i.e. the text centered, the way we had it before?

image

perjo (approved):
LGTM!

MalinAhlberg:
> I'm thinking that maybe we would like to keep the styling, i.e. the text centered, the way we had it before?

Fixed in b98c595

nanjiangshu (commented):
Great work @MalinAhlberg! I've tested with different conditions and all work well. There is just a minor thing that when clicking the next/prev button, the spinning circle is hidden since it is located a couple of rows below the header of the enzyme table. If there is a quick fix, it would be great to place the circle based on the position of the window (e.g. the middle of the window) instead of the position of the header. If it's tricky to do, we can ignore this problem since I believe the delay caused by next/prev button is hardly noticeable.

MalinAhlberg:
> Great work @MalinAhlberg! I've tested with different conditions and all work well. There is just a minor thing that when clicking the next/prev button, the spinning circle is hidden since it is located a couple of rows below the header of the enzyme table. If there is a quick fix, it would be great to place the circle based on the position of the window (e.g. the middle of the window) instead of the position of the header. If it's tricky to do, we can ignore this problem since I believe the delay caused by next/prev button is hardly noticeable.

I could not find any quick fix for this. If someone else has one, please commit it, otherwise I propose we put that in a new issue.

inghylt (approved):
Nice work!

PR #1102, opened 2022-08-18
feat: debounce kcat filter

### LINKED ISSUE #1055: debounce filters for GotEnzymes
**Describe the bug**

The k<sub>cat</sub> filter is the GotEnzymes details pages is not debounced, thus easily leading to confusing results.
edit: make sure all filters are debounced

**Steps to reproduce**

Typing slow leads to confusing filters.

**Expected behavior**

- [ ] incoming API response matches the applied filter

**Screenshots**

**Additional context**

Debounce is already implemented in the Quick Search (GEMs).

perjo:
This PR closes #1055. 

Only behavior of kcat filter in changed, the other GotEnzymes filters a built in Vue-Good-Table and seems to be debounced by default. 

**Type of change**  

**List of changes made**  

Added debounce to kcat min and max filter inputs.

**Screenshot of the fix**  

**Testing**  

1. Go to http://localhost/gotenzymes/reaction/R01677 (for example) 
2. Open developer tools and go to Network tab. 
3. Click "step up" in min value input a couple of times as fast as you can 
4. Verify that a request is only sent for the last value you entered (should be something like `http://localhost/api/v2/gotenzymes/enzymes?filters[reaction_id]=R01677&filters[kcat_values]={"min":4}&pagination[page]=1&pagination[pageSize]=50&pagination[column]=gene&pagination[isAscending]=true`
5. Decrease the value slowly and make sure there is a request passed for every step (and that the data is updated according to the filter) 
6. Type number in the input field and verify that is working too
7. Repeat 1-6 for max value input. 

**Further comments**  


frontend/src/components/shared/RangeFilter.vue
```diff
@@ -1,15 +1,17 @@
...
+      :debounce-events="['click', 'keydown']"
       :class="{ 'input is-danger': !validMin }"
       :title="validMin ? '' : 'Please enter a valid number, e.g. 1.23'"
       class="vgt-input px-2 mr-1"
       type="number"
       placeholder="min"
     />
     <input
-      v-model.number="max"
+      v-debounce:300ms="maxChange"
```
mihai-sysbio, line 13: Could I ask more details on the rationale for this value for the delay, comparing to https://github.com/MetabolicAtlas/MetabolicAtlas/blob/365b6b159f6f445c619be6642c65c7ca27031bc9/frontend/src/components/explorer/gemBrowser/GemSearch.vue#L25

mihai-sysbio (approved):
looks good to me!

frontend/src/components/shared/RangeFilter.vue
e0: I think 300ms should be fine here. In general I think this number should be as little as possible while achieving the effect we want (eliminate race condition). For the GEM search, I think on average the request takes longer (around 600ms from a couple of tests I did now compared to usually sub 100ms response times for enzymes table filtering) so I think it's ok to have slightly longer debounce time there. Wouldn't mind if we had a new issue to test this more in-depth but I think it's ok to have different values now.

e0:
@perjo I haven't tested this PR yet, just wanted to mention that in the future please make sure to use our [branch naming conventions](https://github.com/MetabolicAtlas/MetabolicAtlas/blob/main/CONTRIBUTING.md#creating-a-branch), `feat/` instead of `feature/`, as well as for the [PR title](https://github.com/MetabolicAtlas/MetabolicAtlas/blob/main/CONTRIBUTING.md#creating-a-pr).

mihai-sysbio:
> Good to know! I was doing it by the [NBIS development guidelines](https://github.com/NBISweden/development-guidelines)

It's step number 8 [here](https://github.com/NBISweden/development-guidelines#helpful-commit-messages), but I only know because I added it in.

perjo:
> > Good to know! I was doing it by the [NBIS development guidelines](https://github.com/NBISweden/development-guidelines)
> 
> It's step number 8 [here](https://github.com/NBISweden/development-guidelines#helpful-commit-messages), but I only know because I added it in.

I think I make commits messages as prescribed by conventional commits? The branch naming suggestion still differs in the NBIS and our guidelines. Which is fine, now that I am aware of it.

mihai-sysbio:
> I think I make commits messages as prescribed by conventional commits? The branch naming suggestion still differs in the NBIS and our guidelines. Which is fine, now that I am aware of it.

Getting side-tracked from the PR: it sunds like this should be written up somewhere, do you feel up for sending a PR to the dev docs? Or should we keep that convention just inside this project?

mihai-sysbio:
> Great work @perjo! It seems the filtering works well now except one unexpected behavior that I encountered. As shown in the video recording, a new request is sent to the server when the kcat field is activated and when the cursor is moving around the digits, without changing the value. Such behavior is not observed for the `main` branch.  
> 
> https://user-images.githubusercontent.com/5833722/185433901-e49a2c34-6c2b-44f8-93b8-18e0d21bc0be.mov


Perhaps the trigger should be `['keyup']` instead of `['click', 'keydown']` ?

perjo:
> Great work @perjo! It seems the filtering works well now except one unexpected behavior that I encountered. As shown in the video recording, a new request is sent to the server when the kcat field is activated and when the cursor is moving around the digits, without changing the value. Such behavior is not observed for the `main` branch.
> 
>  kcat-filtering-branch.mov

Well spotted, will look into this.

perjo:
> Perhaps the trigger should be `['keyup']` instead of `['click', 'keydown']` ?

Debounce need to react to `click` since you can step the numeric fields by clicking the arrows. `keyup` or `keydown` may not matter but since `keyup` is default I think we should use that. But we also have to make sure search is not triggered if you haven't entered anything (like pressing an arrow key, this is the error here). This error is actually also present in the Gemsearch field, is also sends another request if an arrow key is pressed.

perjo:
> Debounce need to react to `click` since you can step the numeric fields by clicking the arrows. `keyup` or `keydown` may not matter but since `keyup` is default I think we should use that. But we also have to make sure search is not triggered if you haven't entered anything (like pressing an arrow key, this is the error here). This error is actually also present in the Gemsearch field, is also sends another request if an arrow key is pressed.

Seemed `input` is the proper DOM event to trigger updates. Fixed in [bfafb1d](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1102/commits/bfafb1d3b27558e33f98dcea5e7d165d2f896560).

perjo:
> Seems to work well, good job!
> 
> This is not introduced in this PR, but is related:
> 
> * visit eg http://localhost/gotenzymes/reaction/R01677
> * fill in a max and a min value for the kcat filter
> * delete the values again
> * the table will now be empty as the request sent will be `filters[kcat_values]: {"min":0,"max":0}`
> 
> Do you think it's something that should be fixed here, or put into a new issue?

Well spotted, fixed in [73b824d](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1102/commits/73b824d098604567db8943d2eaed64833871423f)

nanjiangshu (approved):
Looks great!

PR #1104, opened 2022-08-22
feat: RDKit image for metabolite/compound pages

### LINKED ISSUE #1096: metabolite images in GEM Browser and GotEnzymes

As a user, I want to see metabolite images on metabolite/compound pages, so that I can more quickly recognize the molecule I'm after.

**Acceptance Criteria**

- [ ] our own metabolite images in the GEM Browser
- [ ] our own metabolite images in GotEnzymes

**Additional context**

For the GEM Browser, metabolite images are fetched dynamically based on the ChEBI id. However, there can be issues with the requests, and the annotation itself (not all metabolites have associated ChEBI ids). For GotEnzymes, the compounds don't have the ChEBI cross-reference at all.
As part of this feature, it is envisioned to change the source of metabolites from ChEBI to our own, based on the  KEGG id. While this is an about 10% decrease in coverage based on the current cross-references, relying on our own images is a better option.
Alternatively, if there is a KEGG mapping, our own image is loaded; otherwise if there is a ChEBI cross-reference, the request for that external image can be sent.

edit: An attempt should be made to generate the images dynamically first. While JS is preferred, here is a tool that may be used 
https://www.rdkit.org/docs/GettingStartedInPython.html#drawing-molecules

e0:
This PR closes #1096.

**Type of change**  

**List of changes made**  

- Add RDKit generated image for GotEnzymes compound pages.
- Add RDKit generated image for GEM Browser metabolite pages.

**Screenshot of the fix**  

image
image

**Testing**  

- Instructions on how to test
1. Visit `/gotenzymes/compound/C00092`, verify that an image shows up next to the info and references tables.
2. Visit `/gotenzymes/compound/C15825`, verify that **no** image shows up next to the info and references tables and the rest of the page loads normally.
3. Visit `/explore/Human-GEM/gem-browser/metabolite/MAM01637c`, verify that an image shows up next to the info table.
4. Visit `/explore/Human-GEM/gem-browser/metabolite/MAM00420c`, verify that **no** image shows up next to the info table and references tables and the rest of the page loads normally.
5. Please feel free to test some other compound/metabolite pages. I noticed that some images are not as nicely folded as the ChEBI ones (for example `/explore/Human-GEM/gem-browser/metabolite/MAM00053m`). The vast majority look great though and I think this is an ok tradeoff for being able to dynamically generate images. Please let me know if you see any really weird examples though.

**Further comments**  

### DIFF
--- a/frontend/public/index.html
+++ b/frontend/public/index.html
@@ -164,6 +164,7 @@
     <script>
       document.getElementById("init-loading").style.display = "block";
     </script>
+    <script src="https://unpkg.com/@rdkit/rdkit/dist/RDKit_minimal.js"></script>
     <script type="application/ld+json">
       {
         "@context": "https://schema.org",
--- a/api/src/gotEnzymes/compound.js
+++ b/api/src/gotEnzymes/compound.js
@@ -59,7 +59,7 @@ const getSmilesForMetabolite = async ({ formula, crossReferences }) => {
   `;
 
   if (!compounds.length) {
-    return null
+    return null;
   }
 
   return compounds[0].smiles;
--- a/api/src/gotEnzymes/compound.js
+++ b/api/src/gotEnzymes/compound.js
@@ -41,4 +41,28 @@ const getCompound = async id => {
   };
 };
 
-export default getCompound;
+const getSmilesForMetabolite = async ({ formula, crossReferences }) => {
+  const MISSING_PLACEHOLDER = 'A MISSING PLACEHOLDER';
+  // This is needed to make sure the sql query does not fail.
+
+  const compounds = await sql`
+    select smiles from compounds
+    where smiles is not null
+    and (
+      kegg = ${crossReferences['KEGG'] || MISSING_PLACEHOLDER}
+      or meta_net_x = ${crossReferences['MetaNetX'] || MISSING_PLACEHOLDER}
+      or chebi = ${crossReferences['ChEBI'] || MISSING_PLACEHOLDER}
+      or bigg = ${crossReferences['BiGG'] || MISSING_PLACEHOLDER}
+      or formula = ${formula || MISSING_PLACEHOLDER}
+    )
+    limit 1
+  `;
+
+  if (!compounds.length) {
+    return null
+  }
+
+  return compounds[0].smiles;
+};
+
+export { getCompound, getSmilesForMetabolite };
--- a/api/src/gotEnzymes/index.js
+++ b/api/src/gotEnzymes/index.js
@@ -1,4 +1,4 @@
-import getCompound from 'gotEnzymes/compound';
+import { getCompound } from 'gotEnzymes/compound';
 import getDomain from 'gotEnzymes/domain';
 import getEc from 'gotEnzymes/ec';
 import getEnzymes from 'gotEnzymes/enzyme';
--- a/api/src/neo4j/queries/metabolite.js
+++ b/api/src/neo4j/queries/metabolite.js
@@ -5,6 +5,7 @@ import {
   reformatSubsystemSVGs,
 } from 'neo4j/shared/formatter';
 import parseParams from 'neo4j/shared/helper';
+import { getSmilesForMetabolite } from 'gotEnzymes/compound';
 
 const getMetabolite = async ({ id, model, version }) => {
   const [m, v] = parseParams(model, version);
@@ -48,6 +49,26 @@ CALL apoc.cypher.run('
 RETURN apoc.map.mergeList(COLLECT(value.data)) as metabolite
 `;
   const metabolite = await querySingleResult(statement);
+
+  try {
+    const smiles = await getSmilesForMetabolite({
+      formula: metabolite.formula,
+      crossReferences: metabolite.externalDbs.reduce(
+        (acc, curr) => ({
+          ...acc,
+          [curr.dbName]: curr.externalId,
+        }),
+        {}
+      ),
+    });
+
+    if (smiles) {
+      metabolite.smiles = smiles;
+    }
+  } catch (e) {
+    console.error(e);
+  }
+
   metabolite.compartments = [metabolite.compartment];
   return {
     ...metabolite,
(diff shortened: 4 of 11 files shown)


frontend/public/index.html
```diff
@@ -164,6 +164,7 @@ <h3>Loading Metabolic Atlas</h3>
     <script>
       document.getElementById("init-loading").style.display = "block";
     </script>
+    <script src="https://unpkg.com/@rdkit/rdkit/dist/RDKit_minimal.js"></script>
```
mihai-sysbio, line 167: How large is the compressed RDKit? And is it really faster for the user to rely on the CDN rather than Cloudfare when the library is not cached?

e0: It is `140KB` minified locally. Not sure how much that is when compressed but I'm happy to try. Would we need to deploy that to `dev` or can you find out on `localhost`?

mihai-sysbio: Should be able to work just fine locally given some changes in the Nginx config https://github.com/MetabolicAtlas/MetabolicAtlas/blob/1d20628a87c76b2d740651a015a53fcdc3e112da/nginx/prod.nginx.conf#L96

e0: It seems quite tedious to even get this to work locally. According to [the instructions](https://github.com/rdkit/rdkit-js#using-rdkitjs-package-assets), both the `js` and the accompanying `wasm` files need to be included. This would need an additional file loader. I found [this one](https://github.com/BrockReece/vue-wasm), which doesn't seem to be actively developed (might still be fine). The other problem is that the `wasm` file itself is `5MB` locally. I could spend some more time looking into this but it would probably take at least a few more hours and I'm skeptical about having to bundle a `5MB` file to our SPA.

The current solution (CDN script tag) seems to work fine. I think for most cases the performance should be great as well. The script starts loading as soon as the website is loaded and by the time the user gets to a metabolite/compound page it should be ready to go. I have also tested loading a metabolite page directly with uncached CDN (new browser private window) and the image shows up after `2-3s`.

So my recommendation is that we go with the CDN for now, but if you want me to pursue the local option a bit more I can do that as well. Please let me know!

e0: I will open this PR up for review now so others can check it out. We can still change how to use the package later if we want to. Should not affect the functionality.

inghylt:
Very nice work! 

I gave some feedback during standup, but a question that I didn't have time to ask which @mihai-sysbio probably has some input on: Would we like to have some kind of informative text under the images that it was generated through rdkit? E.g. "MetaboliteX via rdkit" or similar

e0:
@inghylt Thanks for the points you brought up this morning!

I looked into the one you mentioned `MAM01934c/C03035` that was missing a `CH3` compared to the dev server. There is several config options and following are some example of how they look like. For the reviewers, please let me know which options we should have. If you are interested here is [the full list of options](https://www.rdkitjs.com/#drawing-molecules-all-options).

If we would like to go with one with the indices or stereo-annotation, would we want to add an option to click on the image to show a larger version in a modal? Maybe this is something we would like to have either way? I think it should be trivial to implement this so I will try to add a commit for it. We can always remove it if it's not desired.

Default (same as now):
image

With Explicit Methyl:
image

With Explicit Methyl + Atom Indices:
image

With Explicit Methyl + Atom Indices + Bond Indices:
image

With Explicit Methyl + Atom Indices + Bond Indices + Stereo-annotation:
image

nanjiangshu:
Looks really cool with different configurations. 
I personally would like to have atom indices on, since it will be easier to specify which part of the molecule when talking with others. If we choose a simple and clear version, I would say version 1 will be best, I don't see a need for explicit `-CH3`

nanjiangshu (commented):
Great work! I tested a number of cases and they all look great. The folding of the molecule for /explore/Human-GEM/gem-browser/metabolite/MAM00053m as done in ChEBI does look better but the folding its self is also misleading since the angle of bonds of the folded illustration is unrealistic. In addition, the RDKit image is vectorized and one gets good resolution when zooming in. 

I can approve this PR if we reach an agreement for the choice of RDkit visualization configuration.

e0:
Thanks for the insight @nanjiangshu! I added the commit for the modal feature now and here is how it looks like for a large compound. This is still with all of options included and I'm happy to change that once we have reached an agreement.

image

MalinAhlberg:
Nice job, @e0, I think this looks really good. The modal with the larger picture is very useful! I have no insight on what configurations should be used, but I do think that the picture for H2O looks a bit empty
2022-08-24-081726_311x325_scrot

e0:
@MalinAhlberg Thanks for this example. I have noticed a few other cases for example `Na+` that also just shows the text. I didn't find any configuration options to expand these but if anyone know how to improve this please let me know.

That little `0` is the atom index btw. If we have atom indices off then it would not be displayed and `H2O` would be centered as well.

perjo (approved):
Looks very nice! I get a

> TypeError: Cannot read properties of null (reading 'short_name')

Skärmavbild 2022-08-25 kl  15 26 52

when visiting

`explore/Human-GEM/gem-browser/metabolite/MAM01637c` or 
`explore/Human-GEM/gem-browser/metabolite/MAM00420c` but I get that in production as well so seems not to be introduced here.

e0:
Just added a commit to remove the extra config options for now since it's the one that's most similar to what we had before. I think we can merge this one in tomorrow unless we get some more input in the morning.

MalinAhlberg (approved):

nanjiangshu (approved):

mihai-sysbio:
> Just added a commit to remove the extra config options for now since it's the one that's most similar to what we had before.

Good idea @e0. The feedback from the researchers was to not use the atom indices, which means one of the first two suggestions (default or with methyl). If we realise changes are needed, it should be easy enough to uncomment those lines.

mihai-sysbio:
> a question that I didn't have time to ask which @mihai-sysbio probably has some input on: Would we like to have some kind of informative text under the images that it was generated through rdkit? E.g. "MetaboliteX via rdkit" or similar

It would be cleaner if we leave this to the  [Metabolite page in the documentation](https://metabolicatlas.org/documentation#gem-browser-metabolite): " The top table of the page contains basic information extracted from the GEM. **When it is possible to infer the SMILES for a metabolite, we will use [RDKit](https://github.com/rdkit/rdkit-js) to draw it to the right side of this table.** |

Issue #1048, opened 2022-06-27
add API tests for GotEnzymes


MalinAhlberg:
**Please describe the feature**

As a developer, I want tests so that I can be sure that everything is working as expected.

**Acceptance Criteria**

- [ ] Tests added for the detail pages (one for each type of data)
- [ ] One or more tests for the search functionality
- [ ] Test the enzymes table

**Additional context**

This issue is partly a leftover from #1032.
The performance aspect is not meant to be tested as part of this issue.

inghylt:
Very nice to have tests in place soon so we don't end up in these situations where api breakage might go unnoticed https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1095#pullrequestreview-1078398586

Issue #1005, opened 2022-05-30
Improve testing code so that it's more robust on model updating 


nanjiangshu:
**Please describe the feature**

As a developer, I want the testing code to be more robust so that it's more robust when models are updated.

**Acceptance Criteria**

- [ ] Broken tests due to model updating are fixed
- [ ] Tests are improved so that they are more robust on model updating

**Additional context**

Related PR https://github.com/MetabolicAtlas/data-files/pull/29

MalinAhlberg:
Potentially good to know for this issue: the current versions are available in `api/test/jest.setup.js` (parsed from `api/src/data/integratedModels.json`)

>

e0:
Please keep in mind that while it would be less tedious to have more robust testing code, it's important to do this in a way that it's not automated to the point where it doesn't detect potential problems. For example if we know for a fact that a model has 1000 reactions and we introduce a bug in data generation that creates 2000 reactions, how should we set the truth for what the number of reactions ought to be?

nanjiangshu:
> Please keep in mind that while it would be less tedious to have more robust testing code, it's important to do this in a way that it's not automated to the point where it doesn't detect potential problems. For example if we know for a fact that a model has 1000 reactions and we introduce a bug in data generation that creates 2000 reactions, how should we set the truth for what the number of reactions ought to be?

I agree. We can not relax all kinds of testing conditions for the sake of robustness.  There is a balance to keep.

perjo:
> For example if we know for a fact that a model has 1000 reactions and we introduce a bug in data generation that creates 2000 reactions, how should we set the truth for what the number of reactions ought to be? 

This should be covered by tests in the data-generation project. Issse link to be added @pdmnyberg

pdmnyberg:
Issue in `data-generation` https://github.com/MetabolicAtlas/data-generation/issues/37

PR #1115, opened 2022-08-24
Test: fix api tests

perjo:

Fixed failing API tests.

**Type of change**  

- [x] Other
 

### DIFF
--- a/api/.eslintrc.js
+++ b/api/.eslintrc.js
@@ -51,5 +51,9 @@ module.exports = {
     'global-require': 'off',
     'prefer-arrow-callback': 'off',
     'func-names': 'off',
+    'jest/expect-expect': [
+      'error',
+      { assertFunctionNames: ['expect', 'validateComponent'] },
+    ],
   },
 };
--- a/api/test/dataOverlay.test.js
+++ b/api/test/dataOverlay.test.js
@@ -8,13 +8,4 @@ describe('data overlay', () => {
     expect(Object.values(data).length).toBeGreaterThan(0);
     expect(Object.values(data)[0].length).toBeGreaterThan(0);
   });
-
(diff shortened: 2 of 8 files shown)


e0 (approved):
Looks great! Really nice that you added the `jest/expect-expect` stuff so we don't get the unnecessary warnings any more.

inghylt:
I think it looks good! One thing I would like to get some feedback on before approving:

- #1005 was created to
    1. Fix the broken tests
    2. Make the test more robust. 
    
       The broken tests are fixed in this PR, but this could be an opportunity to make the tests more robust as well. Note that I am fine with keeping things as is since I'd rather update some tests than miss bugs that creates erroneous data, but I suspect that there are team members with different opinions. @e0 @nanjiangshu @MalinAhlberg @perjo what are your thoughts on this?

e0:
>   2. Make the test more robust.
>      The broken tests are fixed in this PR, but this could be an opportunity to make the tests more robust as well. Note that I am fine with keeping things as is since I'd rather update some tests than miss bugs that creates erroneous data, but I suspect that there are team members with different opinions. @e0 @nanjiangshu @MalinAhlberg @perjo what are your thoughts on this?

I think that would be a much more ambitious task. This current PR is immediately useful and is also nice to get out of the way before #1048. I'm in favor of keeping things as is.

inghylt:
Ok! Then I'm updating #1005

inghylt (approved):

nanjiangshu (approved):
All these tests look great! 

I think we can create a new issue to solve the robustness of tests, that the tests will not break that frequently with model updates. However, I do think test condition should not be too general since otherwise it misses the meaning to catch potential mistakes. One idea could to use less frequently updated models instead of Human-GEM for specific test conditions.

PR #1118, opened 2022-08-25
Test: Add api tests for GotEnzymes

### LINKED ISSUE #1048: add API tests for GotEnzymes

perjo:
This PR closes #1048. Based on the #1115 branch. Code to review is in [gotEnzymes.test.js](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1118/files#diff-ef06ffe07ef20b8d59e1dccd7ccc1e35274732f88cd9d10977d32c35b6755b7c) . 

**Type of change**  

- [x] Other
 
**List of changes made**  

Added API tests for GotEnzymes

**Screenshot of the fix**  

**Testing**  

- Test code, just inspection and running the tests. 

**Further comments**  

Are there any test cases that should be added? 


api/src/neo4j/driver.js
```diff
@@ -1,6 +1,9 @@
 import neo4j from 'neo4j-driver';
 
 const uri = 'bolt://neo4j';
+// To access on host computer (i.e. outside Docker container) use below uri
```
nanjiangshu, line 4: Should such information be written here or in the wiki page? This introduced another way of testing compared to that is documented on the README.

nanjiangshu (commented):
These tests look great and cover a wide range of testing conditions, good job @perjo! 
Since we have mentioned performance (especially searching speed) in a some discussions, I'm thinking if it would be possible to add a test to verify that some slow searches would return  the result within reasonable time?

Otherwise, I think this PR looks great and can be merged.

api/src/neo4j/driver.js
perjo: Don't know, maybe the best would be get it into a config file and then you could use then use that config for running API tests but that was a big bigger. If you do not want this info here then I will remove it.

nanjiangshu: Maybe we can create a new issue to add instructions for testing outside of docker container (if there is such requirement). I just feel a bit unclear with the comments here.

MalinAhlberg (approved):
Very nice with so many tests :tada: ! I think they all look good!

I also agree with @nanjiangshu that we should create an issue for  adding instructions on testing outside of the docker containers.

Also, a question: we I run the tests, the output looks like below. The three last test files get their running times marked in red. Is that expected? And related to [@nanjiangshu's first comment](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1118#pullrequestreview-1085573073), should we keep track of the running time (in another issue)? 
2022-08-26-093344_410x257_scrot

inghylt (approved):
Very nice work! Clean and thorough tests. I agree with @nanjiangshu  that the comments regarding `To access on host computer (i.e. outside Docker container)` are a bit random to have where they are right now, so I would suggest removing them. But that's not a hard requirement for me so I'll approve

api/src/neo4j/driver.js
perjo: Will remove the comments then!

perjo:
Regarding performance tests: it was stated in #1118 

> The performance aspect is not meant to be tested as part of this issue 

but maybe it should be subject to another issue.

perjo:
> Also, a question: we I run the tests, the output looks like below. The three last test files get their running times marked in red. Is that expected? And related to [@nanjiangshu's first comment](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1118#pullrequestreview-1085573073), should we keep track of the running time (in another issue)? 2022-08-26-093344_410x257_scrot

Good question, this seems to be warning that the tests take a long time to run. Seems we can change `slowTestThreshold` to get rid of these warnings (and it makes sense to do that since these are integration tests while the default threshold is adapted for unit tests). Will look into it.

e0:
Haven't tested this but wanted to add a few comments.

It's a bit concerning that the `gotEnzymes` test takes that long to run. Maybe it's due to the indices not being able to fit into the RAM. Was it this slow for everyone?

The `integratedModelsTimeline.svg` should not be checked in here imo. It doesn't specifically hurt but we should probably included it in `.gitignore` in the future since it seems to change every time regardless of if the underlying data changed or not.

perjo:
> It's a bit concerning that the `gotEnzymes` test takes that long to run. Maybe it's due to the indices not being able to fit into the RAM. Was it this slow for everyone?

A vast majority of the time is in the testing of sorting for search enzymes. It is done for twice for each column (may be a bit overdone) where each run is between 500 and 2500 ms when I run it. 

> The `integratedModelsTimeline.svg` should not be checked in here imo. It doesn't specifically hurt but we should probably included it in `.gitignore` in the future since it seems to change every time regardless of if the underlying data changed or not.

Was discussed earlier this week and got the advice to check it in. I haven't understood yet the way it is changed.

e0:
> > The `integratedModelsTimeline.svg` should not be checked in here imo. It doesn't specifically hurt but we should probably included it in `.gitignore` in the future since it seems to change every time regardless of if the underlying data changed or not.
> 
> Was discussed earlier this week and got the advice to check it in. I haven't understood yet the way it is changed.

Ah ok, sorry that I missed this conversation. We can discuss it at a later point then.

PR #1121, opened 2022-08-30
feat: multiple overlay

MalinAhlberg:
This PR closes private issue 138.

Joint work with @inghylt and @e0.

**Type of change**  

**List of changes made**  

Adds possibility to show overlays of different components

**Screenshot of the fix**  

2022-08-30-144945_1342x849_scrot

**Testing**  

- visit eg `explore/Human-GEM/map-viewer/acyl_coa_hydrolysis?dim=2d&panel=1&sel=&search=&coords=-3487.78,-9141.27,0.13,315.43,47.67,-69.89&dataTypes=transcriptomics,fluxomics&dataSources=hpaRna.tsv,HPA_single-cell_reactions.tsv&dataSets=rectum,brain_excitatory%20neurons_0` and verify that you can see the colored overlay for both transcriptomics and fluxomics on the map. In the `Data Overlay` panel, on the right, you should see two boxes, one for each data type.
- verify that you can change data set (Levels) for any of the data types
- verify that you can remove any of the overlay by clicking the cross of the `Overlay` box
- verify that you can reload the page
- try uploading custom data. For for fluxomics (map *Alkaloids biosynthesis*), you can try

 ```
id  tissue 1    tissue 2
MAR01591    0.6 0.3
MAR00447    0.55    0.9
```

- verify that the overlay works in both 3D and 2D.

### CODE REVIEW

File: frontend/src/components/explorer/mapViewer/Svgmap.vue [hunk 1]
```diff
@@ -47,7 +47,7 @@ import MapLoader from '@/components/explorer/mapViewer/MapLoader';
 import MapSearch from '@/components/explorer/mapViewer/MapSearch';
 import { default as messages } from '@/content/messages';
 import { reformatChemicalReactionHTML } from '@/helpers/utils';
-import { DATA_TYPES_COMPONENTS } from '@/helpers/dataOverlay';
+// import { DATA_TYPES_COMPONENTS } from '@/helpers/dataOverlay';
```
inghylt, line 50: I guess this should be removed?

frontend/src/components/explorer/mapViewer/Svgmap.vue
```diff
@@ -348,14 +348,27 @@ export default {
...
       }
-      const allComponents = $(`#svg-wrapper .${this.componentClassName}`);
+      // console.log('apply where computedLevels is:', this.computedLevels);
+      // this.$refs.mapsearch.reset();
+      // if (Object.keys(this.computedLevels).length === 0) {
+      //   Object.values(DATA_TYPES_COMPONENTS).forEach(dataType => {
+      //     $(`#svg-wrapper .${dataType.className} .shape`).attr('fill', dataType.defaultColor);
+      //   });
+      //   return;
+      // }
```
inghylt, line 367: Remove comments

frontend/src/components/explorer/mapViewer/Svgmap.vue
```diff
@@ -348,14 +348,27 @@ export default {
...
+      let allComponents = [];
+      this.componentClassName.forEach(x => {
+        allComponents = [...allComponents, ...$(`#svg-wrapper .${x}`)];
+      });
```
inghylt, line 371:
```suggestion
      // console.log('apply where computedLevels is:', this.computedLevels);
      // this.$refs.mapsearch.reset();
      // if (Object.keys(this.computedLevels).length === 0) {
      //   Object.values(DATA_TYPES_COMPONENTS).forEach(dataType => {
      //     $(`#svg-wrapper .${dataType.className} .shape`).attr('fill', dataType.defaultColor);
      //   });
      //   return;
      // }
      let allComponents = [];
      this.componentClassName.forEach(x => {
        allComponents = [...allComponents, ...$(`#svg-wrapper .${x}`)];
      });
```

frontend/src/components/explorer/mapViewer/Svgmap.vue
inghylt, line 367: remove comments

frontend/src/store/modules/dataOverlay.js
```diff
@@ -93,53 +118,74 @@ const actions = {
...
+        index,
       });
     } catch (e) {
       console.error(e); // eslint-disable-line no-console
       commit('setCurrentDataSource', null);
     }
   },
-  async getDataSet({ commit }, { model, type, filename, dataSet }) {
+  async getDataSet({ commit, state }, { model, type, filename, dataSet, index }) {
+    // console.log('getDataSet', type, filename, dataSet);
```
inghylt, line 129: remove comment

frontend/src/store/modules/dataOverlay.js
```diff
@@ -148,19 +194,48 @@ const mutations = {
...
+      state.currentDataSources = state.currentDataSources.filter((dataSource, i) => i !== index);
+    }, 0);
+    setTimeout(() => {
+      state.dataSets = state.dataSets.filter((dataSet, i) => i !== index);
+    }, 0);
   },
   setCurrentDataSource: (state, currentDataSource) => {
-    state.currentDataSource = currentDataSource;
+    // copy and replace the array to trigger reactive array change detection
+    // console.log('setCurrentDataSource', currentDataSource);
```
inghylt, line 213: remove comment

inghylt (changes requested):
Very nice work @e0! Some feedback/thoughts

Regarding the remove card functionality:
- [Request change] In 2D when you remove a card, the related coloring does not disappear. So I would have to re choose that specific data type and choose 'None' to make it go away. I suggest we fix this so it behaves as in 3D
- Would it be possible to have the cross in the top right corner without the padding so it doesn't shuffle around the rest of the card content when it shows up?
- Should we perhaps not have it in red coloring? Red signals error IMO, I think we can have it more discrete

Regarding the "Add overlay"
- Is the text "Add overlay" confusing? I guess users will figure out the functionality if they play around, but you might interpret that button as that they overlay will be applied when you click it. Maybe rephrase to "Add data type" or similar
- Should we add a message when you hoover the disabled "Add overlay" so user's are informed of why it is disabled? Something like "No more data types available for this model"

Regarding file upload:
- I really like the styling of the modal!
- [Request change] If you upload an erroneous file, the previously listed valid files will be colored orange and not green

frontend/src/store/modules/dataOverlay.js
MalinAhlberg: fixed in 7a66dcf

frontend/src/store/modules/dataOverlay.js
MalinAhlberg: fixed in 7a66dcf

MalinAhlberg:
> * [Request change] In 2D when you remove a card, the related coloring does not disappear. So I would have to re choose that specific data type and choose 'None' to make it go away. I suggest we fix this so it behaves as in 3D

This is fixed in 7b2acda

> * Would it be possible to have the cross in the top right corner without the padding so it doesn't shuffle around the rest of the card content when it shows up?

I changed this is 40a78d2 by using `v-show` instead of `v-if`. Is that an acceptable solution?

> * Should we perhaps not have it in red coloring? Red signals error IMO, I think we can have it more discrete

Changed to black in e5ae158. Is that better?


> * Is the text "Add overlay" confusing? I guess users will figure out the functionality if they play around, but you might interpret that button as that they overlay will be applied when you click it. Maybe rephrase to "Add data type" or similar

Since the cards/boxes have the title *Overlay*, the text "Add overlay" makes sense to me, but more opinions are welcome!

> * Should we add a message when you hoover the disabled "Add overlay" so user's are informed of why it is disabled? Something like "No more data types available for this model"
>
Implemented in c91f49b.


> * [Request change] If you upload an erroneous file, the previously listed valid files will be colored orange and not green

I changed this in  52e8815 so that erroneous files do not get listed, since they are not uploaded and can not be removed. Let me know if think this is ok.

nanjiangshu (commented):
This new feature looks really good. Well done!
I tested randomly and did not find any major problems, there are though a few minor issues.
1. After adding custom data and switch to other pages, e.g. Documentation, when change to the Data overlay page again, the custom data section is spinning, as shown in the screenshot below. I think it should be either emptied or recovered. 

Screenshot 2022-08-31 at 16 42 12

2. In data overlay page, I sometimes managed to add three Overlay panels as shown in the screenshot below. This can usually be reproduced by starting from two Overlay panels and then change to other pages and then move back to the Data overlay page again. It seems the behavior is random since I sometimes repeat exactly the same steps but the status of the `Add overlay` button is different. 
Screenshot 2022-08-31 at 16 52 55

3. Should we also update the Documentation for multiple data overlay?

nanjiangshu:
> > * Is the text "Add overlay" confusing? I guess users will figure out the functionality if they play around, but you might interpret that button as that they overlay will be applied when you click it. Maybe rephrase to "Add data type" or similar
> 
> Since the cards/boxes have the title _Overlay_, the text "Add overlay" makes sense to me, but more opinions are welcome!

I agree with Malin that "Add overlay" sounds better.

inghylt:
I've updated the styling of the cross in 5ae80d308018f35cc605920d4496c8df50990874 and hopefully made it possible switching between different pages without custom data behaving weirdly in 127b6f2dbe52f5ca0fcb378223312bc052ad603f. However, I discovered a bug which we have missed before: copying the url and opening it in a new tab. The data components belonging to the second card get the color grey instead of being colored correctly. Haven't found a fix yet, will continue to research tomorrow

nanjiangshu (commented):
Great work @inghylt. I think all the points mentioned are fixed. I just left a very minor comment regarding the changes in the documentation and otherwise I can approve this PR.

nanjiangshu (approved):
LGTM!

MalinAhlberg:
Nice updates @inghylt and good comments from @nanjiangshu ! I tested the latest version and think it looks and behaves well! The same goes for the documentation. I approve!

One thing we discussed at an earlier stand-up was to let the user choose component types from the ones available [here](https://github.com/MetabolicAtlas/MetabolicAtlas/blob/main/frontend/src/helpers/dataOverlay.js#L5), ie also `metabolomics`. Maybe a new issue for this?

inghylt:
@MalinAhlberg Ah nice that you mentioned that, I thought it made more sense to have in a new issue but did not create one. #1125 has been created now

inghylt (changes requested):
Interaction Partner overlay is broken

frontend/src/components/explorer/mapViewer/DataOverlay.vue
```diff
@@ -215,6 +215,14 @@ export default {
       }
       return this.dataSourcesIndex;
     },
+    filteredDataTypes() {
+      if (this.$route.name === 'interaction') {
+        // do not include fluxomics data for the interaction partners page
+        const dataTypes = this.dataTypes.filter(elem => elem.name !== 'fluxomics');
```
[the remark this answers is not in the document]
MalinAhlberg: I think it's a very relevant concern! However, I'm not familiar enough with the code base to know where such info would be stored. Do you (or perhaps @e0) have any ideas? One option would be creating another issue for this.

MalinAhlberg:
Commit a1707c1 will hide custom files from being listed (under the "Choose a file" button) if their data type is not ok for the current page. Eg fluxomics files will not be listed at "Interaction partners".

frontend/src/components/explorer/mapViewer/DataOverlay.vue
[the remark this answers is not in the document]
inghylt: I vote for refactoring this when we have more data types and use cases

MalinAhlberg:
Great fix @inghylt ! I tested it and it seems to work as expected (but I've said that before for this PR...) and I think the latest commit looks nice and clean.

nanjiangshu:
Nice fix with clean code @inghylt and @e0. I tested the points you mentioned as well as some random tests, the web site works mostly as expected. There are problems I've discovered though. 
1. When you opened two overlay cards in the Map viewer, when you switch to IP, it shows two data overlay cards of the same data type (see attached screenshot below). I think this should be avoided.  Note that only one card was shown after switching before this commit
Screenshot 2022-09-08 at 14 50 59

2.  On the 2D map viewer, add data overlay, the level will be shown on mouse hover. Switch to 3D and back to 2D, no level is shown on mouse hover. The level will be shown again after a hard refresh. I think this is an old problem and I'm not sure if this happens only with my Mac M1 Pro. If the problem is confirmed, we'd probably better to address it in a new issue.

inghylt:
@nanjiangshu commit e46a7079eb3d34e1fb4457f74a357b5c28634c7f should fix the bug regarding hovering (which is actually an old bug we have present on production, great that you caught it!), and as we concluded during standup the multiple overlay card bug is not present. Based on the discussions I also implemented 677c9571761fb96bacb6e683d5d79782bfb11e8e since it was an easy fix to show all uploaded custom data, felt like the most UX friendly implementation.

nanjiangshu:
Great @inghylt that you've fixed the `mouse hover` bug! And I think the looks nice to show all custom uploaded data.

nanjiangshu (approved):
👍

frontend/src/components/Documentation.vue
```diff
@@ -319,14 +320,12 @@
...
+                sidebar also allows uploading of custom data in
                 <a target="_blank" href="https://en.wikipedia.org/wiki/Tab-separated_values">
                   TSV format</a
-                >. If the file is parsed correctly, the file name will be highlighted in green; in
-                case errors are detected, it will be highlighted in red. The expected custom data
-                file should contain at least two columns with headers and using tab delimiter. The
-                first column has to contain gene IDs, identical to the ones in the model. The header
-                of the first column should be named as
+                >. The custom data file should contain at least two columns with headers and use tab
+                delimiter. The first column has to contain gene IDs, identical to the ones in the
```
mihai-sysbio, line 327: The first column can be different types of IDs:
```suggestion
                delimiter. The first column has to contain model IDs for genes, reactions or metabolites.
```

frontend/src/components/Documentation.vue
```diff
@@ -319,14 +320,12 @@
...
+                model. The header of the first column should be named as
```
mihai-sysbio, line 328:
```suggestion
                . The header of the first column should be named as
```

mihai-sysbio (approved):
This has been a massive PR, great job! It's very nice to use and I am eager to approve this PR.
There are a few things I noticed, and I hope you can decide if they should be fixed here or as separate issues:
1. When changing a data type, the previous data type is not cleared. See the video below for 2D. This is not a problem in 3D.
https://user-images.githubusercontent.com/23480589/189317246-a9864d24-4ad5-4eab-a07b-41d2bb6812f7.mp4
2. Changing the data overlay in 3D resets the view (coordinates). I think this has been a bug from before (so it is well suited to be resolved through a separate issue).
3. The header for _Map Viewer_ and _Interaction Partners_ is a bit confusing. A minimal fix would be to remove the word `expression` (as in "expression data"). To make the header better I would suggest moving the box `Data` instead of the current title.
image

inghylt:
> 1. When changing a data type, the previous data type is not cleared. See the video below for 2D. This is not a problem in 3D.
    https://user-images.githubusercontent.com/23480589/189317246-a9864d24-4ad5-4eab-a07b-41d2bb6812f7.mp4

Fixed in 6a281c26c9e132b3a5e00c3eacbe1de5bef09bf7. Also created #1129 since there might be more optimal ways to apply default color to inactive data types

> 2. Changing the data overlay in 3D resets the view (coordinates). I think this has been a bug from before (so it is well suited to be resolved through a separate issue).

Yes I recognize this from before as well. Created #1128

> 3.The header for Map Viewer and Interaction Partners is a bit confusing. A minimal fix would be to remove the word expression (as in "expression data"). To make the header better I would suggest moving the box Data instead of the current title.

I'm not 100% sure what you are suggesting here, please create a new issue or a commit containing the changes

frontend/src/components/explorer/mapViewer/Svgmap.vue
```diff
@@ -162,28 +159,33 @@ export default {
       await this.$store.dispatch('maps/getSvgMap', payload);
       this.bindKeyboardShortcuts();
       this.applyLevelsOnMap();
+      this.setupHoverEventHandlers();
     },
     setupHoverEventHandlers() {
-      const self = this;
+      // construct list of classes to be selected with jquery $(".class1,.class2,.class3,...")
```
perjo, line 165: I am bit curious, why we are using jQuery? Shouldn't plain JS suffice? Below in the same file we are using `document.querySelector`

inghylt: Not sure tbh, we modified existing code when we made these changes. Feel free to create an issue for replacing jQuery, I don't think we should make more changes in this PR if possible

mihai-sysbio: There is already [an issue to remove JQuery](https://github.com/MetabolicAtlas/private-issues/issues/28), so if would be great if the use of this library can be kept to a minimum. It's not a must to resolve this in the current PR, just a thing to keep in mind in the future.

perjo (approved):
Great work, looks good as far as I can see

mihai-sysbio:
> I'm not 100% sure what you are suggesting here, please create a new issue or a commit containing the changes

@inghylt for easier review, I've kept my changes separate in a [separate branch](https://github.com/MetabolicAtlas/MetabolicAtlas/compare/feat/multiple-overlay...feat/data-overlay-box-name?expand=1). What do you think, should it be cherry-picked or have a separate issue and PR?

frontend/src/components/explorer/mapViewer/DataOverlay.vue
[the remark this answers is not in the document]
inghylt: I have documented this discussion [here](https://github.com/MetabolicAtlas/private-issues/issues/74#issuecomment-1246586983), resolving this conversation

Issue #848, opened 2022-01-24
Search can return more than 50 metabolites


e0:
**Describe the bug**

The quick search endpoint should return max 50 results per type. Currently it is possible to return more than 50 metabolites. There is a failing test for this.

**Steps to reproduce**

Run the tests for `api`: `ma-exec api yarn test`. Verify that there is a failing test where more than 50 metabolites are returned.

**Expected behavior**

Make sure that the relevant failing test can pass.

**Additional context**

The test is failing now because the results from `metabolite` and `compartmentalizedMetaboltie` are combined. There is a max limit of 50 for each of them so by combining them it could exceed 50.

perjo:
> Run the tests for api: ma-exec api yarn test. Verify that there is a failing test where more than 50 metabolites are returned.

Which test is this? I backed to before #1115 was merged but I could not find a failing test matching this description? @e0

e0:
@perjo I don't remember when this was fixed. I think it's [this test](https://github.com/MetabolicAtlas/MetabolicAtlas/blob/main/api/test/searchGem.test.js#L9).

perjo:
Ok, a bit of confusion here. Seems problem is fixed by [limiting the number of results per type in the frontend](https://github.com/MetabolicAtlas/MetabolicAtlas/commit/e7a3bb575f2a4a0bb6d61756c9135b497cef915f). (this number has changed back and forth since and is now 10). 

As to why test was failing before it seems the test may have been incorrectly written at first 

`expect(metabolite.length).toBeLessThan(50);` (3b3b01e5388e2fce301f841f824e350a2ddc32f9)

and was got to pass by 

`expect(metabolite.length).toBeLessThanOrEqual(50)` (c3594ea23eff2db5e7d0b1ea2f483f5798823e5b)

To sum up, it seems this issue is fixed in the frontend and not in the API as first suggested, but I guess that is fine and we can close it?

e0:
We discussed this a bit more in #968. I think it is still possible for the results to be greater than the limit, but that is probably not something we want to fix since it has a positive side effect for the frontend. I think we could either leave things as is or even remove this test since it's not something we currently plan to fix.

PR #1123, opened 2022-09-02
Fix: search limit inconsistency

### LINKED ISSUE #968: limit in api search query is not working as expected
**Describe the bug**

**Steps to reproduce**

- Use the api to search with `limit` set to 20, eg `/search?searchTerm=Retinol+metabolism&model=HumanGem&version=1_10_0&limit=20`
- Note that there are 19 metabolites returned, 6 genes etc. The total count is > 20, but each component seems to be limited to <20 hits.
- Do the same search with `limit` 2.
- Note that you get 1 metabolite and 1 subsystem, but 0 for all others, including genes . The total count is now 2, and each component has < 2 hits, sometimes less even when there are hits to be shown.

**Expected behavior**

- [ ] The `limit` should be either by component or in total, and work consistently 

nanjiangshu:
This PR closes #968 and #848 

**Type of change**  

**List of changes made**  

1. The results of GemSearch are sorted by score and <= limit number of top ranked results are returned. This makes sure that the parameter `limit` is more meaningful. 
2. The notification message in GemSearch is changed to `Limited to 50 results (max 10 per type)` to reflect that the limit is for all components, but when showing on the frontend, at maximum 10 per type are shown.
3. The test for GemSearch is updated to reflect that the limit is for all components.

**Screenshot of the fix**  

**Testing**  

**Further comments**  

The results of GemSearch are sometimes different compared to before since some of the low scoring hits exceeding limit 50 are excluded. 

### DIFF
--- a/api/src/neo4j/queries/search/modelSearch.js
+++ b/api/src/neo4j/queries/search/modelSearch.js
@@ -45,12 +45,12 @@ WITH apoc.map.mergeList(apoc.coll.flatten(
 )) as metabolites, mid
 RETURN metabolites {mid: mid, .*}
 `;
-
   if (limit) {
     statement += `
 LIMIT ${limit}
 `;
   }
+
   return queryListResult(statement);
 };
 
@@ -242,9 +242,7 @@ LIMIT 50
         metaboliteIds: groupedByComponents['Metabolite'] || [],
         model,
         version: v,
-        limit:
-          (groupedByComponents['CompartmentalizedMetabolite'] || []).length +
-          (groupedByComponents['Metabolite'] || []).length,
+        limit,
       }),
       fetchGenes({
         ids: groupedByComponents['Gene'],
@@ -282,24 +280,42 @@ LIMIT 50
     compartments,
   };
 
-  const resWithScore = {};
+  let resWithScore = [];
   for (const [component, result] of Object.entries(resObj)) {
     if (result) {
-      resWithScore[component] = result.map(obj => ({
-        ...obj,
-        score: getScore(obj, uniqueIds),
-      }));
-    } else {
-      resWithScore[component] = [];
+      resWithScore = resWithScore.concat(
+        result.map(obj => ({
+          ...obj,
+          score: getScore(obj, uniqueIds),
+          component,
+        }))
+      );
+    }
+  }
+  // sort all results by score in descending order
+  resWithScore.sort((a, b) => b.score - a.score);
+
+  // take only <= limit number of results and group them by components
+  const resWithScoreGroupedByComponent = resWithScore
+    .slice(0, limit)
+    .reduce(function (r, a) {
+      r[a.component] = r[a.component] || [];
+      r[a.component].push(a);
+      return r;
+    }, Object.create(null));
+
+  for (const component of Object.keys(resObj)) {
+    if (!(component in resWithScoreGroupedByComponent)) {
+      resWithScoreGroupedByComponent[component] = [];
     }
   }
 
   return {
-    metabolite: resWithScore.metabolites,
-    gene: resWithScore.genes,
-    reaction: resWithScore.reactions,
-    subsystem: resWithScore.subsystems,
-    compartment: resWithScore.compartments,
+    metabolite: resWithScoreGroupedByComponent.metabolites,
+    gene: resWithScoreGroupedByComponent.genes,
+    reaction: resWithScoreGroupedByComponent.reactions,
+    subsystem: resWithScoreGroupedByComponent.subsystems,
+    compartment: resWithScoreGroupedByComponent.compartments,
   };
 };
 
--- a/api/src/neo4j/queries/search/modelSearch.js
+++ b/api/src/neo4j/queries/search/modelSearch.js
@@ -7,6 +7,47 @@ import {
   getScore,
 } from 'neo4j/queries/search/helper';
 
+const fetchIds = async ({ component, term, model, v, limit }) => {
+  // The search term is used twice, once with exact match and once with
+  // fuzzy match. This seems to produce optimal results.
+  let statement = `
+CALL db.index.fulltext.queryNodes("fulltext", "${term} ${term}~")
+YIELD node, score
+WITH node, score, LABELS(node) as labelList
+OPTIONAL MATCH (node:${model}:${component})
+OPTIONAL MATCH (node)-[${v}]-(parentNode:${model}:${component})
+WHERE node:${model} OR parentNode:${model}
+WITH DISTINCT(
+	CASE
+                WHEN EXISTS(node.id) AND NOT apoc.coll.intersection(${JSON.stringify(
+                  CHILD_LABELS
+                )}, LABELS(node))
+                THEN { id: node.id, labels: labelList, score: score }
+		ELSE { id: parentNode.id, labels: LABELS(parentNode), score: score }
+	END
+) as r 
+WHERE any(r IN r.labels WHERE r="${model}")
+RETURN r
+`;
+
+  if (limit) {
+    statement += `
+LIMIT ${limit}
+`;
+  }
+
+  const results = await queryListResult(statement);
+
+  const uniqueIds = {};
+  for (let [pos, node] of Object.entries(results)) {
+    if (!(node['id'] in uniqueIds)) {
+      uniqueIds[node['id']] = node;
+    }
+  }
+
+  return uniqueIds;
+};
+
 const fetchCompartmentalizedMetabolites = async ({
   ids,
   metaboliteIds,
@@ -45,7 +86,9 @@ WITH apoc.map.mergeList(apoc.coll.flatten(
 )) as metabolites, mid
 RETURN metabolites {mid: mid, .*}
 `;
-
+  // limit is applied for CompartmentalizedMetabolite again since the number of
+  // IDs for Metabolite and CompartmentalizedMetabolite together may exceed the
+  // at the first step when fethcing unique IDs
   if (limit) {
     statement += `
 LIMIT ${limit}
@@ -54,12 +97,12 @@ LIMIT ${limit}
   return queryListResult(statement);
 };
 
-const fetchGenes = async ({ ids, model, limit, version }) => {
+const fetchGenes = async ({ ids, model, version }) => {
   if (!ids) {
     return null;
   }
 
-  let statement = `
+  const statement = `
 WITH ${JSON.stringify(ids)} as gids
 UNWIND gids as gid
 CALL apoc.cypher.run("
@@ -71,21 +114,15 @@ RETURN apoc.map.mergeList(apoc.coll.flatten(
 )) as gene
 `;
 
-  if (limit) {
-    statement += `
-LIMIT ${limit}
-`;
-  }
-
   return queryListResult(statement);
 };
 
-const fetchReactions = async ({ ids, model, limit, version }) => {
+const fetchReactions = async ({ ids, model, version }) => {
   if (!ids) {
     return null;
   }
 
-  let statement = `
+  const statement = `
 WITH ${JSON.stringify(ids)} as rids
 UNWIND rids as rid
 CALL apoc.cypher.run("
@@ -107,20 +144,15 @@ RETURN apoc.map.mergeList(apoc.coll.flatten(
 )) as reaction
 `;
 
-  if (limit) {
-    statement += `
-LIMIT ${limit}
-`;
-  }
   return queryListResult(statement);
 };
 
-const fetchSubsystems = async ({ ids, model, version, limit }) => {
+const fetchSubsystems = async ({ ids, model, version }) => {
   if (!ids) {
     return null;
   }
 
-  let statement = `
+  const statement = `
 WITH ${JSON.stringify(ids)} as sids
 UNWIND sids as sid
 CALL apoc.cypher.run("
@@ -132,20 +164,15 @@ RETURN apoc.map.mergeList(apoc.coll.flatten(
 )) as subsystem
 `;
 
-  if (limit) {
-    statement += `
-LIMIT ${limit}
-`;
-  }
   return queryListResult(statement);
 };
 
-const fetchCompartments = async ({ ids, model, version, limit }) => {
+const fetchCompartments = async ({ ids, model, version }) => {
   if (!ids) {
     return null;
   }
 
-  let statement = `
+  const statement = `
 WITH ${JSON.stringify(ids)} as cids
 UNWIND cids as cid
 CALL apoc.cypher.run("
@@ -157,11 +184,6 @@ RETURN apoc.map.mergeList(apoc.coll.flatten(
 )) as compartment
 `;
 
-  if (limit) {
-    statement += `
-LIMIT ${limit}
-`;
-  }
   return queryListResult(statement);
 };
 
@@ -175,7 +197,7 @@ const modelSearch = async ({ searchTerm, model, version, limit }) => {
     searchTerm,
     model,
     version,
-    limit: limit || 10, // limit per component
+    limit: limit || 10, // limit is applied per component type
   });
 
   return {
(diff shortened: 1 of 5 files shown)


frontend/src/components/explorer/gemBrowser/GemSearch.vue
```diff
@@ -49,7 +49,7 @@
         class="notification is-large is-unselectable has-text-centered is-clickable py-1 mb-1"
         @mousedown="globalSearch()"
       >
-        Limited to 10 results per type. Click here to search all integrated GEMs
+        Limited to 50 results (max 10 per type). Click here to search all integrated GEMs
```
inghylt, line 52: To me this sentence is a bit messy. Do users need to know that we have a limit on 50 results in total? To me it makes sense to just keep the sentence the way it was before, but maybe there are other opinions on this

inghylt:
I think these changes make sense and that the web site look good, but I do feel like it would be nice with input from some of the scientific advisors before making a final decision on this. I have a hard time to know if users preferred things the way they were before or the current implementation

frontend/src/components/explorer/gemBrowser/GemSearch.vue
nanjiangshu: I agree it's a bit messy, but only saying `max 10 per type` gives the use a false feeling that the limit is set on individual component but it's not. Maybe we can just show `max 10 per type` but add a sentence in the documentation to explain it.

mihai-sysbio: How about a shorter version ?
```suggestion
        limited to 10 rows per type; click here to search all integrated GEMs
```

nanjiangshu: Good suggestion! Using `rows` to separate from results. The description of Quick Search are also updated in the Documentation to reflect that the limit of 50 is for all components instead of per type. see d5727ecc.

e0:
@nanjiangshu Thanks for creating this PR. I think the changes look good technically and make the system more accurate. However I still think the previous (buggy) implementation has a side effect that is beneficial for people who use the website. For example if you search for `retinal`, it now displays 10 metabolites, 2 reactions, and 2 genes. On the current production site, it shows 10 metabolites, 7 reactions, 10 genes, and 1 subsystem.

My opinion is that if we want to make this change to the backend, then we should get rid of the limit of 10 per type in the frontend. Maybe an accompanying change for this would be to add an interactive result summary at the top (UI example attached below), indicating the type and count.

Overall, I think we are better off not making any changes at this point. The buggy implementation has a good side effect for the users and I don't think this is a worthy tradeoff at the moment to sometimes display less results. I think it would be a good research issue to reimagine how we can improve the quick search.

image

nanjiangshu:
> @nanjiangshu Thanks for creating this PR. I think the changes look good technically and make the system more accurate. However I still think the previous (buggy) implementation has a side effect that is beneficial for people who use the website. For example if you search for `retinal`, it now displays 10 metabolites, 2 reactions, and 2 genes. On the current production site, it shows 10 metabolites, 7 reactions, 10 genes, and 1 subsystem.



> My opinion is that if we want to make this change to the backend, then we should get rid of the limit of 10 per type in the frontend. Maybe an accompanying change for this would be to add an interactive result summary at the top (UI example attached below), indicating the type and count.
> 
> Overall, I think we are better off not making any changes at this point. The buggy implementation has a good side effect for the users and I don't think this is a worthy tradeoff at the moment to sometimes display less results. I think it would be a good research issue to reimagine how we can improve the quick search.

I see your concern @e0 and I think you made a good point here that we should rethink how to improve the quick search considering that this is the most used feature for MetAtlas according to hotjar. Before doing so, we should have a definition of `what is a good quick search` since without that we can not compare which approaches are better. 
When you use the example ``retinal`, it seems the current approach is better since it shows more hits. Then we are assuming that under the condition <= 10 hits to be shown per type and the same search time, the approach returns more hits are better. I think this assumption also makes sense although it is difficult to judge whether a reaction hit ranked at e.g. 51th place by matching score is better to be shown than a metabolite hit ranked at 11th place.

However, if we all accept this definition, that under the condition <= 10 hits to be shown per type and the same search time, the approach returns more hits are better. I think we can make the following changes to the quick search.
1. The limit should be per type, not for all, since we will not show more then the limit (10) per type set on the frontend. 
In this case, the limit 10 should be passed as the parameter to API
2. The limit 50 for the fuzzy search should not be used as a parameter on the API, since this is only a parameter used to adjust the speed of quick search. Moreover, since we want to get as more hits as possible per type, we can reconsider increasing this number as long as the speed of the quick search is still reasonably fast
3. The limit 10 passed from the API should be set for searching of each component, since it is meaningless to return more than 10 hits per type as more of them will not be shown on the frontend
4. The test should be changed to `GEM search should have max 10 results per type`
5. The notification message and docs should also be changed accordingly



> image

e0:
@nanjiangshu Thanks for writing up this new proposal. I think it's an interesting alternative and I think it would be better in certain scenarios.

I think the main question for me is how important the search score should be compared to how important it is to have results from different types. If the top 100 results for a search-term are 90 metabolites and 10 reactions, the question becomes: should we show 10 metabolites + 10 reactions or 50 metabolites? In this case, if the search score is the same for all results, then it would probably be better to show 10 metabolites + 10 reactions. However if the search score for metabolites are much higher than that of reactions, then I think it is probably better to show 50 metabolites. I do think it would be an edge case for all of the results to have the exact same score though. So I think it makes sense to just take the top 50 results sorted by relevance regardless of type.

I still like this suggestion you presented though and I'm happy to discuss more about this. Currently the following is how I would rank the different alternatives.

1. Backend: top 50 regardless of type (as done in this PR). Frontend: show all 50 results with UX improvements (as mentioned [here](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1123#issuecomment-1238222068)).
2. Backend: top 10 for each type. Frontend: top 10 for each type.
3. Backend: top 50 with bug of possibly additional metabolites (what is currently in production). Frontend: top 10 for each type.
4. Backend: top 50 regardless of type (as done in this PR). Frontend: top 10 for each type.

nanjiangshu:
Thanks for the great suggestions @e0. I'm still in favor of the second choice you listed, i.e. the method that I last proposed. The reason why I'm not in favor of showing all top ranked results is because I'm not sure if the matching score fully represents the relevance of the match. Therefore it might be better to show hits from other components that do not have the best matching score. 

What do others think @inghylt  @mihai-sysbio  @MalinAhlberg @perjo ?

MalinAhlberg:
Thanks for the interesting and detailed discussion, @nanjiang and @e0!

As you've already concluded, it's very hard to know what would be the best option given the very different kind of use cases.  My intuition would be that it's best to set the limit for each type, since I personally like when the (frontend result) list is easy to get an overwiew of. And it's nice if the api and frontend does not show different results. If a user would want more results, they could/should use the global search instead. 
 So, using the numbering from [@e0's comment above](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1123#issuecomment-1239110693), I vote for alternative 2.

inghylt:
Well written summaries! I also think it makes sense to do top 10 per type (alternative 2) since we don't fully control the relevance of the scores provided by neo4j.

mihai-sysbio:
In line with [a previous comment](https://github.com/MetabolicAtlas/MetabolicAtlas/issues/968#issuecomment-1233830119) and the voting above, the option to go for seems to be:

> 2\. Backend: top 10 for each type. Frontend: top 10 for each type.

e0:
Thanks for the input everyone! I can definitely see the argument for option 2, especially since it means that the frontend stays visually consistent for the users. We already have logic to sort the most relevant type at the top so it shouldn't be too much of a concern that less relevant results from other types may be included.

Are we good with continuing the work on this current issue/branch/PR?

nanjiangshu:
Great that we have an agreement now. I think it's OK to continue with this PR and we have lots of valuable discussions here.

e0:
@nanjiangshu I forgot to ask the other day about the relevance of the scores. From what I have seen the scores seem to make sense. If you notice cases where the score does not make sense, that's probably worth looking into more and it would be nice if you can create a bug issue for it.

nanjiangshu:
> Thanks for the input everyone! I can definitely see the argument for option 2, especially since it means that the frontend stays visually consistent for the users. We already have logic to sort the most relevant type at the top so it shouldn't be too much of a concern that less relevant results from other types may be included.

The hit with the highest score might not be the one which the users are expecting since we don't know exactly which component type the user is expecting without further knowledge. For example, if we search for `fatty` in `Human-GEM`, there are 56 hits if we limit the fuzzy search to 50, out of these 56 hits, there are two hits for subsystems but they are ranked at the bottom place. 

```
 {
   name: 'Fatty acid activation (cytosolic)',
   id: 'fatty_acid_activation_cytosolic',
   score: 3.4419190883636475,
   component: 'subsystems'
 },
 {
   name: 'Fatty acid biosynthesis (unsaturated)',
   id: 'fatty_acid_biosynthesis_unsaturated',
   score: 3.4419190883636475,
   component: 'subsystems'
 }  
 ```
 if we show top 50, then the two hits of subsystems will be excluded,  how if the users are expecting hits from subsystems? Of course, we can show all 56 hits, but then we will have two more problem 
 1. Too many hits may be shown, sometimes it can be over 100 even the limit for the fuzzy search is 50
 2. The number of returned hits are varying in a wide range and it is difficult to explain the `limit` thing we are using.

**Comment by nanjiangshu** [comment]
I have implemented the strategy(2) as we discussed. This PR is ready for a new review now.

File: api/src/neo4j/queries/search/modelSearch.js [hunk 2]
```diff
@@ -194,12 +215,8 @@ WITH DISTINCT(
 ) as r 
 WHERE any(r IN r.labels WHERE r="${model}")
 RETURN r
+LIMIT 50
```
MalinAhlberg, line 218: ?

perjo (commented):
Great work! I think there is one thing that could be improved but does not need to be done in this PR. Now the limits `10` and `50` are floating around in the code as [magic numbers](https://wiki.c2.com/?MagicNumber). A think the code would become clearer if these are made into constants with names that explain what they are about.

MalinAhlberg (commented):
Good work! I agree with @perjo about the magic numbers, and it would be good with some comment explaining especially the `50`. But apart from that I think it looks all good!

e0:
Thanks for the great example with using `fatty` @nanjiangshu! It illustrates the problem really well.

I tested what we discussed during stand-up a bit more now regarding top 10 out of 50 vs top 10 out of entire database. I think the following change would work to limit the fuzzy search to a specific type. Maybe it's a starting point if you want to explore this option.

image

nanjiangshu:
Thanks @e0 for the great suggestion that the GEM search can be limited for each component when fetching the unique IDs. After implementing, I find that the new approach usually returns hits from more component types, such as search terms, `fatty`, `transporter`, `bile`, `acid`. As to the performance, I haven't done any solid benchmark but feels the new approach is usually faster. The problem with the magic limit number 50, as @MalinAhlberg and @perjo pointed out, is also gone. There is now only one limit number (default to 10) passed as a parameter to the API. I think this approach not only solves the limit inconsistency problem but also improves the GEM search.

inghylt:
Very nice work, everything looks good and makes sense to me. I think [this comment](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1123#pullrequestreview-1108621672) about magic numbers is still valid but since we are only using the number 10 now I'm fine with merging this without creating a constant.

It seems like one test is broken now though? I will approve once that is fixed or if it turns out that it's only failing on my computer

e0:
I took a quick look through the changes and it looks good I think! I would like to do a more thorough testing and review on Monday, but if you get enough approvals today and want to merge it that's fine too.

MalinAhlberg:
> Very nice work, everything looks good and makes sense to me. I think [this comment](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1123#pullrequestreview-1108621672) about magic numbers is still valid but since we are only using the number 10 now I'm fine with merging this without creating a constant.
> 
> It seems like one test is broken now though? I will approve once that is fixed or if it turns out that it's only failing on my computer

One test fails for me as well. Otherwise I think this looks very good!

nanjiangshu:
Hi @inghylt and @MalinAhlberg , thanks for pointing out the problem with failed test. Sorry that I haven't checked more carefully before. It seems the neo4j querying returned hits labeled with `Reaction` when searching for other components, 
e.g. we are supposing to return hits only for the `Compartment` but it returned a result for `Reaction`. I need to dig more into querying syntax.  
```
CALL db.index.fulltext.queryNodes("fulltext", "RCR12105 RCR12105~")
YIELD node, score
WITH node, score, LABELS(node) as labelList
OPTIONAL MATCH (node:HumanGem:Compartment)
OPTIONAL MATCH (node)-[:V1_11_0]-(parentNode:HumanGem:Compartment)
WHERE node:HumanGem OR parentNode:HumanGem
WITH DISTINCT(
	CASE
                WHEN EXISTS(node.id) AND NOT apoc.coll.intersection(["ExternalDb","PubmedReference"], LABELS(node))
                THEN { id: node.id, labels: labelList, score: score }
		ELSE { id: parentNode.id, labels: LABELS(parentNode), score: score }
	END
) as r 
WHERE any(r IN r.labels WHERE r="HumanGem")
RETURN r

LIMIT 1 
```

nanjiangshu:
Hi @inghylt and @MalinAhlberg, I've made the commit d15280a to make sure only hits of the component in search are returned. Now all tests are passed.

nanjiangshu:
> Very nice work, everything looks good and makes sense to me. I think [this comment](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1123#pullrequestreview-1108621672) about magic numbers is still valid but since we are only using the number 10 now I'm fine with merging this without creating a constant.

@inghylt  Regarding the magic number 10, the number 10 is actually a default value in the backend, and this number will be overridden by the limit set at the frontend at [this line](https://github.com/MetabolicAtlas/MetabolicAtlas/blob/fix/search-limit-inconsistency/frontend/src/store/modules/search.js#L102). Unless we want to configure this limit number in the Docker ENV file, I think it is OK to set the value here.

inghylt (approved):
Nice fix and work in general, this has been a drawn-out issue!

api/src/neo4j/queries/search/modelSearch.js
```diff
@@ -7,6 +7,49 @@ import {
...
+  if (limit) {
+    statement += `
+LIMIT ${limit}
+`;
+  }
+
+  const results = await queryListResult(statement);
+
+  const uniqueIds = {};
+  for (let [pos, node] of Object.entries(results)) {
```
e0, line 44: It is idiomatic to not use any placeholder and simply ignore the values of an array you are not interested in. See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment

```suggestion
  for (let [, node] of Object.entries(results)) {
```

api/src/neo4j/queries/search/modelSearch.js
```diff
@@ -7,6 +7,49 @@ import {
   getScore,
 } from 'neo4j/queries/search/helper';
 
+const fetchIds = async ({ component, term, model, version, limit }) => {
```
e0, line 10: Maybe we could rename this function to something like `searchForIds` instead? This way there would be a clearer distinction between the purpose of this function compared to the purposes of other functions like as `fetchGenes`.

api/src/neo4j/queries/search/modelSearch.js
```diff
@@ -167,48 +211,49 @@ const modelSearch = async ({ searchTerm, model, version, limit }) => {
...
+    'Metabolite',
+    'CompartmentalizedMetabolite',
+    'Gene',
+    'Reaction',
+    'Subsystem',
+    'Compartment',
+  ]) {
+    resultFetchIds.push(
+      fetchIds({ component, term, model, version: v, limit })
+    );
```
e0, line 238: A few points here.

1. The name `resultFetchIds` could perhaps be renamed to something like `idSearchQueries`. `result` could often be used to represent the result of a `Promise` as opposed to the query for the `Promise` itself.
2. The list from lines 228 to 235 seems to have the same contents of `COMPONENT_TYPES`, which we have already defined in another file.
3. This is a personal preference but I think a `map` would be more concise and look better than the `for` loop here.

Together these points could be refactored into the following.

```javascript
  const idSearchQueries = COMPONENT_TYPES.map(component =>
    fetchIds({ component, term, model, version: v, limit })
  );
```

api/src/neo4j/queries/search/modelSearch.js
```diff
@@ -167,48 +211,49 @@ const modelSearch = async ({ searchTerm, model, version, limit }) => {
...
+  ] = await Promise.all(resultFetchIds);
+  const uniqueIds = {
+    ...idsMetabolite,
+    ...idsCompartmentalizedMetabolite,
+    ...idsGene,
+    ...idsReaction,
+    ...idsSubsystem,
+    ...idsCompartment,
+  };
+
```
e0, line 256: This is also a style preference. Since the variables such as `idsMetabolite` are not used beyond `uniqueIds`, we could skip them and have the following instead.

```suggestion
  const ids = await Promise.all(idSearchQueries);
  const uniqueIds = Object.assign({}, ...ids);
```

e0 (approved):
Great work @nanjiangshu! This seems to work really well and the results are much more extensive now.

I have included some refactoring suggestions but they are not absolutely necessary. Please feel free to make changes or ignore as you wish.

nanjiangshu:
Hi @e0, I like your refactoring suggestions and I accept them all. Added in the commit c6c81d1e
Could you and @inghylt take a look at the code again to see that I haven't done any silly mistakes?

e0 (approved):
@nanjiangshu It seems to still work well and all of the tests are passing!

inghylt (approved):
Seems to work well! Also really nice review work from @e0, clean suggestions! :medal_sports:

Issue #1056, opened 2022-07-04
table grouping for GotEnzymes details pages


mihai-sysbio:
**Please describe the feature**

As a user, I want to collapse/group rows in the table, so that I can have a better overview.
_Note_: this issue might require more work, as the grouping is done client side, but the client does not have all the values. The feasibility of this issue should be discussed first. The envision feature is demonstrated in the [documentation of `vue-good-table`](https://xaksis.github.io/vue-good-table/guide/advanced/grouped-table.html).

**Acceptance Criteria**

- [ ] Implement partial matching when filtering
- [ ] The web page documentation is updated accordingly
- [ ] The changes are in alignment with the privacy policy
- [ ] The changes are tested on multiple models and browsers
- [ ] Tests verifying the changes are added
- [ ] The API response is the same as before
- [ ] The changes do not affect the deployment
- [ ] The changes do not reduce the performance

**Additional context**

If we use filtering with "select with search" (example [here](https://vue-multiselect.js.org/#sub-select-with-search)) where the searchable drop down options are the unique results for each column, we could have high performant whole word match

mihai-sysbio:
@mihai-sysbio check again with the researchers

mihai-sysbio:
There should be a research issue to cover the different ideas: in-place multi-select, sidepanel filtering, statistics summary

mihai-sysbio:
Inspiration from https://datanator.info/reaction/XLYOFNOQVPJJNP-UHFFFAOYSA-N,MBPFNOMGYSRGQZ-PBXRRBTRSA-N--%3ENBIIXXVUZAFLBC-UHFFFAOYSA-N,PMMURAAUARKVCB-FTEYRXRINA-N/

![image.png](https://images.zenhubusercontent.com/5e29ab0b256e0f269fabd654/d8cd48fd-09b0-446c-a5ae-dca1f97460c3)

MalinAhlberg:
Improving filtering was also one of the findings from #1154:
Clicking around at different compounds at random + full text fail when filtering + a lot of jumping around

PR #1136, opened 2022-09-14
feat: improve filtering matches

### LINKED ISSUE #1112: Enable progressive keyword filtering (or partial matching) in the GotEnzymes result table

As a user, I want filtering by keywords using partial matching instead of whole word matching so that the filtering of the result table is shortened progressively.

**Acceptance Criteria**

- [ ] Keyword filtering of the result table uses partial matching

**Additional context**

For example, for the page https://metabolicatlas.org/gotenzymes/ec/2.5.1.19, typing in `100` in the field `Gene` results in an empty table, despite that `100` partially matches many of the gene IDs. In many other data table libraries, such as https://datatables.net/, keyword filtering is progressive and shortens the table until no matches are found.

It might be problematic to implement this for all columns. The `Gene` column might contain too many entries for the partial matching to work with resonable performance.

inghylt:
This PR closes #1112 and was created by @e0 and @inghylt 

**Type of change**  

**List of changes made**  

- Using `ILIKE` to allow users to filter the enzymes table using partial and case insensitive matches 

**Screenshot of the fix**  

**Testing**  

- Instructions on how to test
- Visit https://dev.metabolicatlas.org/gotenzymes and play around with different filtering in the enzymes table. Note that on dev the results are cached so don't use the same filter search every time.

**Further comments**  

Please only focus on the response time of applying filters.The response time for gotEnzymes enzymes table is slower than before in general, but it should be possible to separate the queries between populating the entire enzymes table vs applying filters. That will be implemented if the filtering response time is acceptable.

The  filter response time could perhaps be improved by creating index using gist (that would also enable us to try out fuzzy searching as used [here](https://github.com/MetabolicAtlas/MetabolicAtlas/blob/3decdd333e070889890869a3c8162caaaafdda7e/api/src/gotEnzymes/search.js#L5)) but it's not really feasible since it is awfully time consuming. We tried running `create index on enzymes using gist (organism);` and waited for more than an hour and it didn't finish. Image running it for gene, reaction, compound, ec and domain as well, redeploying would take forever.

@e0 noted that if we use filtering with "select with search" (example [here](https://vue-multiselect.js.org/#sub-select-with-search)) where the searchable drop down options are the unique results for each column, we could have the same high performant whole word match as before. That's one of the possible implementations of #1056. If you think the filter performance of this PR it too slow, which makes sense, maybe we should close this issue and work on #1056 directly? Or should we go ahead with the current implementation and ask the researches if they think it is good enough?

Another idea might be to use `ILIKE` only on some columns, but that might result in confusing UX

### DIFF
--- a/api/src/gotEnzymes/enzyme.js
+++ b/api/src/gotEnzymes/enzyme.js
@@ -19,11 +19,9 @@ const getFiltersQueries = filters => {
 
     if (value && value.length > 0) {
       if (field === 'ec_number') {
-        filtersQueries.push(
-          sql`array[${sql`${value.toString()}`}] <@ ${sql`string_to_array(ec_number, ';')`}`
-        );
+        filtersQueries.push(sql`ec_number ~~* ${`%${value.toString()}%`}`);
       } else {
-        filtersQueries.push(sql`${sql(field)} = ${value.toString()}`);
+        filtersQueries.push(sql`${sql(field)} like ${`%${value.toString()}%`}`);
       }
     }
   }
--- a/api/src/gotEnzymes/enzyme.js
+++ b/api/src/gotEnzymes/enzyme.js
@@ -19,9 +19,11 @@ const getFiltersQueries = filters => {
 
     if (value && value.length > 0) {
       if (field === 'ec_number') {
-        filtersQueries.push(sql`ec_number ~~* ${`%${value.toString()}%`}`);
+        filtersQueries.push(sql`ec_number ilike ${`%${value.toString()}%`}`);
       } else {
-        filtersQueries.push(sql`${sql(field)} like ${`%${value.toString()}%`}`);
+        filtersQueries.push(
+          sql`${sql(field)} ilike ${`%${value.toString()}%`}`
+        );
       }
     }
   }
--- a/project-words.txt
+++ b/project-words.txt
@@ -77,6 +77,7 @@ Hotjar
 Hotjar's
 hoverable
 Hyltander
+ilike
 inchi
 Intawat
 jaccard


nanjiangshu:
After testing a few cases on my localhost, my generally feeling is that the performance is too slow, it usually takes >10 seconds to load a page and to to a filtering. For filtering, even when there are only a few rows left in the table, it still takes >10 seconds after typing in one more letter in a filtering field. I think this might be improved by filtering from the content of tables instead of fetching the data from the backend again.  For initial partial, case insensitive searching, I think we can ignore some columns if the performance can not be improved.

inghylt:
@mihai-sysbio Writing here since you won't attend any standups this week:

The changes introduced in this PR allows partial and case insensitive matches. However, this reduced performance significantly, too much to be acceptable. We can't speed things up, or try out fuzzy searching as used [here](https://github.com/MetabolicAtlas/MetabolicAtlas/blob/3decdd333e070889890869a3c8162caaaafdda7e/api/src/gotEnzymes/search.js#L5), by using creating new indexes since the creation of these would take ages. We tried running `create index on enzymes using gist (organism);` and waited for more than an hour and it didn't finish. This would affect both local development and deploying to prod. 

I have also looked at not applying `ilike` on genes and instead keeping the old query but that did not make any improvements. We have also discussed if a solution could be to trigger the first search query in the background, i.e. the full enzyme table without any filters and not paginate the results and once that is fully loaded one can just run filter queries on that result directly. That would still entail very slow filtering until the background search query is done, so likely bad and confusing UX.

@e0 noted that if we use filtering with "select with search" (example [here](https://vue-multiselect.js.org/#sub-select-with-search)) where the searchable drop down options are the unique results for each column, we could have the same high performant whole word match as before this PR. That's one of the possible implementations of #1056. Thus, I suggest we close this issue and try to implement partial matches in #1056 instead

mihai-sysbio:
> Thus, I suggest we close this issue and try to implement partial matches in https://github.com/MetabolicAtlas/MetabolicAtlas/issues/1056 instead

Thanks for documenting the summary, sounds like a good plan so please go ahead

PR #1138, opened 2022-09-19
feat: case insensitive search/filtering for gotenzymes

### LINKED ISSUE #1110: Case insensitive searching and filtering for GotEnzymes

As a user, I want case insensitive searching and filtering so that so that I don't need to always pay attention to the case of the text I'm typing. 

**Acceptance Criteria**

- [ ] Landing page searching is case insensitive
- [ ] Filtering in the result table is case insensitive

**Additional context**

Case insensitive searching is also implemented in the Gem search and Global search for integrated models. 

It might be that there are genes that have the same the name when doing a case-insensitive search.

MalinAhlberg:
This PR closes #1110  and relates to #1112 .

Searching on the GotEnzymes landing page is now case insensitive. Filtering columns of the enzymes table is also case insensitive.

**Type of change**  

**List of changes made**  

- use lower case for the db queries when searching / filtering
- add indixes for lower cased fields to the psql db
- add api tests

**Testing**  

To create the new indexes, either reconstructing the psql db:
`ma-exec pg psql -f scripts/init.sql -U postgres`
or by do this :
` psql -h localhost -p 5432 -U postgres`
(use password from `env-local.env`) and then run:
```
  create index lower_cased_genes on genes ((lower(kegg)));
  create index lower_genes on enzymes ((lower(gene))) ;
  create index lower_domain on enzymes (lower(domain));
  create index lower_organism on enzymes (lower(organism));
  create index lower_compound on enzymes (lower(compound));
  create index lower_reaction on enzymes (lower(reaction_id));
```
This takes some minutes.

- Go to `/gotenzymes` and verify that case is not important when searching, for all catergories (genes, domains, organisms, reactions, compouns). For example, search for `atp6` and make sure that 2 genes are listed in the results
- Visit a details page, eg `/gotenzymes/reaction/R01246` and try filtering by gene `acav_1835`. The resulting table should list `Acav_1835`.
- Also try `/gotenzymes/domain/B` (this is a bit slow, see *Further comments* below). Filter by `b1468` and make sure you see both `b1468` and `B1468`.

**Further comments**  

For most cases, the performance is good on my local machine. The slowest page I found is `/gotenzymes/domain/B`. It is also the one containing the most rows. What actually makes it slow is the total count, which the db for some reason decides to handle quite differently from before (`Parallel Bitmap Heap Scan` instead of `Parallel Index Only Scan`). If the speed is a problem, here are some thoughts:
- do some more testing / researching
- have a separate call for getting the total count so that the first page can be shown as soon as possible
- have a separate call for creating the initial enzymes table for a details page, that uses exact match search. We know that the initial filter (eg `domain = 'B'`) will have the correct case.
- the filtering of the columns are subject to change in #1056.


inghylt (approved):
~Nice work, most functionality seems to work well! The sorting arrows above the columns filters seems to be broken though~ this should have been a request change :')

inghylt (changes requested):
Nice work, most functionality seems to work well! The sorting arrows above the columns filters seems to be broken though

api/src/gotEnzymes/enzyme.js
```diff
@@ -9,6 +9,7 @@ const MATCH_FIELDS = [
   'compound',
 ];
 
+const CASED_FIELDS = ['gene', 'organism', 'domain', 'compound', 'reaction_id'];
```
inghylt, line 12: Just curious, why can't we use `MATCH_FIELDS` instead?

MalinAhlberg:
> The sorting arrows above the columns filters seems to be broken though

Good that you noticed! Fixed in 4cb4e28.

api/src/gotEnzymes/enzyme.js
MalinAhlberg: `MATCH_FIELDS` contains `ec_number`, which does not contain letters. So to avoid unnecessary computations / creating more indexes, we don't want it to be compared in lower case (without an extra index, it would make a search for `ec_numbers` ~4 times slower (on my computer)).

inghylt: Ok! I'm just thinking that on line 26 we should be able to use `MATCH_FIELDS` directly since `field === 'ec_number'` is checked on line 22 right? So one could probably add an if condition to line 78 as well and then we wouldn't need `CASED_FIELDS`. But maybe it's cleaner keeping things as is, up to you!

inghylt: @e0 @nanjiangshu Any thoughts? :)

inghylt:
The current state of the branch is deployed to https://dev.metabolicatlas.org/gotenzymes/

inghylt (approved):
Nice work! Since things will be cached on dev/prod, `gotenzymes/domain/B` will be fast for all but the first user. And the non cached response time was not too slow IMO so I approve!

api/src/gotEnzymes/enzyme.js
nanjiangshu: I think it's cleaner to separate them.

nanjiangshu (approved):
I verified the caching at https://dev.metabolicatlas.org/gotenzymes/reaction/R01677 and I will approve this PR now.

MalinAhlberg:
1d2b8d6 fixes a previously introduced unrelated error, as discussed on Slack.

nanjiangshu:
I approved it again after the cspell commit

PR #1139, opened 2022-09-19
feat: gotenzymes formatted formulas

### LINKED ISSUE #1113: Format chemicals in formulas and equations for GotEnzymes

As a user, I want chemicals in formulas and equations for GotEnzymes are formatted so that they are easier to read.

**Acceptance Criteria**

- [ ] Chemicals in formulas and equations are formatted as in Gem browser

**Additional context**

The formatting in the GEM Browser is not based on a string but on an object. For example, we know that the charge information is taken into account when formatting the formula. Also, the formatting of the reactions might not be reusable.

MalinAhlberg:
This PR closes #1113.

**Type of change**  

**List of changes made**  

- Formulas for compounds are formatted (using existing code)
- Equations for reactions are formatted (new code)

**Screenshot of the fix**  

2022-09-19-162741_845x173_scrot

2022-09-19-165015_419x172_scrot

**Testing**  

- Inspect the formula at `gotenzymes/compound/C00097`
- Inspect the equation at `gotenzymes/reaction/R01677`
- Inspect the new function [`chemicalEquation`](https://github.com/MetabolicAtlas/MetabolicAtlas/blob/4e94d7bfd02a3c932522e52e6f2fe41cf7871e44/frontend/src/helpers/chemical-formatters.js#L22)
- Please browse different details pages under GotEnzymes to make sure everything looks ok.

**Further comments**
- No tests are added since this is frontend code, so the acceptance criteria are not fulfilled.
- Do we want links from the equation sections, as for the GEM browser?
2022-09-19-162841_821x160_scrot


frontend/src/helpers/chemical-formatters.js
```diff
@@ -17,4 +17,15 @@ export default function chemicalFormula(formula, charge) {
...
   return form;
-}
+};
+
+const chemicalEquation = formula => {
+  let equation = formula.replace('<=>', '⇔');
+  equation = equation.replace('=>', '⇒');
+  equation = equation.replaceAll(/(\S)\+/g, '$1<sup>+</sup>');
+  equation = equation.replaceAll(/(\S)-\s/g, '$1<sup>-</sup> ');
+  equation = equation.replaceAll(/([A-Z])(\d+)/g, '$1<sub>$2</sub>');
```
nanjiangshu, line 27: This may bring in some unwanted formatting, e.g. for the reaction `R00073`, the number 2 for the compound `Prostaglandin G2` should not be `sub`. Basically, only a few chemicals such as H2, O2 and H2O should be formatted but it is hard to include everything wanted while excluded unwanted. Maybe we can just ignore this line of formatting.

frontend/src/helpers/chemical-formatters.js
```diff
@@ -17,4 +17,15 @@ export default function chemicalFormula(formula, charge) {
     form = form.replace(/([-+])$/, '<sup>$1</sup>');
   }
   return form;
-}
+};
+
+const chemicalEquation = formula => {
+  let equation = formula.replace('<=>', '⇔');
+  equation = equation.replace('=>', '⇒');
+  equation = equation.replaceAll(/(\S)\+/g, '$1<sup>+</sup>');
```
nanjiangshu, line 25: Fe2+ in the equation for `/gotenzymes/reaction/R00078` is also incorrectly formatted. Maybe we can ignore formatted of the charge as well.

e0: Comparison of GotEnzymes page vs. GEM Browser page below. I assume there is not much we can do about the order, `Oxygen` vs. `O2`, and the missing compartments. Is the only problem here that some numbers are not wrapped in `sup` tags?

image
image

frontend/src/helpers/chemical-formatters.js
e0: It seems that we may have the same problem in the GEM Browser, for example: `/explore/Human-GEM/gem-browser/reaction/MAR04223`. If this is the case, I'm ok with it being wrong in the same way in the GotEnzymes page for now, then we could create a new bug issue to track this problem for both pages.

image

frontend/src/helpers/chemical-formatters.js
MalinAhlberg: ec75ed5 improves this to 
2022-09-20-131447_505x64_scrot

frontend/src/helpers/chemical-formatters.js
MalinAhlberg: That's tricky! And a nice observation that we have the same problem in the GEM Browser. To me, keeping it and creating a new issue would make sense then, if @nanjiangshu does not object.

MalinAhlberg:
The failing CSpell test is not related to this branch (but to the error in `main`) and will be fixed when #1138 is merged.

frontend/src/helpers/chemical-formatters.js
nanjiangshu: I'm OK with it. Let's keep it as is and create a new issue to solve it for both here and the GEM browser.

frontend/src/helpers/chemical-formatters.js
nanjiangshu: This commit solves this problem but may bring in some new problems. But I guess the new problems will be in the same category as mentioned in the comment https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1139#discussion_r975111855. I can approve it and then we leave the problem in the new issue. Mostly some edge cases.

nanjiangshu (approved):

frontend/src/helpers/chemical-formatters.js
MalinAhlberg: I created #1140 for this. Please fill in more examples if you know of any.

e0 (approved):
LGTM!

PR #1146, opened 2022-09-26
fix: use the value of the param

### LINKED ISSUE #914: Unexpected behaviour of the param `full` for the api endpoints `/compartments`
**Describe the bug**

The API endpoint `/api/v2/compartments/` does not work as expected.

For example: 
The parameter `full` must be specified for the `compartments` API in order to retrieve all the metabolites, genes, reactions and subsystems contained by the given compartment. Without this parameter, i.e. the endpoint `/api/v2/compartments/golgi_apparatus?model=HumanGem`
does not return information of genes, reactions and metabolites.

However, how this parameter `full` should be set is unexpected.  
The following two endpoints return 
`/api/v2/compartments/golgi_apparatus?model=HumanGem&full=false`
`/api/v2/compartments/golgi_apparatus?model=HumanGem&full=true`
return the same result containing information of all metabolites, genes, reactions and subsystems. 

While
`/api/v2/compartments/golgi_apparatus?model=HumanGem&full` and
`/api/v2/compartments/golgi_apparatus?model=HumanGem`
return the same result but without information of genes, reactions and metabolites. 

**Steps to reproduce**

**Expected behavior**

The returned result should be meaningful depending on the value of the parameter `full`. 

The following cases should be considered: 
1. `/api/v2/compartments/golgi_apparatus?model=HumanGem&full=true`  
2. `/api/v2/compartments/golgi_apparatus?model=HumanGem&full=false` 
3. When the parameter `full` is not set, i.e. `/api/v2/compartments/golgi_apparatus?model=HumanGem`, what should be the default value
4. When the parameter `full` is set to something else than `true` or `false`, i.e. `/api/v2/compartments/golgi_apparatus?model=HumanGem&full=somethingelse`  

**Screenshots**

**Additional context**

MalinAhlberg:
This PR closes #914.

The param `full` in `/compartmets/{id}`, will be false when set to a falsy value or when left out.

**Type of change**  

**List of changes made**  

- Updated the handling of the parameter
- Updated the swagger docs

**Screenshot of the fix**  

**Testing**  

- Take a look at the updated swagger description
- Test that:
  - `api/v2/compartments/golgi_apparatus?model=HumanGem&full=true` give full results
  - `api/v2/compartments/golgi_apparatus?model=HumanGem&full` give full results
  - `api/v2/compartments/golgi_apparatus?model=HumanGem` give short results
  - `api/v2/compartments/golgi_apparatus?model=HumanGem&full=false` give short results

**Further comments**  

See [slack thread](https://nbisweden.slack.com/archives/C01D13EKUV7/p1663840452835189?thread_ts=1663838914.788989&cid=C01D13EKUV7) for a discussion of how the parameter should behave.

### DIFF
--- a/api/src/neo4j/queries/compartment.js
+++ b/api/src/neo4j/queries/compartment.js
@@ -3,6 +3,9 @@ import parseParams from 'neo4j/shared/helper';
 
 const getCompartment = async ({ id, model, version, full }) => {
   const [m, v] = parseParams(model, version);
+  // full is undefined when the param is not present, otherwise a string
+  const falsy = ['false', '""', '0', 'null', 'undefined', 'NaN'];
+  const showFull = !falsy.includes(`${full}`.toLowerCase());
 
   const statement = `
 CALL apoc.cypher.run("
@@ -16,10 +19,10 @@ CALL apoc.cypher.run("
   RETURN {
     subsystems: COLLECT(DISTINCT({id: s.id, name: ss.name})) ,
     reactionsCount: COUNT(DISTINCT(r)) ${
-      full ? ', reactions: COLLECT(DISTINCT(r))' : ''
+      showFull ? ', reactions: COLLECT(DISTINCT(r))' : ''
     },
     metabolitesCount: COUNT(DISTINCT(cm)) ${
-      full ? ', metabolites: COLLECT(DISTINCT(cm))' : ''
+      showFull ? ', metabolites: COLLECT(DISTINCT(cm))' : ''
     }
   } as data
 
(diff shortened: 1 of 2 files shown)


e0 (approved):
Seems to work nicely. Nice job! Great that you updated the swagger docs as well.

inghylt (approved):
Nice, seems to work well!

`/api/v2/compartments/golgi_apparatus?model=HumanGem&full=silly` gives a full result, which IMO is expected, but just wanted to raise that example as well

nanjiangshu (approved):
Looks great and all tests works well. One could argue whether `/api/v2/compartments/golgi_apparatus?model=HumanGem&full=silly` should return shortened or full result but I agree with @inghylt it's fine with the current setup. Moreover, the description in the Swagger doc is clear enough.

e0:
Since it does not seem too clear, here is a bit more about the motivation of only returning shortened results when the value for `full` is false.

This endpoint has a default behavior where the shortened result are returned and `full` is an optional query param. Therefore I think it's a better bet that when `full` is actually included, it's more likely that the user is looking for an alternative that is not the default, which is the full result. The exception is that when the user explicitly sets a falsy value for `full`, where the shortened result is returned again.

PR #1147, opened 2022-09-26
fix: duplicates in gem search

### LINKED ISSUE #1124: duplicated results in Gemsearch
**Describe the bug**

Gemsearch shows duplicated results when searching for e.g. `Retinol metabolism`

**Steps to reproduce**

In Gemsearch, select `Human-GEM` and search for the keyword `Retinol metabolism`.  The hit `MAM02834c` is duplicated as shown in the screenshot.

**Expected behavior / Acceptance criteria**

- [ ] search results that match 100% with the model content in the yml in the `data-files` repository
- [ ] The results shown in Gemsearch are unique 

**Screenshots**

**Additional context**

The reason why two hits for `MAM02834c` are shown is because the `mid` of the two hits are different. However, since these two hits show the same information in the frontend and more importantly, they direct to the same metabolite page, only one of them should be shown. 
```
{
  name: 'retinol',
  compartment: 'Cytosol',
  mid: 'retinol',
  id: 'MAM02834c',
  score: 9.824196815490723,
  component: 'metabolites'
},
{
  name: 'retinol',
  compartment: 'Cytosol',
  mid: '11_cis_retinol',
  id: 'MAM02834c',
  score: 6.724960803985596,
  component: 'metabolites'
}, 
```

If the problem is in the source yml, it should then be reported in the corresponding repository.

MalinAhlberg:
This PR closes #1124.

**Type of change**  

**List of changes made**  

The bug was in how the neo4j query was done, where the `mid`s would get linked to the wrong names.

**Testing**  

Search for `Retinol metabolism` in the gem search and verify that you only get each item listed once.
Please also test other cases, and inspect the neo4j query.

**Further comments**  

The same bug also appears in the global search. See `search?term=retinol+metabolism`, click the `Metabolites` tab and filter by Model, so that you only see the Human-GEM.
I started on a fix for this, but it requires some more testing.

### DIFF
--- a/api/src/neo4j/queries/search/modelSearch.js
+++ b/api/src/neo4j/queries/search/modelSearch.js
@@ -60,6 +60,11 @@ const fetchCompartmentalizedMetabolites = async ({
   if (!ids) {
     return null;
   }
+  // create a neo4j mapping of the ids to null (null representing a metaboliteId)
+  // regex replacement needed for neo4j to accept the object
+  const mappedIds = JSON.stringify(
+    ids.map(x => ({ mid: null, cmid: x }))
+  ).replace(/"([^"]+)":/g, '$1:');
 
   let statement = `
 WITH ${JSON.stringify(metaboliteIds)} as mids
@@ -70,23 +75,24 @@ UNWIND
   END AS mid
 OPTIONAL MATCH (:Metabolite:${model} {id:mid})-[${version}]-(cm:CompartmentalizedMetabolite)
 WITH cm.id as cmid1, mid
-WITH ${JSON.stringify(ids)} as cmids2, cmid1, mid
-WITH collect(cmid1)+cmids2 as cmids, mid
+WITH ${mappedIds} as cmids2, cmid1, mid
+WITH cmids2+collect({mid: mid, cmid: cmid1}) as cmids
+
 UNWIND cmids as cmid
 
 CALL apoc.cypher.run('
-  MATCH (ms:MetaboliteState)-[${version}]-(:Metabolite)-[${version}]-(:CompartmentalizedMetabolite:${model} {id: $cmid})
-  RETURN { id: $cmid, name: ms.name } as data
+  MATCH (ms:MetaboliteState)-[${version}]-(:Metabolite)-[${version}]-(:CompartmentalizedMetabolite:${model} {id: $cmid.cmid})
+  RETURN { id: $cmid.cmid, name: ms.name, mid: $cmid.mid } as data
   
   UNION
   
-  MATCH (:CompartmentalizedMetabolite:${model} {id: $cmid})-[${version}]-(c:Compartment)-[${version}]-(cs:CompartmentState)
-  RETURN { id: $cmid, compartment: cs.name } as data
+  MATCH (:CompartmentalizedMetabolite:${model} {id: $cmid.cmid})-[${version}]-(c:Compartment)-[${version}]-(cs:CompartmentState)
+  RETURN { id: $cmid.cmid, compartment: cs.name } as data
 ', {cmid:cmid}) yield value
 WITH apoc.map.mergeList(apoc.coll.flatten(
 	apoc.map.values(apoc.map.groupByMulti(COLLECT(value.data), "id"), [value.data.id])
-)) as metabolites, mid
-RETURN metabolites {mid: mid, .*}
+)) as metabolites
+RETURN metabolites
 `;
   // limit is applied here again since the number of IDs for Metabolite and
   // CompartmentalizedMetabolite together may exceed the value of limit


inghylt (approved):
Nice work! I can't say that I spontaneously understand why the regex is written as it is, but I don't think that's something that needs to be further documented either

nanjiangshu:
When playing with the neo4j querying, I found this problem with the `WITH` statement 

I have a question regarding how WITH works in neo4j.
The following two queries outputs different results, however, the only difference of the two queries is that the Query1 carried over cmid1 at line 11.  In the Documentation https://neo4j.com/docs/cypher-manual/current/clauses/with/, it says one can even use the * to carry over all variables but it does not mention that carry over one variable may affect the result of another.  Are there any good explanations for that?
Query1:
```neo4j
     1	WITH  ['retinol', '11_cis_retinol'] as mids
     2	UNWIND
     3	  CASE
     4	      WHEN mids = [] THEN [null]
     5	      ELSE mids
     6	  END AS mid
     7	OPTIONAL MATCH (:Metabolite:HumanGem
     8	{id:mid})-[]-(cm:CompartmentalizedMetabolite:HumanGem)
     9	WITH cm.id as cmid1, mid
    10	WITH  ["MAM02834e", "MAM02834c", "MAM02834r"] as cmids2, cmid1, mid
    11	WITH collect(cmid1) as collect_cmid1, cmids2, cmid1, mid
    12	WITH collect_cmid1+cmids2 as cmids, mid
    13	return cmids
```
Output1:
```
["MAM02834r", "MAM02834e", "MAM02834c", "MAM02834r"]
["MAM02834c", "MAM02834e", "MAM02834c", "MAM02834r"]
["MAM02834e", "MAM02834e", "MAM02834c", "MAM02834r"]
["MAM00291e", "MAM02834e", "MAM02834c", "MAM02834r"]
["MAM00291c", "MAM02834e", "MAM02834c", "MAM02834r"]
```
Query2:
```neo4j
     1	WITH  ['retinol', '11_cis_retinol'] as mids
     2	UNWIND
     3	  CASE
     4	      WHEN mids = [] THEN [null]
     5	      ELSE mids
     6	  END AS mid
     7	OPTIONAL MATCH (:Metabolite:HumanGem
     8	{id:mid})-[]-(cm:CompartmentalizedMetabolite:HumanGem)
     9	WITH cm.id as cmid1, mid
    10	WITH  ["MAM02834e", "MAM02834c", "MAM02834r"] as cmids2, cmid1, mid
    11	WITH collect(cmid1) as collect_cmid1, cmids2, mid
    12	WITH collect_cmid1+cmids2 as cmids, mid
    13	return cmids
```
Output2:
```
["MAM02834r", "MAM02834c", "MAM02834e", "MAM02834e", "MAM02834c", "MAM02834r"]
["MAM00291e", "MAM00291c", "MAM02834e", "MAM02834c", "MAM02834r"]
```

I moved this discussion from Slack to here so that it's easier to keep track of it. 

I was confused by the behavior of `WITH` at the beginning since at nowhere in the neo4j documentation mentions how carrying over variables may change the result of querying. After discussing with @shan this morning, I realized that when a variable is carried over in the `WITH` clause, the results will be categorized by this variable even it’s not to used in the following RETURN statement. Therefore, the result of `Query1` has five rows since `cmid1` has five rows. Then the use of aggregation `collect` merged all result to one row if there is no carrying over variables, but if there are carrying over variables, like what the line `11` is doing, the result will be categorized again by the number of rows of the other carrying over variables. 

I think the reason why duplicates are returned in the old solution is that the metIDs (`MAM...`) retrieved by metabolite names (e.g. `retinol`) are combined with the three metIDs, i.e. ["MAM02834e", "MAM02834c", "MAM02834r"] without the metabolite name. After that, the combined metIDs (`MAM...`) are treated in the same way to search in the database then in the final stage the mid are carried over.   
```
WITH apoc.map.mergeList(apoc.coll.flatten(
	apoc.map.values(apoc.map.groupByMulti(COLLECT(value.data), "id"), [value.data.id])
)) as metabolites, mid
RETURN metabolites {mid: mid, .*}   
```

I think the solution @MalinAhlberg made makes sense. It assigns a `null` mid to the  three metIDs ["MAM02834e", "MAM02834c", "MAM02834r"] so that they can be distinguished from the metIDs retrieved by searching the metabolite names (represented by `cmid1` in the query above) .

nanjiangshu:
I can already approve this PR but want to ask if we should fix the duplication problem in this PR or in a new issue?

MalinAhlberg:
Thanks for investigating and for documenting, @nanjiangshu :clap: !

Regarding the global search, both options are fine with me, but I will probably not spend much time on this until next sprint, so if we decide to solve it in this PR, the PR will likely be left hanging.

nanjiangshu:
> Regarding the global search, both options are fine with me, but I will probably not spend much time on this until next sprint, so if we decide to solve it in this PR, the PR will likely be left hanging.
It makes sense to me to solve the duplication problem for the global search in a new issue, just remember to refer to this issue.

nanjiangshu (approved):

PR #1182, opened 2022-10-27
feat: reallocate prod vm resources

### LINKED ISSUE #1180: update resource allocation for new prod VM

As a [type of user], I want [an action] so that [a benefit/a value].
As a developer, I want to have an updated resource allocation for the new production VM, so that when deploying the stack can take full advantage of the new resources.

**Acceptance Criteria**

- [ ] The changes do not affect the deployment process

**Additional context**

The [overview of VMs is described here](https://github.com/MetabolicAtlas/docs/wiki/VM-overview).

mihai-sysbio:
This PR closes #1180.

**Type of change**  

**List of changes made**  

Increasing the limits set through Docker compose.

**Testing**  

- Instructions on how to test: check the new production server at https://metatlas-prod3.csbi.chalmers.se/

**Further comments**  

The [`reservations`](https://docs.docker.com/compose/compose-file/deploy/#resources) were removed because for our deployment process they don't serve a purpose - the target VMs don't change that often, and when they do it's a resource increase.

### DIFF
--- a/docker-compose-remote.yml
+++ b/docker-compose-remote.yml
@@ -20,21 +20,17 @@ services:
     restart: unless-stopped
     deploy:
       resources:
-        reservations:
-          memory: 0.5G
         limits:
-          memory: 1G
+          memory: 1.5G
 
   api:
     command: yarn start
     restart: unless-stopped
     deploy:
       resources:
-        reservations:
-          memory: 0.5G
         limits:
-          cpus: "0.30"
-          memory: 1G
+          cpus: "0.40"
+          memory: 1.5G
 
   neo4j:
     environment:
@@ -44,11 +40,9 @@ services:
     restart: unless-stopped
     deploy:
       resources:
-        reservations:
-          memory: 9G
         limits:
-          cpus: "0.95"
-          memory: 10G
+          cpus: "1.40"
+          memory: 13G
     ulimits:
       nofile:
         soft: 80000
@@ -81,8 +75,8 @@ services:
         reservations:
           memory: 3G
         limits:
-          cpus: "0.50"
-          memory: 3G
+          cpus: "0.70"
+          memory: 5G
 
 volumes:
   letsencrypt:


nanjiangshu (commented):
If I remember correctly, the capacity for the production server is more than the dev server `dev-metatlas`. Shall we not set a higher limitation for the production server? 
If we don't want to keep two versions of constrains for the `prod` and `dev` server, maybe we can just remove the constrain for the dev server at all.

nanjiangshu (approved):
Looks good. More discussions can be found in the slack thread https://nbisweden.slack.com/archives/C01D13EKUV7/p1666940818050709

e0:
Performance seems ok from what I tested. The 3d map for cytosol took ~56s to load though.

Since it's not clear atm how this impacts the dev server, which has less than the allocated resources, I think we should hold off on merging this PR until it has been deployed to the dev server and tested. If anything breaks there, we should consider updating this PR to include different allocations for different deploy targets (contexts).

mihai-sysbio:
> Since it's not clear atm how this impacts the dev server, which has less than the allocated resources, I think we should hold off on merging this PR until it has been deployed to the dev server and tested.

This branch has been deployed to the `dev` server without a hitch. `docker stats` doesn't show any weird allocation, except that the sum of the limits is higher than what this VM can handle. Please test (and re-deploy if you want to).

mihai-sysbio:
> we should consider updating this PR to include different allocations for different deploy targets (contexts)

It's a good idea if we have more production servers. For the `dev` servers it's fine if there are no guarantees. I think I've heard the idea that we could have a banner at the top of the page just to let users know that they are looking at the `dev` server.

MalinAhlberg (approved):
This seems to work well on the new production server! On the dev server, I could at first not load the cytosol 3d map, but got a 504 (gateway time-out). It worked a few minutes later though, so I guess it just needed more time.
Also tried redeploying to dev, which worked fine.

mihai-sysbio:
> at first not load the cytosol 3d map, but got a 504 (gateway time-out). It worked a few minutes later though, so I guess it just needed more time

That's a typical behaviour - that particular request is very resource-intensive, and it takes a little for Neo4j to rev its engines. And after the first successful request, the caching kicks in.

PR #1190, opened 2022-11-02
fix: use stop instead of kill to gracefully shutdown containers

### LINKED ISSUE #1131: neo4j container sometimes exits after restarting
**Describe the bug**

It seems that sometimes when restarting the stack, the neo4j container would fail to start. Currently it's unclear what the issue is.

**Work-around**
When this happens, deleting the neo4j container and running `start-stack` again would solve the issue. This is usually pretty fast as well since the data does not need to be re-imported.

**Steps to reproduce**

It's a bit random how this can be reproduced but usually running `stop-stack` followed by `start-stack` would cause it to exit.

**Expected behavior**

- [ ] Running `start-stack` should always work for the neo4j container.
- [ ] Verify that deployment is not broken from the changes

**Additional context**

This might be a related issue: https://stackoverflow.com/questions/73370705/why-neo4j-docker-container-restart-causes-the-container-to-hang-and-quit

I have also tested this on neo4j version 4.1.11 but it did not solve the issue.

Research between `down`, `stop`, and `kill`.

e0:
This PR closes #1131.

**Type of change**  

**List of changes made**  

Use `docker compose stop` instead of `docker compose kill` when running `stop-stack`. This sends a graceful shutdown message `SIGTERM` instead of `SIGKILL`.

After applying this change, I restarted the stack several times without `neo4j` failing to start, whereas before it was failing almost every other time.

This is a quick little read along with an informative but heart-breaking comic that illustrates the differences: https://linuxhandbook.com/sigterm-vs-sigkill/

**Screenshot of the fix**  

**Testing**  

- Instructions on how to test

After `source proj.sh`, restart (`stop-stack` followed by `start-stack`) a few times and verify that `neo4j` always starts without problems.

**Further comments**  

Relevant docs:
- https://docs.docker.com/engine/reference/commandline/compose_down/
- https://docs.docker.com/engine/reference/commandline/compose_kill/

### DIFF
--- a/proj.sh
+++ b/proj.sh
@@ -33,7 +33,7 @@ function start-stack {
 }
 
 function stop-stack {
-  docker compose --env-file $CHOSEN_ENV -f docker-compose.yml -f docker-compose-local.yml kill
+  docker compose --env-file $CHOSEN_ENV -f docker-compose.yml -f docker-compose-local.yml stop
 }
 
 function clean-stack {


MalinAhlberg (approved):
Nice! I also tested multiple times, and no complaints from the neo4j container :star:. 

Also very good (but sad :crying_cat_face: ) comic provided :D.

inghylt (approved):
Ah, great fix! And 10/10 comic, would be nice to have comics as a mandatory part of the PR description in the future : D

nanjiangshu (approved):
Looks good!

PR #1202, opened 2022-11-04
fix: pass on env vars to multistage docker image

### LINKED ISSUE #1201: missing variable for analytics
**Describe the bug**
In production, the requests to Matomo (analytics) fail because the `siteId` is undefined.

**Steps to reproduce**
Using a browser that does not send the `Do Not Track` header, inspect the network requests, and reload the page. A request to `csbi.chalmers.se` with a long url will fail.

**Expected behavior**

- [ ] the `siteId` should have the value defined in the `.env` file instead of `undefined`

mihai-sysbio:
This PR closes #1201 by updating the name of the environment variables in the multistage image.

**Type of change**  

**List of changes made**  

Change from `VUE_APP_*` to `VITE_VUE_APP_*`, and removing these from the `local` docker compose.

**Testing**  

- Instructions on how to test: on a remote server, using a browser that does not send the Do Not Track header, inspect the network requests, and reload the page. A request to csbi.chalmers.se contain the same `siteId` as in the `.env` file.

**Further comments**  


MalinAhlberg (approved):
LGTM!

docker-compose-local.yml
```diff
@@ -15,9 +15,6 @@ services:
     container_name: frontend
     build: frontend
     command: yarn serve
-    environment:
-      - VITE_VUE_APP_MATOMOID=${VITE_VUE_APP_MATOMOID}
-      - VITE_VUE_APP_HOTJAR=${VITE_VUE_APP_HOTJAR}
```
e0, line 20: Is there never a case where the developer wants to test these locally?

mihai-sysbio: To the best of my knowledge, it would be detrimental to have these analytics services running locally.

For `MATOMOID` there would have to be a separate side identifier, otherwise the stats would get mixed between `localhost` and `production` which would be bad. And the stats generated for this new side identifier would be irrelevant. So, in practice, I would say that no local testing is relevant.

For `HOTJAR`, the benefit of getting user feedback is relevant only for real users. For `localhost` I cannot think of any advantage in having HotJar record any sessions.

e0: Don't we already have different identifiers for `dev` and `prod` for `MATOMOID`? I was thinking we could use the `dev` one to test locally.

mihai-sysbio: There still seems to be a `dev` siteId. What are you thinking of using it for - what kind of testing?

e0: In general it's good practice to be able to have access to all parts of the project locally. By default I think most of us leave these env vars unset locally. So I don't really understand why the possibility should be taken away.

mihai-sysbio: This part of the project would either be not used (no variables configured in the local environment), or generate unneeded requests that would not bring any useful information. Moreover, a new developer could wonder why are tracking features enabled locally.

> why the possibility should be taken away

This possibility was introduced only 2 months ago in this [commit](https://github.com/MetabolicAtlas/MetabolicAtlas/commit/a0d0f588639b701e58b2f77528eec07a6e5fec55). So far, to me, its presence has not added any benefits to the project, so it mainly resembles dead code from my point of view.

It seems to feel strongly about it, so even if I really don't follow why, I am fine reverting the deletion.

e0: Sorry, I will try to be a bit more patient in this explanation. This along with a lot of other best practice stuff, such as including the packages lock file, has become obvious to me but it's definitely still a valid question.

For both matomo and hotjar, it could be useful to have them locally to be able to test how new features are tracked without having to deploy them. For example when creating a PR, it could be useful for the PR author to view the recordings of the PR reviewers to help assess the UX.

More importantly, if any code changes are done that impacts the implementation of matomo or hotjar directly, it would be ideal to validate that they are working without having to deploy it to the dev server. We do have instances where we need the dev server to be at a certain state for preliminary paper reviewing for example.

I feel strongly about things like this because it's best practices that I have learnt and I think they should be adopted unless there are really convincing arguments against it.

mihai-sysbio: Thanks for laying out the arguments. While I agree with them in principle, as a developer working with the stack I have not felt the need for these features: there isn't enough time to assess the PR review through HotJar; Matomo does not bring benefits in this setup; deployment to development should be performed regardless.

Commit 03ca8521 has reverted the removal, so that this PR can be merged.

e0: I rarely have to test these either locally.

The core principle here is that one should be able to test as much as possible locally without having to deploy.

e0 (approved):

PR #1222, opened 2022-11-21
fix: use search terms that produce rows with unique scores

### LINKED ISSUE #1143: gotenzymes api tests failing
**Describe the bug**

The tests introduced in #1138 sometimes fail.

**Steps to reproduce**

- Run the tests multiple times, on different machine until they fail.

 It seems to be some randomness involved since this does not happen every time. The scores for many of the search results are exactly the same, which affects the ordering of the results and therefore the comparison.

**Expected behavior**

- [ ] Tests never fail

**Screenshots**

**Additional context**

e0:
This PR closes #1143.

**Type of change**  

**List of changes made**  

Use search terms that return results that do not contain any duplicate scores. Reason: the order for results with the same score is not important. If this reasoning is not sound and we want to actually ensure that the order is always consistent even for the same results, please let me know. We would have to change some implementation then.

**Testing**  

- Instructions on how to test
1. Run the tests for `GotEnzymes` : `ma-exec api yarn test test/gotEnzymes.test.js`.
2. Repeat this several times and verify that the tests always pass.

**Further comments**  

### DIFF
--- a/api/test/gotEnzymes.test.js
+++ b/api/test/gotEnzymes.test.js
@@ -2,19 +2,22 @@ import fetch from 'node-fetch';
 
 describe('gotEnzymes', () => {
   describe('search', () => {
-    it.each(['r08948', 'zurr', 'bacteria', 'hsa', 'nad'])(
+    it.each(['aspmetasp', 'dioxat', 'glnasngln', 'vacuole', 'zurr'])(
       'should return same results matches for %p no matter case',
       async searchTerm => {
-        const res_lower = await fetch(
-          `${API_BASE}/gotenzymes/search/${searchTerm}`
-        );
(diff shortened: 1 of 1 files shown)


MalinAhlberg (approved):
Nice job and nice refactoring!

The only potential problem I can think of is that we do not test any reaction or compound ids (eg `R01080` vs `r01080` or `C00003` vs `c00003`), as these always seem to give a number of hits with equal score.

I approve anyway, since these [other tests cover these types of examples](https://github.com/MetabolicAtlas/MetabolicAtlas/blob/main/api/test/gotEnzymes.test.js#L429).

I tested running this ~20 times without getting any errors.

nanjiangshu (approved):

Issue #1221, opened 2022-11-21
Nodes on 3d maps lose their color after being clicked


MalinAhlberg:
**Describe the bug**

When a node on a 3d map is clicked, it goes red. When clicking another node, the first node goes white, instead of going back to its original color.

**Steps to reproduce**

- Visit https://metatlas-dev.csbi.chalmers.se/explore/Human-GEM/interaction-partners/MAM01369e?expandedIds=ENSG00000164363&dataTypes=gene&dataSources=hpaRna.tsv&dataSets=None
- Click a sequence of nodes, notice that all clicked nodes go white

**Expected behavior**

- [ ] Nodes get their original color back after being highlighted

**Additional context**

The production server does not have this problem, so it's introduced quite recently.

PR #1234, opened 2022-11-29
feat: add toggle states for gene and label buttons

### LINKED ISSUE #1107: Show difference when clicking the show/hide button in the Map Viewer

As a user, I want the `show/hide` button looks different after clicking so that I know clearly whether the current status is show or hide.

**Acceptance Criteria**

- [ ] The status of the show/hide button can be intuitively distinguished after clicking. 
- [ ] The web page documentation is updated accordingly

**Additional context**

e0:
This PR closes #1107.

**Type of change**  

**List of changes made**  

- Change button appearance (`fa-eye` or `fa-eye-slash`) depending on the toggle state.
- Change tooltip message (e.g. 'Show labels' or 'Hide labels') depending on the toggle state.

**Screenshot of the fix**  

image

**Testing**  

- Instructions on how to test

1. Visit an interaction partners page, for example: `/explore/Human-GEM/interaction-partners/ENSG00000158578`.
2. Verify that the label toggle button works and that the state (button appearance and tooltip/hover message) changes after clicking on it.
3. Visit a 2d map viewer page, for example: `/explore/Human-GEM/map-viewer/acylglycerides_metabolism?dim=2d`.
4. Verify that the gene and subsystem toggle buttons work and that the states (button appearance and tooltip/hover message) change after clicking on it.
5. Visit a 3d map viewer page, for example: `/explore/Human-GEM/map-viewer/acylglycerides_metabolism?dim=3d`.
6. Verify that the gene and label toggle buttons work and that the states (button appearance and tooltip/hover message) change after clicking on it.

**Further comments**  

In the 3d map viewer page, the node colors get set to white after toggling the genes button. This bug is not introduced with this PR. It may be related to #1221 and if not a new issue should be created for it.

### DIFF
--- a/frontend/src/components/explorer/mapViewer/MapControls.vue
+++ b/frontend/src/components/explorer/mapViewer/MapControls.vue
@@ -23,10 +23,10 @@
     <span
       v-if="toggleSubsystems"
       class="button p-2"
-      title="Show/Hide subsystem"
-      @click="toggleSubsystems()"
+      :title="`${showLabels ? 'Hide' : 'Show'} subsystems`"
+      @click="handleToggleSubsystems()"
     >
-      <i class="fa fa-eye-slash">&thinsp;S</i>
+      <i class="fa" :class="[showSubsystems ? 'fa-eye-slash' : 'fa-eye']">&thinsp;S</i>
     </span>
     <span
       v-if="toggleBackgroundColor"
@@ -99,6 +99,7 @@ export default {
     return {
       showGenes: true,
       showLabels: true,
+      showSubsystems: true,
     };
   },
   computed: {
@@ -122,6 +123,10 @@ export default {
     handleToggleLabels() {
       this.showLabels = !this.showLabels;
       this.toggleLabels();
+    },
+    handleToggleSubsystems() {
+      this.showSubsystems = !this.showSubsystems;
+      this.toggleSubsystems();
     },
     handleToggleFullScreen() {
       if (this.isFullscreenDisabled) {
--- a/frontend/src/components/explorer/mapViewer/MapControls.vue
+++ b/frontend/src/components/explorer/mapViewer/MapControls.vue
@@ -4,11 +4,21 @@
     <span class="button" title="Zoom out" @click="zoomOut()">
       <i class="fa fa-search-minus"></i>
     </span>
-    <span v-if="toggleGenes" class="button p-2" title="Show/Hide genes" @click="toggleGenes()">
-      <i class="fa fa-eye-slash">&thinsp;G</i>
+    <span
+      v-if="toggleGenes"
+      class="button p-2"
+      :title="`${showGenes ? 'Hide' : 'Show'} genes`"
+      @click="handleToggleGenes()"
+    >
+      <i class="fa" :class="[showGenes ? 'fa-eye-slash' : 'fa-eye']">&thinsp;G</i>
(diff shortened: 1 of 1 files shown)


inghylt:
Nice work!

> I get a bit confused by the symbols, however. To me, it would make sense to show the open eye when eg gene labels are visible, and that the eye would be closed if they are hidden. But I totally understand the reasoning behind having them as is, just wanted to check if this is the standard? I googled a bit, but couldn't find any good examples on other pages.

I agree with this, but not something I feel strongly about.

I think I found another bug though: The background color button change takes only takes affect after you interact with the screen in some way, for example by clicking or zooming. Should I report this as a new bug or should we fix it in this PR?

e0:
> I get a bit confused by the symbols, however. To me, it would make sense to show the open eye when eg gene labels are visible, and that the eye would be closed if they are hidden. But I totally understand the reasoning behind having them as is, just wanted to check if this is the standard? I googled a bit, but couldn't find any good examples on other pages.

Really good observation! I was thinking about this as well during the implementation. It's a dilemma with these toggle buttons and I can see how it would make sense for the icon to represent its current state. The other way, and how we have done, is to have the icon to represent the purpose of clicking on it. You can find some interesting discussion about this here: https://ux.stackexchange.com/questions/1318/should-a-toggle-button-show-its-current-state-or-the-state-to-which-it-will-chan 

For our case, we have the fullscreen toggle button in the map viewer and it also represents the action as opposed to the current state. So I think the way we have it here is at least consistent with that button.

e0:
> I think I found another bug though: The background color button change takes only takes affect after you interact with the screen in some way, for example by clicking or zooming. Should I report this as a new bug or should we fix it in this PR?

That's a good catch! Usually I think we should have a new issue for this, but it should be a one-liner and I think I can just try to fix it here if you are ok with that. I don't think we need to have different button states for this one though. What do you think?

inghylt:
@e0 I agree with you both regarding trying to fix it in this PR and that it is fine to not have different button states!

MalinAhlberg:
> > I get a bit confused by the symbols, however. To me, it would make sense to show the open eye when eg gene labels are visible, and that the eye would be closed if they are hidden. But I totally understand the reasoning behind having them as is, just wanted to check if this is the standard? I googled a bit, but couldn't find any good examples on other pages.
> 
> Really good observation! I was thinking about this as well during the implementation. It's a dilemma with these toggle buttons and I can see how it would make sense for the icon to represent its current state. The other way, and how we have done, is to have the icon to represent the purpose of clicking on it. You can find some interesting discussion about this here: https://ux.stackexchange.com/questions/1318/should-a-toggle-button-show-its-current-state-or-the-state-to-which-it-will-chan
> 
> For our case, we have the fullscreen toggle button in the map viewer and it also represents the action as opposed to the current state. So I think the way we have it here is at least consistent with that button.

Good points, and thanks for the link! I agree it can/should be kept as is then.

MalinAhlberg (approved):

nanjiangshu (approved):
@e0 Very elegant implementation. I agree that it's a bit weird the show/hide icon does not represent the current status but the status it will bring on click. However, I think it is a standard usage of the show/hide icon and also inconsistent with the rest of icons, e.g. zoom in and zoom out.

inghylt (approved):
10/10 button work :star:

PR #1235, opened 2022-11-29
fix: reapply colors after node has been clicked

### LINKED ISSUE #1221: Nodes on 3d maps lose their color after being clicked

inghylt:
This PR closes #1221

**Type of change**  

**List of changes made**  

Colors are reapplied after a node has been clicked

**Screenshot of the fix**  

**Testing**  

- Instructions on how to test
Please visit both 3D network viewer and IP partner. Try out networks of different sizes and with multiple simultaneous overlays.

**Further comments**  

There is some lag, but IMO (and @e0:s :D) it is not a problem. The lag only annoys me a bit when applying reaction overlay on  Cytosol and Extracellular, i,e, massive networks, but since those are really too big to be of any value for researchers anyway I think it is acceptable. 

Note that there is no lag even on large networks such as `/explore/Human-GEM/interaction-partners/MAM03343c?expandedIds=MAM02039c`

### DIFF
--- a/frontend/src/components/explorer/interactionPartners/DetailsPage.vue
+++ b/frontend/src/components/explorer/interactionPartners/DetailsPage.vue
@@ -597,6 +597,7 @@ export default {
       const [id, type] = this.getElementIdAndType(element);
       this.clickedElmId = id;
       this.clickedElm = { id, type, n: element.n };
+      this.applyColors();
     },
     getElementIdAndType(element) {
       let type = 'metabolite';
--- a/frontend/src/components/explorer/interactionPartners/DetailsPage.vue
+++ b/frontend/src/components/explorer/interactionPartners/DetailsPage.vue
@@ -536,10 +536,13 @@ export default {
       // map appears "flat".
       this.controller.setCamera({ x: 0, y: 0, z: lz });
     },
-    async applyColors() {
+    async applyColors({ deselect } = { deselect: false }) {
       if (this.controller) {
         const colors = {};
-        this.network.nodes.forEach(node => {
+        const nodes = deselect
+          ? this.network.nodes.filter(n => n.id === this.clickedElmId)
+          : this.network.nodes;
+        nodes.forEach(node => {
           let color = colorToRGBArray(this.defaultMetaboliteColor);
 
           // TODO: use this when implementing compartment and subsystem highlight
@@ -571,7 +574,7 @@ export default {
               color = colorToRGBArray(this.defaultMetaboliteColor);
             }
           }
-          if (node.id === this.clickedElmId) {
+          if (node.id === this.clickedElmId && !deselect) {
             colors[node.id] = colorToRGBArray(NODE_SELECT_COLOR);
           } else {
             colors[node.id] = color;
@@ -594,10 +597,10 @@ export default {
       this.$store.dispatch('interactionPartners/setCoords', payload);
     },
     async selectElement(element) {
+      await this.applyColors({ deselect: true });
       const [id, type] = this.getElementIdAndType(element);
       this.clickedElmId = id;
       this.clickedElm = { id, type, n: element.n };
-      this.applyColors();
     },
     getElementIdAndType(element) {
       let type = 'metabolite';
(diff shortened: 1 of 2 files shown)


e0 (commented):
Nice fix and looks really good!

On my machine there was no significant lag even for the larger networks.

Could we also address the problem (the node colors get set to white after toggling the genes button) mentioned in the **Further comments** section in [this PR](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1234#issue-1467839343)? I used your same solution and added `this.applyColors();` to the `toggleGenes` method and it seems to solve it.

inghylt:
> Could we also address the problem (the node colors get set to white after toggling the genes button) mentioned in the Further comments section in https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1234#issue-1467839343? I used your same solution and added this.applyColors(); to the toggleGenes method and it seems to solve it.


Nice suggestion, fixed in 26b0f210e4b8cada7b05764b5e9ee0c0a3581d0f

e0:
> I am a bit concerned about the lag, though. On my machine, using cytosol with applied overlay, it takes almost 10s for the selection color to appear if I click two nodes sequentially (which often happens if you for example want to move the network). It takes a few more seconds for the information card in the left sidebar to load. And trying to turn the network is not really doable. If this is only on my machine, I think it's fine, but maybe we should do some more testing before merging this?

When we were discussing this before I was recalling that researchers have mentioned that they usually do not interact too much with these larger networks. How does this perform for you on the smaller ones though? I'm not sure how much of the performance is impacted by running this locally and we can also try deploying this to the dev server to make sure.

MalinAhlberg:
> When we were discussing this before I was recalling that researchers have mentioned that they usually do not interact too much with these larger networks. How does this perform for you on the smaller ones though? I'm not sure how much of the performance is impacted by running this locally and we can also try deploying this to the dev server to make sure.

On small networks, there is no noticeable lag. On medium sized, the lag is ok for clicking (0-2s). I think the biggest problem is if you have a network that is dense (such as mitochondria, golgi, transport reactions or extracellular), so that when trying to turn/move it, you will accidentally click nodes and everything will freeze for some seconds.
I think it makes sense to assume that researchers don't use the larger networks much, and a bit of lag is acceptable there. I'm still a bit concern though, I guess one scenario could be that you color a (small) 3d network, then change to a bigger one (applied overlay stays) and try to just have a look (which usually involves clicking a bit) and everything hangs.
But yes, it would be nice to see how this behaves on the dev server. I won't have time to deploy/test today, but can do that tomorrow.

nanjiangshu (approved):
Cool that the problem can be fixed with just few lines of code. The lagging is noticeable when the network is large or when there are multiple data sets overplayed, but I think it's not a big problem since it affects only a few very large networks like Cytosol and even before the fix it is also a bit awkward to work with. 

There is a small problem I noticed is that when a node is selected, the color of the links connected to the selected node  becomes yellow which in many cases is almost invisible, give the user a false feeling that the links are gone. I think we can probably choose a better color for the links connected to the node. Since it's an old problem, maybe we should create a new issue for that. I approve this PR.

MalinAhlberg (approved):
Works really well with the last fix :star:

frontend/src/components/explorer/interactionPartners/DetailsPage.vue
```diff
@@ -536,10 +536,13 @@ export default {
       // map appears "flat".
       this.controller.setCamera({ x: 0, y: 0, z: lz });
     },
-    async applyColors() {
+    async applyColors({ deselect } = { deselect: false }) {
       if (this.controller) {
         const colors = {};
-        this.network.nodes.forEach(node => {
+        const nodes = deselect
+          ? this.network.nodes.filter(n => n.id === this.clickedElmId)
```
inghylt, line 543: Ah this was a much cleaner way of obtaining the required node data compared to what I had in mind, nice!

e0 (approved):

PR #1236, opened 2022-11-29
fix: interaction partners table search

### LINKED ISSUE #1167: Weird filtering behavior in IP reactions table
**Describe the bug**

The filter/search in the reactions table on the IP page behaves inconsistently, both in the old and the new implementation.

**Steps to reproduce**

- Go to eg `/explore/Human-GEM/interaction-partners/MAM02634c`
- Note that there are multiple rows with reactions starting with `MAR0`
- Search for `MAR0`. Only a few of the rows  are displayed.

**Expected behavior**

- [ ] Searching for a substring of a reaction name works

**Screenshots**

**Additional context**

e0:
This PR closes #1167.

**Type of change**  

**List of changes made**  

Fix two bugs that prevented the search from completing.

**Screenshot of the fix**  

image

**Testing**  

- Instructions on how to test
1. Visit `/explore/Human-GEM/interaction-partners/MAM02634c`.
2. Try searches that should match and filter certain results, for example `MAR0` (as mentioned in the bug description), `l-`, `sLC`, `25` etc.
3. Verify that the table updates accordingly (graying out the unmatched ones and placing them below the matched ones).

**Further comments**  

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -14,7 +14,7 @@
     "format": "prettier --write src"
   },
   "dependencies": {
-    "@metabolicatlas/3d-network-viewer": "^0.1.31",
+    "@metabolicatlas/3d-network-viewer": "^0.1.32",
     "@panzoom/panzoom": "^4.4.3",
     "@vueuse/head": "^0.7.9",
     "axios": "^0.24.0",
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -14,7 +14,7 @@
     "format": "prettier --write src"
   },
   "dependencies": {
-    "@metabolicatlas/3d-network-viewer": "^0.1.32",
+    "@metabolicatlas/3d-network-viewer": "^0.1.33",
     "@panzoom/panzoom": "^4.4.3",
     "@vueuse/head": "^0.7.9",
     "axios": "^0.24.0",
--- a/frontend/src/components/explorer/interactionPartners/ReactionTable.vue
+++ b/frontend/src/components/explorer/interactionPartners/ReactionTable.vue
@@ -189,8 +189,8 @@ export default {
       if (this.tableSearch === '') {
         this.matchingReactions = Array.prototype.slice.call(this.sortedReactions);
         this.unmatchingReactions = [];
-        if (this.$refs.matchingTableBody.style) {
-          this.$refs.matchingTableBody.style.display = '';
+        if (this.$refs.matchingTableBody[0].style) {
+          this.$refs.matchingTableBody[0].style.display = '';
         }
       } else {
         this.matchingReactions = [];
@@ -207,7 +207,7 @@ export default {
                   if (k === 'id') {
                     match = el[k].toLowerCase() === t;
                   } else {
-                    match = el[k].toLowerCase().includes(t);
+                    match = typeof el[k] === 'string' && el[k].toLowerCase().includes(t);
                   }
                   if (match) {
                     matches = match;
@@ -230,9 +230,9 @@ export default {
 
         // fix disappearing row/cell borders
         if (this.matchingReactions.length === 0) {
-          this.$refs.matchingTableBody.style.display = 'none';
+          this.$refs.matchingTableBody[0].style.display = 'none';
         } else {
-          this.$refs.matchingTableBody.style.display = '';
+          this.$refs.matchingTableBody[0].style.display = '';
         }
       }
     },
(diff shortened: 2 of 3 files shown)


inghylt:
Really nice fix!

Another example:

`/explore/Human-GEM/interaction-partners/ENSG00000168970` and search for `MAR07`, `Lysosome` etc.

e0:
> But why does searching for `ump` not hide any rows in this table `/explore/Human-GEM/interaction-partners/ENSG00000076685?expandedIds=&dataTypes=gene&dataSources=hpaRna.tsv&dataSets=None`?

Good question! It seems that we have are searching through all of the properties for all of the metabolites and genes and not just the displayed names. With the following row for example, the gene `NT5C` has `cdN;DNT-1;DNT1;PN-I;UMPH2` for its synonyms, which includes the characters `UMP`.

image

I do think we could do a better job at presenting this. For example highlighting the matched elements. I think this belongs in a new issue though if we want to do it.

nanjiangshu (approved):
Nice fix @e0! The feature to grey-out the unmatched records looks great! 
There is room for improvement such as the highlighting the keyword as you have already mentioned, or show the number of matched records, but those should be addressed in a new issue. I approve this PR.

nanjiangshu:
Sorry for my early approval. I found a problem with the table search that the network zoomed out after typing `mar` in the searching field, as shown in the screen recording.


https://user-images.githubusercontent.com/5833722/204576893-8ef934ea-709f-4ac3-8d66-ffe7df4b0ee0.mov

This seems to be an old problem since it exist on the production server as well https://metabolicatlas.org/explore/Human-GEM/interaction-partners/ENSG00000168970?expandedIds=&dataTypes=gene&dataSources=hpaRna.tsv&dataSets=None

I'm not sure if you want to fix this problem in this issue or create a new issue for that.

MalinAhlberg:
> > But why does searching for `ump` not hide any rows in this table `/explore/Human-GEM/interaction-partners/ENSG00000076685?expandedIds=&dataTypes=gene&dataSources=hpaRna.tsv&dataSets=None`?
> 
> Good question! It seems that we have are searching through all of the properties for all of the metabolites and genes and not just the displayed names. With the following row for example, the gene `NT5C` has `cdN;DNT-1;DNT1;PN-I;UMPH2` for its synonyms, which includes the characters `UMP`.
> 
> image
> 
> I do think we could do a better job at presenting this. For example highlighting the matched elements. I think this belongs in a new issue though if we want to do it.

Thanks for the explanation! I created #1237  for this.

MalinAhlberg:
> Sorry for my early approval. I found a problem with the table search that the network zoomed out after typing `mar` in the searching field, as shown in the screen recording.
> 
>  IP-table-search-zoomout-network.mov 
> This seems to be an old problem since it exist on the production server as well https://metabolicatlas.org/explore/Human-GEM/interaction-partners/ENSG00000168970?expandedIds=&dataTypes=gene&dataSources=hpaRna.tsv&dataSets=None
> 
> I'm not sure if you want to fix this problem in this issue or create a new issue for that.

Very interesting :D. It's triggered when just typing `r` in the search field. For me, it would make sense to create a new issue for this, so I will approve this PR (but I'm not against fixing it as part of this PR).

MalinAhlberg (approved):

e0:
Thanks for reporting this issue @nanjiangshu and confirming it @MalinAhlberg!

In the network viewer we had some keyboard shortcuts enabled by default. Any time a user presses `r`, `q`, `l`, or `g`, it would perform an action. `r` is for `resetCamera`. This also affects the 3d map viewer when typing in the map search input.

This interferes so much with the search so I thought we would fix it in this PR. I have updated the network viewer to disable keyboard shortcuts by default. The keyboard shortcuts are really problematic, for example pressing `g` would toggle the `genes` even though we haven't allowed for this in the interaction partners network toolbar. I don't think we have officially documented the usage of these so I think it's safe to make this change. One can still enable keyboard shortcuts but that would be an opt-in. Please let me know if this should be reconsidered though.

To ensure the package gets properly updated, please rebuild then `frontend` before testing this again: `docker-compose --env-file env-local.env -f docker-compose.yml -f docker-compose-local.yml up -d --no-deps --build frontend`.

nanjiangshu (approved):
Great that you found the problem and solved it nicely. I think it makes sense to disable these shortcut keys and since we have never documented the usage of these shortcuts for the 3D network viewer, I don't think there should be any problem to disable it.

MalinAhlberg (approved):
Good fix, @e0! I also think it makes to disable the shortcuts.

Also good catch from @inghylt. I approve anyway, since I don't think it needs to be fixed in this PR.

e0:
Thank you for the careful testing @inghylt!

That was a really annoying bug and should have been included in the previous version in network viewer. I have updated this again and think it works as it should now.

inghylt (approved):
Great work :clap:

PR #1238, opened 2022-11-30
fix: clear coords when switching dim

### LINKED ISSUE #971: inconsistent behaviour when moving between 2d and 3d
**Describe the bug**
Switching between 2D and 3D is not working as expected. The behavior depends on which state you're in when you enter the page / reload the page. 

**Steps to reproduce**

Example:
- Go to the MapViewer and select Cystol, part 1
- Change to 3D
- Reload the page
- Change back to 2D
You will  see an empty map.

**Expected behavior**

- [ ] Repeating steps above will result in that the correct map is shown

**Screenshots**

**Additional context**

This is related to #837.
The problem seems to be that [`this.$route.query.dim`](https://github.com/MetabolicAtlas/MetabolicAtlas/blob/develop/frontend/src/components/explorer/MapViewer.vue#L281) is not updated when switching to another dimension, it will always keep the value from when the page is first loaded (or reloaded).

inghylt:
This PR closes #971 and was written by @e0 

**Type of change**  

**List of changes made**  

- Coordinates from the url are not kept when switching between 2d and 3d 

**Screenshot of the fix**  

**Testing**  

- Instructions on how to test

Test  #971 first reported problem
1. Go to the MapViewer and select Cystol, part 1
2. Change to 3D
3. Reload the page
4. Change back to 2D
5. Check the map looks good

Test  #971 second reported problem

1. Go to the map viewer and select Acyl-CoA Hydrolysis
2. Change to 3D
3. Reload the page
4. Change to map Alkaloids biosynthesis
5. Change to 2D
6. The map is shown really quick and then disappears again

Please also play around with the 2d and 3d functionality +  copy pasting the url to make sure everything is working as intended

**Further comments**  

### DIFF
--- a/frontend/src/components/explorer/MapViewer.vue
+++ b/frontend/src/components/explorer/MapViewer.vue
@@ -268,12 +268,12 @@ export default {
         return;
       }
 
-      const queryString = Object.entries(newQuery)
-        .map(e => e.join('='))
-        .join('&');
-
       if (newQuery.dim === this.$route.query.dim || (newQuery.dim && !this.$route.query.dim)) {
         // if Map viewer (2D or 3D) is not changed, keep the url path
+        const queryString = Object.entries(newQuery)
+          .map(e => e.join('='))
+          .join('&');
+
         const url = `${this.$route.path}?${queryString}`;
         history.replaceState(history.state, '', url); // eslint-disable-line no-restricted-globals
       } else {
@@ -290,6 +290,11 @@ export default {
             urlPath = urlPath.replace(new RegExp(this.$route.params.map_id), this.currentMap.id);
           }
         }
+        const queryString = Object.entries(newQuery)
+          .filter(([key]) => key !== 'coords')
+          .map(e => e.join('='))
+          .join('&');
+
         const url = `${urlPath}?${queryString}`;
         await this.$router.push(url);
         history.replaceState(history.state, ''); // eslint-disable-line no-restricted-globals


nanjiangshu (commented):
Nice fix! I tried with the example provided and also played with some random examples by copy/paste urls and open in different browsers, all seem to work very well. 

There is a problem that I found when playing with Cytosol map when data overlay is applied. 
As shown in the video below, when the 3D map is opening, there is a spinning circle. However, when the map is loaded, the spinning circle disappeared but the map is still now showing up.
If you try to move the move a few times, the map may disappear after appearing. I think it would be nice to add a spinning circle even at this stage. 

But I think we can address this problem in a new issue. 
https://user-images.githubusercontent.com/5833722/204822622-ee0630f5-5918-4776-966d-3b9e8e8a1458.mov

inghylt:
@nanjiangshu I've tried to recreate the problem. On my localhost, when moving between 2D to 3D with gene and reaction overlay applied it takes a long time to load and the canvas is empty for another 1-2 sek, but after that I have no problem. Same behavior on prod only much faster response time. If I haven't missed anything when trying to recreate the problem, I don't think it's something we need to fix, but feel free to create an issue if you see fit!

nanjiangshu:
Hi @inghylt, I did some more tests and find that the problem is caused by the recoloring of the nodes, for the current main branch, commit a63814da, the canvas was empty for +10 seconds on my localhost but hardly noticeable for the commit tags/3.1 (<0.5s). Interestingly the network will disappear again after first appearing and after a few seconds it will reappear and stay. I guess this might be caused by double calling of `this.applyColors();`.

I've created the issue #1239 to address this problem.

nanjiangshu (approved):

MalinAhlberg:
Nice fix :+1: !

I can't produce the first bug (locally on `main`, on `dev` or `prod`) anymore, but I can verify that the second bug is fixed by this PR.
Another thing: when visiting `explore/Human-GEM/map-viewer/alkaloids_biosynthesis?dim=3d&panel=0&sel=&search=&dataTypes=None&dataSources=None&dataSets=None` and then changing to 2d, the map blinks a bit (quickly switches between very zoomed in and standard zoom). This happens also on `prod`, so is not introduces in this PR, but I wonder if it might be related enough to be worth fixing?
If not, I will approve this PR :)

inghylt:
@MalinAhlberg Yeah I've noticed that blinking to. IMO we don't need to spend time on that since the map viewer still functions as expected and I don't think users will notice/will be bothered by a minor blink

MalinAhlberg:
> @MalinAhlberg Yeah I've noticed that blinking to. IMO we don't need to spend time on that since the map viewer still functions as expected and I don't think users will notice/will be bothered by a minor blink

I definitely think the will notice it, at least for me it's really eye-catching. But I will approve of this PR and create a new issue for it :)

MalinAhlberg (approved):

PR #1240, opened 2022-12-02
fix: correct error message for bad value for expanded ids

### LINKED ISSUE #1174: Providing bad value for the url param expandedIds gives incorrect error message
**Describe the bug**

Providing a bad value for the url param `expandedIds` gives the user an error saying that the provided mainId does not exist, even if it does.

**Steps to reproduce**

- visit `/explore/Human-GEM/interaction-partners/ENSG00000093072`. Note that it works.
- visit `/explore/Human-GEM/interaction-partners/ENSG00000093072?expandedIds=badvalue`. 

**Expected behavior**

- [ ] the error shows the correct reason, no network is shown (give invalid ids, if possible all of them otherwise the first one)

**Screenshots**

 
**Additional context**

Go for the easiest solution

MalinAhlberg:
This PR closes  #1174.

The api will return the failing node id when calling `/interaction-partners` and `interaction-partners-expansion`.
The frontend will show the failing node id.

**Type of change**  

**List of changes made**  

Api will return an error message `Invalid id xxx` when provided with an incorrect value for interaction partners
Frontend will parse and show that id

**Screenshot of the fix**  

2022-12-02-084350_1435x295_scrot

**Testing**  

- Visit Interaction Partners, eg `explore/Human-GEM/interaction-partners/MAM03769e`, and make sure you can expand nodes as expected
- Visit Interaction Partners using a bad value for the main node, eg `explore/Human-GEM/interaction-partners/nonono` and make sure that the error message makes sense
- Try adding a bad value for the expanded node, make sure the error message makes sense, eg
  - `/explore/Human-GEM/interaction-partners/MAM03769e?expandedIds=crash`
  - `/explore/Human-GEM/interaction-partners/MAM03769e?expandedIds=MAM03768e,crash`
- Make a call to the api and make sure the output looks ok, eg `/api/v2/interaction-partners-expansion/MAM03769e?model=HumanGem&version=1_12_0&expanded[]=MAM03768e&expanded[]=crash`

**Further comments**  

When providing multiple erroneous values for the expanded ids, one of them will be shown. Which one will depend on which query from the [`Promise.all`](https://github.com/MetabolicAtlas/MetabolicAtlas/blob/ac94d0293197f67ade534ee7fdc2cc45071ebed4/api/src/neo4j/queries/interactionPartners.js#L135) that will fail first.
If the main node id is invalid, that will always be reported as the failing id.

### DIFF
--- a/api/src/endpoints/neo4j.js
+++ b/api/src/endpoints/neo4j.js
@@ -62,6 +62,9 @@ const fetchWith = async (req, res, queryHandler) => {
     const result = await queryHandler(payload);
     res.json(result);
   } catch (e) {
+    if (e.message.startsWith('Invalid id')) {
+      return res.status(404).send(e.message);
+    }
     if (e.message === '404') {
       return res.sendStatus(404);
     }
--- a/api/src/neo4j/queries/interactionPartners.js
+++ b/api/src/neo4j/queries/interactionPartners.js
@@ -50,7 +50,14 @@ WITH apoc.map.mergeList(apoc.coll.flatten(
 )) as reaction, component
 RETURN { component: component, reactions: COLLECT(reaction)}
 `;
-  const result = await querySingleResult(statement);
+  let result;
+  try {
+    result = await querySingleResult(statement);
+  } catch (e) {
+    const error = new Error(`Invalid id: ${id}`);
+    throw error;
+  }
+
   let links = [];
   let nodes = [];
   let unique = new Set();
--- a/api/src/neo4j/queries/interactionPartners.js
+++ b/api/src/neo4j/queries/interactionPartners.js
@@ -133,13 +133,15 @@ const getInteractionPartnersExpansion = async ({
   let expandedNodes = {};
 
   const expandedNetworks = await Promise.all(
-    expanded.map(nodeId =>
-      getInteractionPartners({
-        id: nodeId,
-        model,
-        version,
-      })
-    )
+    expanded
+      .filter(n => n.trim())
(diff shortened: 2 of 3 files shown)


nanjiangshu (commented):
Great fix @MalinAhlberg! It works well for most of the scenarios I can come up with. 
However, there is a small problem when the field `expandedIds` contains only the letter `[ ,:]`, then the error message will show as 
Screenshot 2022-12-02 at 10 04 11
Maybe we can consider to replace the string as empty if the string contains only these letters. Another thought is that the users probably do not expect a notFound error message when the field `expandedIds` accidentally contains only a white space. 
More specifically, I think
1. if `expandedIds` contains only white spaces, it should be replaced with an empty string so that no error message will be shown (if the mainID has a hit)
2. If `expandedIds` contains only letters `[,:]`, show error message as 
    show letters `,` and `:` instead of blank field.
But I'm also OK to ignore the second point, since I do think the use case for such condition is extremely rare.

inghylt (approved):
Really nice fix!

I think the scenarios with `[ ,:]` reported in this comment https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1240#pullrequestreview-1202190729 are so unlikely that we don't need to spend time on fixing them, but up to you @MalinAhlberg

MalinAhlberg:
Thanks for the reviews @nanjiangshu and @inghylt! Since the errors @nanjiangshu found were easy to fix, I did so, see 5adc21b.

PR #1244, opened 2022-12-05
fix: consistent about links

### LINKED ISSUE #1068: Inconsistent behavior when clicking links to the about pages on mobile devices
**Describe the bug**

Clicking links to some sections of the About pages from the Table of Content does does not scroll to the right place.

**Steps to reproduce**

- Click eg "Contact us" from the "Table of Content" on a mobile device (scrolls to the "Contact us" section)
- Compare to clicking "News" (stays on the menu)

**Expected behavior**

- [ ] Clicking a link from the TOC will consistently either scroll to the correct place or stay on the menu
- [ ] Clicking a link from the Nav Bar will  consistently either scroll to the correct place or stay on the menu
- [ ] Links at the "Documentation" pages still work as expected
- [ ] Tested on different browsers and devices

**Additional context**

See `frontend/src/content/about.js` The TOC uses links rather than router names, to allow scrolling to anchors on the page. The Nav Bar uses the router names.

See also discussions at https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1064.

inghylt:
This PR closes #1068.

**Type of change**  

**List of changes made**  

Scroll to the section title when there are no subsections. Fix added for both News and ELIXIR section since the bug was present for both of them

**Screenshot of the fix**  

**Testing**  

Instructions on how to test
- Click eg "Contact us" from the "Table of Content"
- Compare to clicking "News" 
- Click eg "Contact us" from the "Table of Content"
- Compare to clicking "ELIXIR" 
 
Do this on normal and a mobile device

**Further comments**  

I have noticed when visiting Elixir for the first time the scrolling does not always end up right at the title. For News that does not occur, so I think the problem might have something to do with the logos. The `setTimeout` was added to fix it, works well in Chrome, but in Firefox I can still recreate the problem. Best way to recreate it is to visit `http://localhost/about/elixir` until you notice that the title is not at the top. But since the original problems are fixed and ELIXIR is still functional, I think this is acceptable. Do you guys agree?

### DIFF
--- a/frontend/src/components/about/Elixir.vue
+++ b/frontend/src/components/about/Elixir.vue
@@ -79,6 +79,9 @@ export default {
       ],
     };
   },
+  mounted() {
+    setTimeout(() => document.getElementById('title-ELIXIR').scrollIntoView(), 50);
+  },
 };
 </script>
 
--- a/frontend/src/components/about/Elixir.vue
+++ b/frontend/src/components/about/Elixir.vue
@@ -79,6 +79,7 @@ export default {
       ],
     };
   },
+  // Implemented since this section does not have a subsection anchor to use in MetabolicAtlas/frontend/src/content/about.js
   mounted() {
     setTimeout(() => document.getElementById('title-ELIXIR').scrollIntoView(), 50);
   },
--- a/frontend/src/components/about/News.vue
+++ b/frontend/src/components/about/News.vue
@@ -47,5 +47,8 @@ export default {
       news,
     };
   },
+  mounted() {
+    document.getElementById('title-News').scrollIntoView();
+  },
 };
 </script>
--- a/frontend/src/components/about/News.vue
+++ b/frontend/src/components/about/News.vue
@@ -47,6 +47,7 @@ export default {
       news,
     };
   },
+  // Implemented since this section does not have a subsection anchor to use in MetabolicAtlas/frontend/src/content/about.js
   mounted() {
     document.getElementById('title-News').scrollIntoView();
   },
--- a/frontend/src/layouts/AboutLayout.vue
+++ b/frontend/src/layouts/AboutLayout.vue
@@ -5,7 +5,7 @@
       <div class="columns is-variable is-8 pt-5">
         <TableOfContents :links="tocLinks" />
         <div class="column content has-text-justified">
-          <h3 v-if="title" class="title is-3 mt-0">{{ title }}</h3>
+          <h3 v-if="title" :id="'title-' + title" class="title is-3 mt-0">{{ title }}</h3>
           <div :class="{ 'content-container': !fullContentWidth }">
             <slot name="contents" />
           </div>


MalinAhlberg:
Nice fixes!!

I still find these inconsistencies, though:
- on chrome/firefox+mobile:
  - clicking links from the drop-down meny (from the top navbar): only news and sometimes elixir works, the rest do scroll down

-  firefox/chrome:
   - clicking links from the drop-down meny (from the top navbar): news / elixir will make the page scroll away from the menu, the rest won't

- chrome:
   - clicking links from the left side-bar: elixir scrolls further to the top then the rest of the links (About header is shown)

e0:
@inghylt This works well from my testing, I didn't have any problems using Firefox.

I do want to suggest an alternative solution that is more global though. You can revert the other changes to test this.

image

I haven't tested this too extensively with other pages. You can read more about this here: https://router.vuejs.org/guide/advanced/scroll-behavior.html#delaying-the-scroll

nanjiangshu:
I can reproduce all points that @MalinAhlberg mentioned, that is, while it solves some inconsistency on the mobile device, it brings other inconsistency on the desktop version, namely, the `News` and `ELIXIR` section. 
With the suggested changes from @e0, it solves the inconsistency for `News` on the desktop version, but the `ELIXIR` still the same, i.e. the title `ELIXIR` moves to the top without showing the title of `About`, on the desktop version.

inghylt:
Me and @e0 had a look at this together, please review the latest changes. We were able to find one bug which occur if you press the same link in the drop down meny multiple times, then the start position ends up a bit further down than expected. But we agreed on that this is not a common use case and is acceptable since it's still possible to read the text

MalinAhlberg (approved):
Great fix :v:!

I cannot find anymore bugs, everything seems to work well.

nanjiangshu:
Looks great for the mobile view when I tested all link jumping! However, the margin to the top are different for some sections, especially on the desktop view, e.g. `/about/platform#introduction` compared to the rest of the sections in the platform section. All pages in the `Related resources` section. For `/about/terms#privacy-policy` and `/about/terms#license`  the `About` title and banner are shown. I think we can probably ignore ``/about/terms#privacy-policy` and `/about/terms#license`, but for the `/about/platform#introduction` and `/about/resources#databases` and only need to change at two places
```diff
+++ b/frontend/src/components/about/Platform.vue
@@ -1,7 +1,7 @@
 <template>
   <about-layout id="Platform" title="">
     <template v-slot:contents>
-      <h3 id="introduction" class="title is-3 mt-0">Introduction</h3>
+      <h3 id="introduction" class="title is-3 pt-6">Introduction</h3>
```

```diff
+++ b/frontend/src/components/about/Resources.vue
@@ -6,7 +6,7 @@
           <!-- eslint-disable-next-line vue/require-v-for-key -->
           <div class="columns">
             <div class="column is-offset-2 has-text-centered-mobile">
-              <h4 :id="category.toLowerCase()" class="title is-4 pt-4">{{ category }}</h4>
+              <h4 :id="category.toLowerCase()" class="title is-3 pt-6">{{ category }}</h4>
```
Of course, it's up to you to change it or not.

inghylt:
@nanjiangshu Have a look at d379b12 and see if that is a good fix for what you raised! I tried the changes you suggested but that messed up the overall layout so hopefully this somewhat over engineered solution might fix things :)

nanjiangshu:
@inghylt sorry for my immature suggestions. With the commit  d379b12 there is still slight inconsistency for `Introduction` and added more weird behavior for `ELIXIR` and moreover, the code becomes more complicated to read. 
I guess it's better we revert this commit. Too bad that I wasted your time to work on this over engineered version.  
I can approve this PR at the previous commit.

nanjiangshu (approved):

e0:
I took a quick look at this. It's a good details @nanjiangshu spotted. For consistency and usability, I think it's best solved if the paddings can be changed to margins instead. Removing the padding makes sure that the section header is as near the top of the page as possible. The problem is that the headings already have a default margin so using the bulma helpers, changing from `pt-6` to `mt-6` for example, changes the vertical spacing between the sections. Personally I'm ok with this change but not sure how others feel.

@nanjiangshu For the Terms of use sections. I think this (title and banner are shown) is only a problem because the page is not long enough. If you shorten the height of the window it will look more like they are jumping to those sections.

@inghylt if you are interested please try this (`pt-` to `mt-`) out in Resources and Platforms. I'm fine with or without this change like the others.

MalinAhlberg:
I think everything looks very good!

nanjiangshu:
Looks great now @inghylt! Thanks for your patience for this PR 🙃

PR #1246, opened 2022-12-05
feat: implement parallelization for related reactions

### LINKED ISSUE #784: Explore if we can speed up neo4j queries through parallelization

inghylt:
This PR closes #784, written by @inghylt and @MalinAhlberg 

**Type of change**  

**List of changes made**  

- Make use of mapParallel2 to get faster response time

**Screenshot of the fix**  

**Testing**  

- Instructions on how to test

I have compared this branch and main on my localhost. The data returned from the db is equal, just a difference of order where `upperBound` is placed. Please try this branch out and compare the response time of different components in the gembrowser yourself. Try to find subsystems and genes with a lot of related reactions if possible

##### Metabolites
`/explore/Human-GEM/gem-browser/metabolite/MAM02040r`
`/explore/Human-GEM/gem-browser/metabolite/MAM02519c`
`/explore/Human-GEM/gem-browser/metabolite/MAM01371m`

These all have response time of ~40%, ~40% and ~90% of the original one, seems like the more related reaction the bigger the improvement

`http://localhost/explore/Human-GEM/gem-browser/metabolite/MAM02039c`
Seems like the limit is not enforced since all results 2735 are returned and not only 1000. However, the response time is a bit faster compared (~10%) to main where the limit of 1000 is enforced

##### Genes
`/explore/Human-GEM/gem-browser/gene/ENSG00000204371`
`/explore/Human-GEM/gem-browser/gene/ENSG00000136943`
Here the results varies, response time of ~220% for the first link and 95% for the second one

##### Subsystems
`/explore/Human-GEM/gem-browser/subsystem/beta_oxidation_of_odd_chain_fatty_acids_mitochondrial`
`/explore/Human-GEM/gem-browser/subsystem/alanine_aspartate_and_glutamate_metabolism`
Here the results varies, response time of ~120% for the first link and  90% for the second one

**Further comments**  

- The results is not 100% improvement for all components, but definitely a decrease for metabolites with a lot of related reactions. Should we keep mapParallel2 or not given these results?
- ~Should we enforce the limit or remove it completely?~ Created #1250
- ~Leaving the PR in draft mode and waiting with fixing cspell errors until the questions above are answered~ Fixed in aeba432


e0:
> * The results is not 100% improvement for all components, but definitely a decrease for metabolites with a lot of related reactions. Should we keep mapParallel2 or not given these results?

The tradeoff seems ok for me and I would vote for the `mapParallel2` implementation. The increased times are mostly for the smaller responses, so for me it's ok if most of these are going from `70ms` to `100ms` when we can get drastic decreases for the larger responses. Especially since we are caching the API requests.

> * Should we enforce the limit or remove it completely?

We should enforce the limit as we had before. To use limit together with `collect`, we should use this syntax: `WITH collect (r)[..1000] as reaction`.

inghylt:
@e0 Just curious, why should we enforce the limit as we had before?

e0:
> @e0 Just curious, why should we enforce the limit as we had before?

Oh I assumed it was a mistake/bug that the limit is not working in this PR. If we want to evaluate if the limit should be removed completely we should test this a bit more thoroughly. Especially on the frontend to see how the webpage performs when too many rows need to be rendered.

inghylt:
@e0 Haha it was not intentional but I thought the results were interesting since we get the full table in equal time as compared to with the limit.

I've tried `/explore/Human-GEM/gem-browser/metabolite/MAM02039g` which has 2735 rows, and if you press "See reactions from all compartment" it's 4571 rows. When you click on a link in one of the rows or in the gem browser table the response time is slower than usual, but other than that there is no difference that I can find. I think the slower response time is acceptable but have a try and see what you think

@nanjiangshu can you come up with a components with more related reactions that H+?

inghylt:
Thank you for the input @MalinAhlberg  @e0! I'm keeping the limit then, created #1250 for future removal

MalinAhlberg (approved):
Nice :star:

nanjiangshu:
> I've tried `/explore/Human-GEM/gem-browser/metabolite/MAM02039g` which has 2735 rows, and if you press "See reactions from all compartment" it's 4571 rows. When you click on a link in one of the rows or in the gem browser table the response time is slower than usual, but other than that there is no difference that I can find. I think the slower response time is acceptable but have a try and see what you think
> 
> @nanjiangshu can you come up with a components with more related reactions that H+?

I've checked all major currency metabolites, e.g. ATP, ADP, NADH, NAD+, H2O, H+, Pi, and H+ is the metabolites with most related reactions. There should be no other metabolites or genes have more related reactions than that.

nanjiangshu (approved):
Nice very work! I'm always happy to see such a big speed boost. 

During testing, I found another problem that the browser hangs at quick search when `restrict to current compartment` is false when searching for currency metabolites. 

Screenshot 2022-12-07 at 15 40 23
I guess one solution would be set `restrict to current compartment` to `false` when quick search is in focus. Shall we create a new issue for that?

e0:
> During testing, I found another problem that the browser hangs at quick search when `restrict to current compartment` is false when searching for currency metabolites.

Just to double check, this problem is not caused by the current PR right? If so please create a new bug issue for this and we can continue the discussion there. FWIW I tested it really quickly now and couldn't reproduce it.

nanjiangshu:
Here is a wired bug
For the metabolite H+,  http://localhost/explore/Human-GEM/gem-browser/metabolite/MAM02040c
It shows 1000 reactions for Cytosol 
but it shows 998 reactions for all compartments.

I think it is related to how limit is set, since when the limit is removed,
it shows 2015 reactions for Cytosol, and 3475 reactions for all compartments. 

The problem exists in the main branch, so I guess this bug can also be addressed in the issue #1250

nanjiangshu (approved):

PR #1249, opened 2022-12-06
feat: include installation of font

### LINKED ISSUE #774: improve loading of map thumbnail images

As a tech collaborator, I want to have load map thumbnails faster and with correct font rendering, so that I can show correct thumbnails without waiting for too long.

There are two problems:

1. For larger maps, the loading time can take a really long time (over one minute).
2. Fonts are not being rendered, it just shows rectangles.

A suggested solution is to pre-process all of the images from SVG into PNG format. Perhaps this is more suitable to be done in the [data-generation repo](https://github.com/MetabolicAtlas/data-generation). This should solve both of the problems, especially if the PNG files are resized into smaller ones (max 2000px wide for example), since it offloads the preprocessing time to the build-step and the text in PNG format would not cause any problems.

If we want to go with other solutions that takes each problem individually, it might make more sense to create two new issues to replace this one.

**Additional context**
[This comment](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/753#issuecomment-964079642), as well as discussion in the related issue, includes examples of the problems.

e0:
This PR closes #774.

**Type of change**  

**List of changes made**  

- Install a default font (`ttf-dejavu`) alongside `fontconfig`. (Problem before was that no fonts were installed.)
- Install other build tools to address `Pango-CRITICAL` errors.
- Update `sharp` version.

**Screenshot of the fix**  

image

**Testing**  

- Instructions on how to test
1. Rebuild the `api` container: `docker-compose --env-file env-local.env -f docker-compose.yml -f docker-compose-local.yml up -d --no-deps --build api`.
2. Visit a thumbnail endpoint, for example: `/api/v2/svg-thumbnail/mitochondria?model=Human-GEM&height=300` or `/api/v2/svg-thumbnail/acylglycerides_metabolism?model=Human-GEM&height=300`.
3. Verify that it does not take too long to load and that the text loads instead only showing rectangles.

**Further comments**  

We have [previously documented](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/753#issuecomment-964079642) that the thumbnail takes a really long time to generate. When I tested now, the same endpoint (`/api/v2/svg-thumbnail/mitochondria?model=Human-GEM&height=300`) takes under `2s` for me. So the speed seems to be no longer a problem. Please let me know if this still takes too long for you though.

### DIFF
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -1,5 +1,8 @@
 FROM node:16-alpine
-RUN apk add --no-cache fontconfig
+
+RUN apk add --no-cache --virtual .gyp python3 make g++
+RUN apk --no-cache --update add fontconfig ttf-dejavu
+
 WORKDIR /project
 COPY . .
 RUN yarn
--- a/api/package.json
+++ b/api/package.json
@@ -23,7 +23,7 @@
     "node-fetch": "2",
     "postgres": "^3.2.4",
     "sanitize-filename": "^1.6.3",
-    "sharp": "^0.30.5",
+    "sharp": "^0.31.2",
     "swagger-ui-express": "^4.3.0"
   },
   "devDependencies": {
--- a/api/yarn.lock
+++ b/api/yarn.lock
@@ -6079,16 +6079,16 @@ shallow-clone@^3.0.0:
   dependencies:
     kind-of "^6.0.2"
 
-sharp@^0.30.5:
-  version "0.30.5"
-  resolved "https://registry.yarnpkg.com/sharp/-/sharp-0.30.5.tgz#81c36fd05624978384ac6bd2744d23f9c82edefd"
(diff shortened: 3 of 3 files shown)


inghylt (approved):
Works well for me!

api/Dockerfile
```diff
@@ -1,5 +1,8 @@
 FROM node:16-alpine
-RUN apk add --no-cache fontconfig
+
+RUN apk add --no-cache --virtual .gyp python3 make g++
+RUN apk --no-cache --update add fontconfig ttf-dejavu
```
MalinAhlberg, line 4: just curious, why is `--virtual` and `--update` needed at these lines?

MalinAhlberg (approved):
Works well! And is very fast, good job!

api/Dockerfile
e0: Sorry but I didn't look enough into this. I sampled these from a few different sources for people who had problems with `pango` (a dependency for `fontconfig`) and missing fonts. I imagine we don't need either here but I saw a few posts similar to this. Hope it's just some best practices that I don't fully understand.

nanjiangshu (approved):
Awesome!

PR #1265, opened 2022-12-12
fix: blinking SVG map when switching between 2D and 3D

### LINKED ISSUE #1242: Map blinks when switching between 3d and 2d
**Describe the bug**

When switching between 2d and 3d maps in the map viewer, the zoom level of the map appears to change quickly a few times, making the map "blink".

**Steps to reproduce**

- visit `/explore/Human-GEM/map-viewer/alkaloids_biosynthesis?dim=3d&panel=0&sel=&search=&coords=-288.36,-660.92,0.5,0,0,500&dataTypes=None&dataSources=None&dataSets=None`
- switch between 2d and 3d a few times
- notice that the map "blinks" before settling 

**Expected behavior**

- [ ]  Map does not blink

**Additional context**

Discussed in https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1238#issuecomment-1334938529.

nanjiangshu:
This PR closes #1242 

**Type of change**  

**List of changes made**  

Fixed the problem that the SVG map blinks when switching back from the 3D map viewer 

**Screenshot of the fix**  

**Testing**  

- Instructions on how to test
Open the Map Viewer, switch between 2D and 3D and see if the SVG map blinks
Try a few examples

**Further comments**  

### DIFF
--- a/frontend/src/components/explorer/mapViewer/MapLoader.vue
+++ b/frontend/src/components/explorer/mapViewer/MapLoader.vue
@@ -1,6 +1,6 @@
 <template>
   <div v-show="loading" class="map-loader loading">
-    <a class="button is-loading"></a>
+    <a class="button is-warning is-loading"></a>
   </div>
 </template>
 
@@ -17,11 +17,11 @@ export default {
 };
 </script>
 
-<style lang="scss">
+<style lang="scss" scoped>
 .map-loader {
   z-index: 10;
   position: absolute;
-  background: black;
+  background: white;
   top: 0;
   left: 0;
   width: 100%;
@@ -34,7 +34,7 @@ export default {
     font-weight: 1000;
     display: table-cell;
     vertical-align: middle;
-    background: black;
+    background: white;
     border: 0;
   }
 }
--- a/frontend/src/components/explorer/mapViewer/Svgmap.vue
+++ b/frontend/src/components/explorer/mapViewer/Svgmap.vue
@@ -114,8 +114,10 @@ export default {
     }),
   },
   watch: {
-    async mapData() {
-      await this.init();
+    async mapData(newM, oldM) {
+      if (newM.id !== oldM.id) {
+        await this.init();
+      }
     },
     componentClassName() {
       this.setupHoverEventHandlers();


inghylt (approved):
Nice work!

MalinAhlberg:
Nice and clean fix!

For me, this makes the map blink a lot less :star:.
However, it still starts by showing a real zoomed-in version of the map, like the one below,  for a short while before getting the correct size
2022-12-12-140824_1381x891_scrot

I find this a bit annoying, but if no one else is bothered, I'm fine with ignoring it and approving this PR.

e0:
I also still notice a blink and I think it's from the loader showing up for a split second before displaying the map. I think making the map loader to have a light background and adding a small fade animation should help. If you want to test this out I have added a commit here: https://github.com/MetabolicAtlas/MetabolicAtlas/commit/4e5062298015186ead0ca7f1801e892a2d77f76c

What do you think @nanjiangshu @inghylt @MalinAhlberg ?

Please feel free to ignore/modify/cherry-pick it. If we are not adding it here I think we could add a new issue to address it after more discussion.

inghylt:
I actually notice what @MalinAhlberg describes in this https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1265#issuecomment-1346468118 more with the light background in this commit https://github.com/MetabolicAtlas/MetabolicAtlas/commit/4e5062298015186ead0ca7f1801e892a2d77f76c. I would prefer to just keep things as in this PR but I don't have a strong preference

e0:
> I actually notice what @MalinAhlberg describes in this [#1265 (comment)](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1265#issuecomment-1346468118) more with the light background in this commit [4e50622](https://github.com/MetabolicAtlas/MetabolicAtlas/commit/4e5062298015186ead0ca7f1801e892a2d77f76c). I would prefer to just keep things as in this PR but I don't have a strong preference

Thanks for testing! I see what you mean and I think it feels like that for me too. I think this problem is more due to the added fade animation rather than the color change though. I added a new commit that removes the animation now: https://github.com/MetabolicAtlas/MetabolicAtlas/commit/54ec9934992c0dac783ca5021184c36625476ab0 @inghylt could you please try this again and see which background you prefer?

MalinAhlberg:
Thanks for sharing the new styles! I think the last one (54ec993 ) is the best, to me it gives the least impression of blinking.

inghylt:
> Thanks for sharing the new styles! I think the last one (https://github.com/MetabolicAtlas/MetabolicAtlas/commit/54ec9934992c0dac783ca5021184c36625476ab0 ) is the best, to me it gives the least impression of blinking.

I agree!

nanjiangshu:
Thanks @e0 for your effort trying to minimizing the blinking by changing the background. It does minimized the blinking feeling slightly but I think the black background gives the user a better feeling of waiting when showing the 3D maps, especially when loading large maps. I would say it's probably better to keep the background as before. 
By the way, I'm not quite sure about the effect of `is-warning` in 
```javascript
<a class="button is-warning is-loading"></a>
```
If that is important, I can add it to this PR.

e0:
> Thanks @e0 for your effort trying to minimizing the blinking by changing the background. It does minimized the blinking feeling slightly but I think the black background gives the user a better feeling of waiting when showing the 3D maps, especially when loading large maps. I would say it's probably better to keep the background as before. By the way, I'm not quite sure about the effect of `is-warning` in
> 
> ```js
> <a class="button is-warning is-loading"></a>
> ```
> 
> If that is important, I can add it to this PR.

Regarding dark vs light. I think avoiding the blinking effect is more important than the better feel of the dark background, especially since it's very relevant for this PR as well. It seems like @MalinAhlberg and @inghylt prefer the light variant too (please correct me if this is the wrong impression).

The `is-warning` modifier is the only variant that makes the loading graphic dark so that it has a good contrast against the light background.

image

nanjiangshu:
Thanks for your explanation @e0! I've added your commit with the white background.

PR #1275, opened 2022-12-15
chore: start on gotenzymes deployment improvements

### LINKED ISSUE #1087: Improve speed of deploying gotenzymes

As a developer, I want to be able to deploy quickly so that I don't waste time and reduce website downtime.

**Acceptance Criteria**

- [ ] Research on possibilities to make the gotenzymes part of the deployment faster, possibly by transferring it zipped ([seems like the size is unreasonably big](https://nbisweden.slack.com/archives/C01D13EKUV7/p1660307090584869?thread_ts=1660301951.150769&cid=C01D13EKUV7)?)
- [ ] Might be beneficial to add option when deploying to leave out the gotenzymes part
- [ ] Currently one has to manually re-import and re-index the database after deployment, this should be automatic based on if there is new data or based on the option from the previous item
- [ ] Possibly better document what is being done when deploying gotenzymes so everyone is fully onboard of what's going on when deploying

**Additional context**

MalinAhlberg:
This PR might close #1087.

**Type of change**  
- [x] Other
 
**List of changes made**  

- Import gotenzymes data during the build stage of `pg`, so that the container is ready on startup, and so that the website is accessible as much as possible (no downtime after redeploying, before the db is read and indexes created etc)
- zip data so that less data is transferred to server (only when deploying, not locally)
- do not touch the pg container on a standard deploy (expect for updating its password)
- a new command `deploy-gotenzymes` is added to update the data on the servers
- `import-gotenzymes` is added to reread the data locally

**TODO**
- I got this to work well on my machine. It also works on the dev server when using a small test db, but when using the real one, I get errors about disk space. The server does not seem to actually be out of disk space, though.
- Verify that zipping actually decreases the time spent on deploying
- What about the cache when running `deploy-gotenzymes`? Does nginx has to be told to drop the cache for gotenzymes?
- Make sure the `pg` container is started/stopped etc when expected (eg check `start-stack` etc command to see if the profile should be added)
- Verify that things actually work as expected
- Cleaning up
- Update documentation

------------

**Testing**  

- Instructions on how to test

**Further comments**  

- [ ] I have performed a self-review of my own code and commented any hard-to-understand areas

### DIFF
--- a/docker-compose-local.yml
+++ b/docker-compose-local.yml
@@ -59,10 +59,11 @@ services:
     environment:
       - FTP_IP=127.0.0.1
       - FLAGS=-Y 0 -d -d
+
   pg:
-    image: postgres:14.3-alpine
-    volumes:
-      - ./pg/scripts:/scripts
-      - ./pg/input_data:/input_data
+    profiles: ["gotenzymes"]
+    build:
+      context: pg
+    command: postgres -c stats_temp_directory=/tmp
     ports:
       - "127.0.0.1:5432:5432"
--- a/docker-compose-remote.yml
+++ b/docker-compose-remote.yml
@@ -60,12 +60,12 @@ services:
           memory: 200M
 
   pg:
+    profiles: ["gotenzymes"]
     build:
       context: pg
+      dockerfile: prod.Dockerfile
     restart: always
     command: postgres -c stats_temp_directory=/tmp
-    volumes:
-      - /var/lib/docker-volumes/pg/postgres-data:/var/lib/postgresql/data
     deploy:
       resources:
         reservations:
--- a/pg/prod.Dockerfile
+++ b/pg/prod.Dockerfile
@@ -1,16 +1,19 @@
 FROM postgres:14.3-alpine as dataloader
 
 
-COPY input_data /input_data/
-
 COPY scripts/init.sql /docker-entrypoint-initdb.d/
 ENV POSTGRES_USER=postgres
 ENV POSTGRES_PASSWORD=change-me
 ENV PGDATA=/data
 
+COPY enzymes-deploy.zip  /input_data/enzymes.zip
+RUN ["rm", "-f", "input_data/enzymes/*", "input_data/supplementary/*"]
+RUN ["unzip", "-oq", "input_data/enzymes.zip"]
+
 # Make sure postrgres is not actually started after data import
 RUN ["sed", "-i", "s/exec \"$@\"/echo \"skipping...\"/", "/usr/local/bin/docker-entrypoint.sh"]
 RUN ["/usr/local/bin/docker-entrypoint.sh", "postgres"]
+
 
 FROM postgres:14.3-alpine
 ####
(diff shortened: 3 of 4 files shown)


docker-compose-local.yml
```diff
@@ -59,10 +59,11 @@ services:
     environment:
       - FTP_IP=127.0.0.1
       - FLAGS=-Y 0 -d -d
+
   pg:
-    image: postgres:14.3-alpine
-    volumes:
-      - ./pg/scripts:/scripts
-      - ./pg/input_data:/input_data
+    profiles: ["gotenzymes"]
```
MalinAhlberg, line 64: the profile is what makes it easy to exclude this container from standard commands (`build-stack`, `deploy-stack` etc.

docker-compose-remote.yml
```diff
@@ -60,12 +60,12 @@ services:
           memory: 200M
 
   pg:
+    profiles: ["gotenzymes"]
     build:
       context: pg
+      dockerfile: prod.Dockerfile
     restart: always
     command: postgres -c stats_temp_directory=/tmp
-    volumes:
```
MalinAhlberg, line 67: do not use volumes anymore, data is kept in the image and will be updated on a rebuild of this

pg/Dockerfile
```diff
@@ -1,4 +1,18 @@
...
 COPY scripts/init.sql /docker-entrypoint-initdb.d/
+ENV POSTGRES_USER=postgres
+ENV POSTGRES_PASSWORD=change-me
+ENV PGDATA=/data
+
+# Make sure postrgres is not actually started after data import
+RUN ["sed", "-i", "s/exec \"$@\"/echo \"skipping...\"/", "/usr/local/bin/docker-entrypoint.sh"]
+RUN ["/usr/local/bin/docker-entrypoint.sh", "postgres"]
+
+FROM postgres:14.3-alpine
```
MalinAhlberg, line 15: use multistage build, where the first step reads and process all data and the second and steals/copies the db data and then runs as a standard postgres container

pg/prod.Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM postgres:14.3-alpine as dataloader
+
+
+COPY scripts/init.sql /docker-entrypoint-initdb.d/
+ENV POSTGRES_USER=postgres
+ENV POSTGRES_PASSWORD=change-me
+ENV PGDATA=/data
+
+COPY enzymes-deploy.zip  /input_data/enzymes.zip
```
MalinAhlberg, line 9: in this dockerfile, the `input_data` folder is not copied, only the zipped file (produced by `proj.sh`)

proj.sh
```diff
@@ -60,11 +60,34 @@ function deploy-stack {
...
 }
 
+function deploy-gotenzymes {
+# Will deploy gotenzymes only
+  trap 'unset DOCKER_CONTEXT; trap - INT' INT TERM
+  CHOSEN_ENV="env-${1:-$LOCALENV}.env"
+  #echo "env is" $CHOSEN_ENV
+  pushd pg
+  rm enzymes-deploy.zip
+  zip -rq enzymes-deploy.zip input_data/enzymes input_data/supplementary
```
MalinAhlberg, line 76: make sure to only include relevant subdirs in zip

nanjiangshu:
Another PR #1285 is created to replace this one

PR #1280, opened 2022-12-19
refactor: rewrite jQuery stuff in plain JavaScript

### LINKED ISSUE #1156: remove jQuery

As a developer, I want to remove jQuery from the project so that there is a consistent way of manipulating the UI and the web app loads faster.

**Acceptance Criteria**

- [x] Get rid of all instances of `jquery` in `frontend`. Replicate the functionality using Vue or vanilla JavaScript.

**Additional context**

This should probably be started after #1144 since there is a few instances of `jquery` that looks like they could be removed (along with `cytoscape`.

e0:
This PR closes #1156.

**Type of change**  

- [x] Other
 
**List of changes made**  

- Rewrite jQuery instances using plain JavaScript in:
  - `frontend/src/components/SearchTable.vue`
  - `frontend/src/components/explorer/mapViewer/Svgmap.vue`
- Remove jQuery from the project (`frontend`)

**Testing**  

- Instructions on how to test
1. Visit a 2D map page, for example `/explore/Human-GEM/map-viewer/golgi_apparatus?dim=2d`.
2. Verify that the map works as before. Please try different interactions such as using data overlay, mouse hovering and clicking, toggles, search etc.
3. Visit a global search page, for example `/search?term=mitoc`.
4. Verify that the search works as before.

**Further comments**  

This change shaves another `~12.6 %` off the bundle size (`frontend/stats.html`), going from `2.22 MB` to `1.94 MB`.

### DIFF
--- a/frontend/src/components/SearchTable.vue
+++ b/frontend/src/components/SearchTable.vue
@@ -315,7 +315,6 @@
 
 <script>
 import { mapGetters, mapState } from 'vuex';
-import $ from 'jquery';
 import 'vue-good-table-next/dist/vue-good-table-next.css';
 import { VueGoodTable } from 'vue-good-table-next';
 import Loader from '@/components/Loader.vue';
@@ -633,7 +632,7 @@ export default {
     next();
   },
   updated() {
-    $('#search').focus();
+    document.getElementById('search').focus();
   },
   methods: {
     fillFilterFields() {
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -22,7 +22,6 @@
     "bulma-timeline": "^3.0.5",
     "eslint-config-prettier": "^8.5.0",
     "file-saver": "^2.0.2",
-    "jquery": "3.6.0",
     "nprogress": "0.2.0",
     "vue": "^3.2.0",
     "@ckpack/vue-color": "1.2.0",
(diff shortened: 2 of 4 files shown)


frontend/src/components/SearchTable.vue
```diff
@@ -633,7 +632,7 @@ export default {
     next();
   },
   updated() {
-    $('#search').focus();
+    document.getElementById('search').focus();
```
inghylt, line 635: Maybe silly question,  but what functionality does this line contribute with?

e0: This focuses the search input element each time after a search is performed. I assume it's convenient for people who are performing multiple searches. Personally I don't know how often this is needed but at the same time I don't think it hurts anything to have it.

nanjiangshu (approved):
I don't find any problem after the refactoring. Nice that we have a consistent code base.

e0:
> One question though, the results are ordered a bit differently on my local computer compared with prod in global search (try searching for `mul`, `gem` for example and compare different pages). It's not a major difference and it is present on `main` as well. Is it only on my local machine or do you guys have the same behavior? If so, should we look into this?

Thanks for the careful testing and that's a good catch! I can notice the difference between my `localhost` and the `prod` server as well. It seems like `localhost` matches the `dev` server for me though. So I think we can at least conclude that it's not introduced in this PR. We should do some more investigation into this and open up an issue if necessary.

e0:
Issue #1282 created

PR #1285, opened 2022-12-21
Chore: improve gotenzymes deployment

### LINKED ISSUE #1087: Improve speed of deploying gotenzymes

nanjiangshu:
This PR closes #1087 

Since the strategies are quite different from the draft PR #1275, a new branch and PR is created. 

**Type of change**  

- [x] improvement of the deployment
 
**List of changes made**  

Given the condition that the GotEnzymes input_data will not be changed frequently in the future, the following changes have been made to avoid frequent data transferring and rebuilding of the GotEnzymes database. 

1. `input_data` is not copied to the container but mounted at  `/var/lib/docker-volumes/pg/input_data/` for the remote server, by doing so, the size of the image `pg` is reduced from 3.04 GB to 216 MB. `input_data` is uploaded to the remote server by `rsync` and do not need to be transferred through docker at every deployment. 
2. For local deployment, the GotEnzymes database is mounted `pg/postgres-data`. The container `pg` is based on the same image `pg` as for the remote deployment. By doing so, the GotEnzymes database will be built only at the first `start-stack` but will not run if it already exists 
3. A command `update-gotenzymes` is added to update the GotEnzymes database either locally or remotely 
    - `update-gotenzymes` for updating the database locally 
    - `update-gotenzymes REMOTE-CONTEXT` for updating the database remotely 
    
   For Mac M1 user, please select `VirtioFS` for file sharing to speed up postgresql database building process.
    
4. `docker system prune --all` is run at each `deploy-stack` to avoid out of disk space problem.

In addition, some verification is added to the helper script, e.g. `deploy-stack` will return the usage of `context` is not provided. The README is also updated accordingly.

**Screenshot of the fix**  

**Testing**  

`source proj.sh` before running the test
For locally deployment, also need to rebuild the pg. 
Test also starting from a clean docker setup. 

**Further comments**  

1. The time of `update-gotenzymes` is normally when running on the remote server with Ubuntu 20.04

```
time -p update-gotenzymes dev2: real 1183.84 user 0.36 sys 0.23
time -p update-gotenzymes dev: real 629.97 user 0.30 sys 0.16  
```

However, on my local machine with Mac M1, it takes 3 hours to build the database at first run of `start-stack`. It seems postgres is using 1-2% of CPU when running the building procedure in the background. I don't know if it happens only to my laptop. Nevertheless, it needs to be done only once. 

2. Regarding the deployment speed, it takes about half the time when there is a clean stack locally. However, when you have deployed to the remote server once, the old setup is sometimes faster. It seems the 3GB pg image is not transferred at all when cache exist. 

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ update-gotenzymes
 For remote servers, the init script is configured to run automatically if the database has not been initialized. To reconstruct the database, delete the mounted volume for the database on the remote server (located at `/var/lib/docker-volumes/pg/postgres-data`) and deploy again.
 
 ## Description of helper commands
-**Note that the following commands should be run in a bash shell**
+**Note that the following commands should be run in a bash shell** (the current shell can be replaced using: `exec bash`)
 - To bootstrap the project: `build-stack`
 - To run the project: `start-stack`
 - To stop the project: `stop-stack`
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ Finally, start the Docker containers with
 start-stack
 ```
 
-Given successful deployment, the frontend should be accessible at: `http://localhost/`. If you encounter any problems try looking at the logs `logs api` / `logs frontend`.
+Given successful deployment, the frontend should be accessible at: `http://localhost/`. It may take a while before the GotEnzymes database is ready if it has not been built before. If you encounter any problems try looking at the logs `logs api` / `logs frontend`.
 
 To deploy the stack to a remote server, create another `ENV` file, e.g. `env-dev.env`, and modify it accordingly:
 
@@ -92,7 +92,7 @@ deploy-stack dev
 To reconstruct the database for GotEnzymes on the local (development) machine. Run the following. This should take ~10 minutes.
 
 ```bash
- ma-exec pg psql -f scripts/init.sql -U postgres
+update-gotenzymes
 ```
 
 For remote servers, the init script is configured to run automatically if the database has not been initialized. To reconstruct the database, delete the mounted volume for the database on the remote server (located at `/var/lib/docker-volumes/pg/postgres-data`) and deploy again.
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -89,13 +89,13 @@ deploy-stack dev
...
 ```bash
- ma-exec pg psql -f scripts/init.sql -U postgres
+update-gotenzymes
 ```
 
 For remote servers, the init script is configured to run automatically if the database has not been initialized. To reconstruct the database, delete the mounted volume for the database on the remote server (located at `/var/lib/docker-volumes/pg/postgres-data`) and deploy again.
 
 ## Description of helper commands
-
+**Note that the following commands should be run in a bash shell**
```
e0, line 98:
```suggestion
**Note that the following commands should be run in a bash shell** (the current shell can be replaced using: `exec bash`)
```

e0 (approved):
Excellent work @nanjiangshu!

This worked really smoothly for me both locally and when deploying to the dev server. I love that you used `rsync` for file transfer to minimize the transfer size even when the data updates.

On my M1 Mac the time to build the database was not any longer compared to before. For reference, I'm running macOS 13.1, Docker Desktop v4.15.0. I have selected `VirtioFS` as the file sharing implementation in Docker Desktop settings (General).

inghylt:
It works well locally! For some reason my computer won't let me deploy, but that's a problem present on main as well so not it's not a specific problem with this branch. But since I can't verify how deployment is affected I won't approve of this.

I also think we should wait with merging this until we know whether @mihai-sysbio and members of the Aspeboda team are able to deploy using this branch.

mihai-sysbio (approved):

Issue #718, opened 2021-10-18
Add / improve script for installation help


MalinAhlberg:
As a developer or new user, I want a helpful script to guide me through installation, so that I can get started faster.

The basic idea is to update `proj.sh` to perform installation checks (eg checking versions, data paths etc).
Related to https://github.com/MetabolicAtlas/data-generation/pull/25,
especially comments about connecting the script to the function `generate-data` in `proj.sh` ([this](https://github.com/MetabolicAtlas/data-generation/pull/25#pullrequestreview-777082014), [this](https://github.com/MetabolicAtlas/data-generation/pull/25#issuecomment-941935940) and [this](https://github.com/MetabolicAtlas/data-generation/pull/25#issuecomment-940751282)).

PR #1296, opened 2023-03-08
fix: Address the issues mentioned in Github issue #1295

### LINKED ISSUE #1295: Several observed bugs in `proj.sh` relating to code injection, code duplication, the use of unquoted variables, possible interfearence with external code via environment variables, readability, etc.
**Describe the bug**

The `proj.sh` dot-script is supposed to be sourced (either via `source proj.sh` or `. ./proj.sh`) to install several convenient shell functions into the current shell session, along with numerous variables used by the project. These are shell functions that a developer is expected to use several times during an ordinary day of working with this project.  

I have collected several issues together with suggestions for solving most of these:

- The dot-script pollutes the current shell environment with unnecessary variables (some that contain passwords). It also modifies the `PATH` variable by adding `/usr/local/bin` to the end every time the script is sourced. This is an issue since developers may need to move between different projects and may inadvertently affect the running of unrelated code by the presence of certain variables or shell functions.  Depending on the shell's configuration, it may also expose passwords to any software started from the developer's shell. 
	- It would be better if sourcing the script only added the necessary functions to the environment without modifying existing environment variables or adding new ones.  
	- The environment could be made local by running the functions in sub-shells. Each function that needs it could initialise its local environment by calling a common `_setup-envinoment` function. This way, sourcing the `proj.sh` script *only* adds the shell functions defined within the script and does not pollute the environment with passwords etc.  
	- A `deactivate` function could be implemented that removes the installed shell functions from the environment. This function would not need to reset any shell or environment variables as none were changed.  
- Several variables are unquoted in the code. This makes it impossible to e.g. work with directories with spaces in their paths or passwords containing spaces, tabs, filename globbing characters, etc. This affects some uses of `$CHOSEN_ENV`, `$DATA_GENERATOR_PATH`, `$DATA_FILES_PATH`, `$NEO4J_USERNAME`, `$NEO4J_PASSWORD`, and a few uses of `$@`.  
	- Quoting these variables is possible.  
- There are two instances of code injection. Neither of these is (probably) a serious matter, but both could easily result in errors if a password happened to include quotes. Code injections should be fixed as a matter of principle, as assessing the real security impact is difficult.  
	- When `cypher-shell` is called within an in-line `bash -c` script, executed in the `neo4j` container, the two option-arguments `$NEO4J_USERNAME` and `$NEO4J_PASSWORD` are injected by the parent shell into the in-line script (their use is also unquoted which prompts the shell to split their values on whitespace and to apply filename globbing on the result).  
		- Either pass these two values as arguments to the `bash -c` script or, better yet, do not wrap the `cypher-shell` invocation in a `bash -c` script. In either case, the two arguments should be quoted.  
	- When `psql` is executed inside the `pq` container to change a password, `$POSTGRES_PASSWORD` is injected by the parent shell into the SQL statement that performs the change.  
		- We may parametrise the `ALTER USER` statement to use a variable set on the `psql` command line using `-v` or `--variable`. The `psql` utility would take care of sanitising the values of such variables. Note that the SQL statement has to be passed to `psql` via the command's standard input stream, as the statement given by `--command` or `-c` is not processed for parameter placeholders.  
- The command `docker compose` is being called with exactly the same arguments in a number of places.  
	- This may be extracted into a helper function, `_docker-compose`, that does the `docker compose` call with the static arguments, with the given variable arguments added to the end of the argument list. This not only gets rid of code duplication but also helps with readability, as it results in code with much shorter lines.  
- Regarding code reuse, all instances of `docker exec` and `docker compose exec` could look virtually the same.  
	- Reuse the `ma-exec` function for this.  
- The command `docker` may not have a `compose` sub-command on some systems where a `docker-compose-cli` package (or similar) is not available or has not been installed.  
	- Use `docker-compose` instead of `docker compose`, for the developer's convenience, and set the `DOCKER_BULDKIT` to `1` for the command to use [the improved BuildKit backend](https://docs.docker.com/build/buildkit/), just like `docker compose` would do by default.  
- There are some instances of the standard utility `cp` being called with an absolute path. This is inelegant.  
	- If there is an issue on some platform, for some developer, causing the wrong `cp` command is being used, then `/bin` should be added *first* in the `$PATH` variable and `command cp` should be used rather than just `cp`.  This ensures that the external command `cp` is called rather than a shell function or alias and that the shell would first search for this command in `/bin`.  
- There is a comment about a `-q` option in the `generate-data` function:  
  > enable flag "-q" to force overwritting existing data files  

    
  This comment is either out of place, unneeded, or should refer to the `--reset-db` option rather than `-q`.  
	- Correcting this comment would probably be the best approach.  
- There are some instances of using a mix of long and short options with `docker compose`, which reduces the code's readability and may cause confusion.  
	- To avoid confusion, spell `-d` as `--detach`, `-v` as `--volumes`, and `-f` as `--follow` in the relevant calls. Also, use long options with other non-standard commands where it serves to document the code.  
- Two `docker compose` calls (in `deploy-stack` and in `start-stack`) have no associated error handling, which means that if they fail, the rest of the code in the function would likely also fail. Likewise, the call to `generate-data` in `deploy-stack` and `import-db` may also fail.  
	- Selective use of `set -e` would allow us to bail out of functions at the first failing command. Note that all functions do not need this (typically only functions with multiple steps). Also, note that the effect of this shell option would be localised if each function executed within its own sub-shell as discussed at the start.  
- [Docker compose environment files](https://docs.docker.com/compose/environment-variables/env-file/) are not generally shell scripts. The shell code disregards this and sources the environment file.  
	- Using two separate environment files would not be practical, even though creating the Docker environment file from a shell script would be trivial.  
	- Since quoting works very differently in the two file formats, the passwords in the environment file will need to be picked carefully.
- Shell function names should not contain dashes.
  - We may have to assume that the developer uses a shell that allows this, to remain backward compatible.

kusalananda:
This PR closes #1295.

**Type of change**  

- [ ] Other
 
**List of changes made**  

This will be filled out soon.

### DIFF
--- a/proj.sh
+++ b/proj.sh
@@ -1,83 +1,163 @@
-# To make sure docker-compose is in the path
-export PATH=$PATH:/usr/local/bin
-LOCALENV="local"
-METATLAS_DEFAULT_ENV="env-${LOCALENV}.env"
-CHOSEN_ENV=$METATLAS_DEFAULT_ENV
+_setup-environment () {
+  # Helper function that initializes the needed environment variables.
+  "${got_environment:-false}" && return
+  got_environment=true
+  
+  # Ensure that both docker and docker-compose is in $PATH (assumed to
+  # be living in /usr/local/bin), and that /bin is first.
+  PATH=/bin:$PATH:/usr/local/bin
 
-function generate-data {
-  # enable flag "-q" to force overwritting existing data files
-  echo "Using $CHOSEN_ENV"
-  source $CHOSEN_ENV && yarn --cwd $DATA_GENERATOR_PATH start $DATA_FILES_PATH "$@"
-  /bin/cp -rf $DATA_GENERATOR_PATH/neo4j/* neo4j/import
-  /bin/cp -rf $DATA_GENERATOR_PATH/dataOverlay api/
-  /bin/cp -rf $DATA_GENERATOR_PATH/gemRepository/* frontend/public/assets/gemRepository/
-  /bin/cp  -f $DATA_FILES_PATH/integrated-models/integratedModels.json api/src/data/
-  /bin/cp  -f $DATA_FILES_PATH/gemsRepository.json api/src/data/
-  /bin/cp -rf $DATA_FILES_PATH/svg api/
-  /bin/cp -rf $DATA_FILES_PATH/repository ftp/
-  /bin/cp -rf $DATA_FILES_PATH/repository api/
+  # Bail out if docker or docker-compose are not found.
+  if ! { command -v docker && command -v docker-compose; } >/dev/null
+  then
+    echo 'Missing essential Docker components' >&2
+    return 1
+  fi
+
+  LOCALENV=local
+  CHOSEN_ENV=./env-$LOCALENV.env
+
+  printf 'Using %s\n' "$CHOSEN_ENV"
+  . "$CHOSEN_ENV" || return
 }
 
-function build-stack {
+_docker-compose () (
+  set -e
+
+  # Helper function to reduce clutter and repetition.
+  _setup-environment
+
+  DOCKER_BUILDKIT=1 docker-compose --env-file "$CHOSEN_ENV" \
+    -f docker-compose.yml \
+    -f docker-compose-local.yml \
+    "$@"
+)
+
+generate-data () (
+  set -e
+
+  _setup-environment
+
+  # Call generate-data with the option --reset-db to force overwriting
+  # existing data files.  See ../data-generation/index.js
+  yarn --cwd "$DATA_GENERATOR_PATH" start "$DATA_FILES_PATH" "$@"
+
+  command cp -rf "$DATA_GENERATOR_PATH"/neo4j/* neo4j/import
+  command cp -rf "$DATA_GENERATOR_PATH"/dataOverlay api/
+  command cp -rf "$DATA_GENERATOR_PATH"/gemRepository/* frontend/public/assets/gemRepository/
+  command cp  -f "$DATA_FILES_PATH"/integrated-models/integratedModels.json api/src/data/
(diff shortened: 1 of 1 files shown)


mihai-sysbio:
Since the `proj.sh` file has been buggy at multiple times in the past, I would add to the requirements that _everyone_ tests the commands (especially if using different operating systems).

Food for thought: is this a good time to also resolve #718? It's not part of this sprint, but it if you feel it would be more efficient to resolve it now, you have my support to go for it.

kusalananda:
> Since the `proj.sh` file has been buggy at multiple times in the past, I would add to the requirements that _everyone_ tests the commands (especially if using different operating systems).
> 
> Food for thought: is this a good time to also resolve #718? It's not part of this sprint, but it if you feel it would be more efficient to resolve it now, you have my support to go for it.

Yes, it makes sense to address #718. I will do this in one or several further commits. Thanks.

kusalananda:
> Since the `proj.sh` file has been buggy at multiple times in the past, I would add to the requirements that _everyone_ tests the commands (especially if using different operating systems).
> 
> Food for thought: is this a good time to also resolve #718? It's not part of this sprint, but it if you feel it would be more efficient to resolve it now, you have my support to go for it.



> > Since the `proj.sh` file has been buggy at multiple times in the past, I would add to the requirements that _everyone_ tests the commands (especially if using different operating systems).
> > Food for thought: is this a good time to also resolve #718? It's not part of this sprint, but it if you feel it would be more efficient to resolve it now, you have my support to go for it.
> 
> Yes, it makes sense to address #718. I will do this in one or several further commits. Thanks.

Now implemented along with ideas and suggestions found in comments to #1295

jonandernovella (approved):
Really nice refactor, it was truly needed!

KattisLej:
Starting review

KattisLej (approved):
Works very good, the install-check was a super nice command

PR #1299, opened 2023-03-08
Vue Typescript and Composition API POC

perjo:
This is POC to investigate how to use TypeScript in the frontend code of the Metabolic Atlas project. 

**Why**: [TypeScript](https://www.typescriptlang.org/) gives type-safety and better help from your IDE. 

**How, simple answer**: Just add `lang = "ts"` to the script tag in the component.  This can be done one component by one, no need for a full rewrite. 

**How, full answer**: To get the best TypeScript support we should use the [Composition API](https://vuejs.org/guide/extras/composition-api-faq.html) to write components. We should do this anyway, since this is [the preferred way to write components in Vue 3](https://vuejs.org/guide/extras/composition-api-faq.html#why-composition-api). 

**Suggested way forward**: 

- When making new components, always use TypeScript and Composition API. 
- Migrate not yet migrated components while working with them. 

**Type of change**  
- [x] Other
 
**List of changes made**  

- Added TypeScript to project
- Used Composition API and TypeScript in component frontend/src/components/explorer/mapViewer/DataOverlayValidation.vue 
- Configured ESLint to report violations on WARN level.

### DIFF
--- a/frontend/.eslintrc.js
+++ b/frontend/.eslintrc.js
@@ -3,7 +3,7 @@ module.exports = {
   env: {
     node: true,
   },
-  extends: ['plugin:vue/vue3-recommended', '@vue/airbnb', 'prettier'],
+  extends: ['@vue/typescript/recommended', '@vue/airbnb', 'prettier'],
   rules: {
     'no-console': process.env.NODE_ENV === 'production' ? 'error' : 'warn',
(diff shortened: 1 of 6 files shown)


dbampalikis:
Closing since the PR is not currently relevant according to prioritisation

PR #1304, opened 2023-03-10
Fix: Remove unused code

perjo:
**Related issue(s) and PR(s)**  
This PR closes https://app.zenhub.com/workspaces/metatlas-sprint-607745622d49260019ce115a/issues/gh/metabolicatlas/private-issues/37

**Type of change**  
- [x] Other: Removal of unused code 
 
**List of changes made**  

Removed code which seems not to be used. 

**Screenshot of the fix**  

**Testing**  

Regression test the whole application. 

**Further comments**  

**Definition of Done checklist**  
- [x] My code meets the Acceptance Criteria
- [x] My code follows the [NBIS style guidelines](https://github.com/NBISweden/development-guidelines)
- [x] I have performed a self-review of my own code and commented any hard-to-understand areas
- [x] Tests and lint/format validations are passing
- [x] My changes generate no new warnings

PR #1313, opened 2023-03-14
Revert "Fix: Remove unused code"

perjo:
Reverts MetabolicAtlas/MetabolicAtlas#1304

Since things seem to have been broken by this merge we can conclude that there is no safe way to detect unused code in this code base.

### DIFF
--- a/frontend/src/components/Repository.vue
+++ b/frontend/src/components/Repository.vue
@@ -358,6 +358,22 @@ export default {
     }
   },
   methods: {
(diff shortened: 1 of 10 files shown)


KattisLej (approved):
I think this is a good idea

mihai-sysbio:
How about removing the unused CSS that was detected in the original issue?

jonandernovella (approved):

PR #1316, opened 2023-03-15
fix: indicate when MapsAvailable component is scrollable

### LINKED ISSUE #1114: For the Map viewer side table, indicate when map listing exceeds the area

As a user, I want a visible indication (e.g. a scrollbar be shown on hover) when the map listing in the Map Viewer exceeds the area so that I know there are hidden items. 

For example, for the webpage https://metabolicatlas.org/explore/Human-GEM/gem-browser/metabolite/MAM02026c
it is unclear to the user that there are more items in the map listing without scroll the area. 

One way to solve could be using pagination.

**Acceptance Criteria**

- [ ] An indication is shown on hover if there are hidden items (suggested to not force scrollbar to be shows but some other visual indicator)
- [ ] The changes are tested on multiple browsers and OS

**Additional context**

Scrollbar is not shown also for other webpages, and it seems it is a default behavior for web-browsers.
Maybe it is natural for people to scroll a webpage even there's no scrollbar, so the fix do not need to extend to other places.

pdmnyberg:
This PR closes #1114

It solves the problem by adding a small shade as indication of scrollable content in the MapsAvailable component. This is mostly useful for platforms which hide the scrollbars of elements.

**Type of change**  
 
**List of changes made**  
- Add top and bottom indicators of hidden content for the MapsAvailable component

**Screenshot of the fix**  
scroll_combine

**Testing**  
- Rebuild project
- Make sure the indicators work for: http://localhost/explore/Human-GEM/gem-browser/metabolite/MAM02026c
- Make sure the indicators are missing for: http://localhost/explore/Human-GEM/gem-browser/metabolite/MAM00162c


mihai-sysbio (commented):
I've looked at this on all major browsers on macOS and it works as expected 👍🏻 

In mobile view I would be tempted to enlarge the grey gradient surface - could you have a look at that @pdmnyberg?

Also, a general comment about event listeners: even though these are triggered in a component (and nicely removed when unmounted), they act as "global" triggers. So if there is any way to avoid it, it's generally preferred (no change is requested here, to be clear).

Another thing to keep in mind is that there are also other places in the website that are scrollable and have the same issue as here (the user not knowing if the area is scrollable). So if there is any way of making this code more generic that would be good (perhaps though this could be taken up in a new issue?).

One last idea: would it make sense to use [scrollLeft](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollLeft)?

pdmnyberg:
I agree that global triggers are not that great. In this case I am unsure of how to respond to layout changes without it though. It might be that I need to refresh some of my CSS and JS knowledge. So if anyone has a suggestion I am interested in hearing it :)

I think that it might be possible and preferrable to break this out into a component or maybe some utility concept but I think I would need a few more cases to look at in order to know how to generalize it in a way that fits with the current code base. A new issue for that seems like a good idea. Maybe you could include know examples of more scroll areas in such an issue @mihai-sysbio 

By `scrollLeft` do you mean generalizing the concept to support horizontal scrolling? Otherwise I am not sure exactly what you mean. (In general I think one should avoid scrolling on more than one axis at a time so I would expect to use either `scrollLeft` or `scrollTop`)

KattisLej (approved):
Great improvement, well done!

pdmnyberg:
@mihai-sysbio I agree that the size of the gradient on mobile could benefit from being larger so I tried doubling it for `$tablet` sizes. I think it looks a bit better.

perjo (commented):
Codewise good and I guess it does what is expected...BUT...I would not understand that there was scrollable content from this shade! 

To me, this hiding of the scrollbar is a crazy design decision in macOS (possibly elsewhere). Shouldn't try we set up a permanently visible scrollbar instead? I guess there are [hacks](https://fjorgedigital.com/insights/blog/how-to-always-show-an-inline-scroll-bar-on-mac/) available to accomplish this? I realize there are downsides to this as well. 

Or maybe add an overlay text saying "more content below" or something?

pdmnyberg:
> Codewise good and I guess it does what is expected...BUT...I would not understand that there was scrollable content from this shade!
> 
> To me, this hiding of the scrollbar is a crazy design decision in macOS (possibly elsewhere). Shouldn't try we set up a permanently visible scrollbar instead? I guess there are [hacks](https://fjorgedigital.com/insights/blog/how-to-always-show-an-inline-scroll-bar-on-mac/) available to accomplish this? I realize there are downsides to this as well.
> 
> Or maybe add an overlay text saying "more content below" or something?

Fair point @perjo I am fairly used to seeing this kind of solution to indicate more content so I might be a bit biased. There might be other solutions to investigate when it comes to showing the actual scrollbar but I think that might be more efficient to explore on a mac.

My personal preference is as follows `native solution?` > `current solution` > `pagination` > `custom scrollbar` where I would like to avoid a `custom scrollbar` as much as possible.

Maybe I could try adding the text as a start and then we'll see if the current solution feels hopelessly lost.

mihai-sysbio:
I think the situation is worse on tables/phones, see this entry on desktop for a longer map viewer list https://metabolicatlas.org/explore/Human-GEM/gem-browser/gene/ENSG00000161533

image

Another idea would be to add an arrow icon in the center of the greyed area.

perjo (approved):
These indicators I do understand! Well done!

PR #1321, opened 2023-03-18
fix: use correct compose and environment files

### LINKED ISSUE #1319: proj.sh can no longer do remote deployments
**Describe the bug**

After @kusalananda 's refactor, we are missing the possibility of doing remote deployments using the env-dev.env environmental file + the docker-compose-remote.yml

kusalananda:
This PR closes #1319 

When calling `deploy-stack` (as e.g. `deploy-stack dev`), the compose file `docker-compose-local.yml` was used unconditionally. Instead, the `docker-compose-remote.yml` should have been used. It is also unclear whether the correct environment file was being used. 

**Type of change**  

- [ ] Other
 
**List of changes made**  

The "internal" function `_docker_compose` can now take an optional first argument that can be either of the two strings `remote` or `local`.  Depending on what string is used as the first argument, the corresponding compose file will be used.  If no first argument is given, or if the first argument is neither the string `remote` nor `local`, then the `docker-compose-local.yml` file will be used, as usual.

The "internal" function `_setup-environment` no longer modifies `CHOSEN_ENV` if this variable is already set (by e.g. `deploy-stack`). This makes it possible for `deploy_stack` to pre-set `CHOSEN_ENV` to the environment file chosen by the user (or to the local environment file if no file is picked).

The `LOCALENV` state variable was removed.

There was also a minor change to a tangentially related variable name since it was unclear by a casual glance what this variable actually contained (`compose` was changed to `compose_cmd`).

**Testing**  

- Instructions on how to test

When testing, remember to re-source the `proj.sh` dot-script using either `source proj.sh` or `. ./proj.sh`.  Also remember that the `deploy-stack` function takes an optional argument which influences what environment file to use, so use e.g. `deploy-stack dev` to use `env-dev.env` file.

I have tested this locally, but have obviously not tried to do a real deployment using the update code.


mihai-sysbio:
Thanks for the fix @kusalananda. I haven't waited for the deploy to finish but it indeed picking up the right environment file. To be clear, this is not a review, just an intro for the following:


I'm just wondering about the implications of the following line:

> remember that the deploy-stack function takes an optional argument which influences what environment file to use

What are then the differences between `start-stack` (never takes args) and `deploy-stack` (when run with no args?

kusalananda:
> Thanks for the fix @kusalananda. I haven't waited for the deploy to finish but it indeed picking up the right environment file. To be clear, this is not a review, just an intro for the following:
> 
> I'm just wondering about the implications of the following line:
> 
> > remember that the deploy-stack function takes an optional argument which influences what environment file to use
> 
> What are then the differences between `start-stack` (never takes args) and `deploy-stack` (when run with no args?

If `deploy-stack` is run without command line arguments, it will pick up the `env-local.env` environment file and use the `docker-compose-remote.yml` compose file. With an argument, it would use the environment file given by `env-<argument>.env` (e.g. `env-dev.env`) and the same compose file as when given no argument. This is the same behaviour as in the old code.


> 
> _edit_: Actually, the technical differences are not my key concern. I'd like to ensure that the list of available commands is simple and clean (no duplicates) for the benefit of anyone trying the stack.

The difference between `start-stack` and `deploy-stack`, apart from what environment and compose files are being used is that `start-stack` uses the following options with `docker compose`:
```
up --detach
```

`deploy-stack` uses
```
--project-name metabolicatlas \
up --detach --build --force-recreate --remove-orphans --renew-anon-volumes
```

Note that this uses `--build` and `--force-recreate` to force rebuilding the containers. I have not modified these options from the original code.

The `deploy-stack` call additionally modifies the password of the `postgres` user in the PostgreSQL database, setting it to whatever `$POSTGRES_PASSWORD` happens to be.  The `start-stack` call won't do that.

kusalananda:
> Good work! The deploy stack command now works on the development server.
> 
> However, at the moment I am not able to stop the remote stack. It seems as if it's talking to the local context by default.

Hmm... I wonder what the best solution would be. Would it, for example, be best to stop the deployed service as part of the `deploy-stack` call, so that a deployed stack is stopped before it is started again, or do you need a separate `stop-deployed` call?  Or `stop-stack dev`, or `deploy-stack stop`, or something similar?

jonandernovella:
> > Good work! The deploy stack command now works on the development server.
> > However, at the moment I am not able to stop the remote stack. It seems as if it's talking to the local context by default.
> 
> Hmm... I wonder what the best solution would be. Would it, for example, be best to stop the deployed service as part of the `deploy-stack` call, so that a deployed stack is stopped before it is started again, or do you need a separate `stop-deployed` call? Or `stop-stack dev`, or `deploy-stack stop`, or something similar?
> 
> If there is a more natural way of organising the start/stop functions, I'd be happy to look into sorting this out. For example, we could have a `stack-start` and a `stack-stop` set of function that take an optional argument. If there is an argument, these would do deployment, otherwise they would do local start or stop. This would be a 5 minute job for me to put together since all the bits are already in the code.

That sounds reasonable to me. In that case, the corresponding documentation should be updated with the updated commands.

mihai-sysbio:
The existence of a remote stop-stack is beyond the scope of this issue, so let's take this part as a new issue in a backlog refinement.

As a practical solution, one can use `docker context` to set the remote context of `dev`, and that shell will execute the docker commands on the remote machine.

jonandernovella (approved):
Works as expected!

perjo (approved):
Looks good!

PR #1324, opened 2023-03-21
fix: Explore performance improvements wrt memory allocations

jonandernovella:
Closes #1324 and https://github.com/MetabolicAtlas/private-issues/issues/31

I updated the dev environment’s memory settings and compared the old settings with the new ones by doing performance tests.Also, the locust tests were updated to reflect the new endpoints.
 
## The data

Here you can see failure statistics of a 2 minutes run with 100 users over the listed endpoints:
- on the left the current production settings:
      - NEO4J_dbms_transaction_timeout=40s
      - NEO4J_dbms_lock_acquisition_timeout=40s
      - NEO4J_dbms_memory_heap_initial__size=5G
      - NEO4J_dbms_memory_off__heap_max__size=8G
      
- on the right are the new recommended settings
      - NEO4J_dbms_transaction_timeout=40s
      - NEO4J_dbms_lock_acquisition_timeout=40s
      - NEO4J_dbms_memory_heap_initial__size=4700m
      - NEO4J_dbms_memory_off__heap_max__size=4700m
      - NEO4J_dbms_memory_pagecache_size=4900m

      
image
      
And the response time statistics:

bild

      

## The conclusions

- The differences look pretty minor. And the resource allocations are also considerably lower
- The response times also look very similar. There are only one or two outliers that we have disregarded for this analysis.
- I would say this test a good start. Perhaps an indication that we overallocated resources for the current database.
- Regarding the recommendation: I would say that we could try with more pagecache settings and see if it gets any better. Also, if it doesn’t we should possibly explore scaling neo4j
- The interesting bit perhaps is that with both setups failures start with ~30 concurrent users


mihai-sysbio:
@jonandernovella great job providing this info!

I was wondering if there were any stats on the response time - how does that change up until the 30 user mark?

pdmnyberg (approved):
LGTM

mihai-sysbio:
@jonandernovella thanks for updating the issue with the response times.

PR #1325, opened 2023-03-22
Fix/1322 make links behave consistently

### LINKED ISSUE #1322: Prevent opening anchor tag links in new window
**Describe the bug**
As a user I don't want the application to open a new tab each time I click on a link.

When using anchor links the new page opens in a new window.

- [ ] use Shan’s way of dealing with `<a> tags and click.prevent=”e =>f(e)”` on the tags in the GEM Browser

Code example:
https://github.com/MetabolicAtlas/MetabolicAtlas/blob/cafee5f4189b09171d1ef4ecd343[…]43ac8/frontend/src/components/explorer/gemBrowser/GemSearch.vue

Problem example:
https://metabolicatlas.org/explore/Human-GEM/gem-browser/reaction/MAR09296 

**Expected behavior**
- [ ] A  link should not open a new window

**Additional context**

pdmnyberg:
This PR closes #1322

Link targets have been updated to behave consistently as follows:
- Local links open in the current window/tab
- External links open in a new window/tab

**Type of change**  
 
**List of changes made**  
- Add cypress tests of links
- Remove old non-working cypress tests
- Make sure tested vue components work well with cypress expectations
- Make sure all links follow the guideline -> external links `target="_blank"`

**Testing**  
- Check cypress tests
- Run cypress tests `cd frontend && npx cypress run cypress`
- Do some clicking around


Jessica-Sandler (approved):
Looks very good and everything seems to work as expected!

frontend/cypress/e2e/links.cy.js
```diff
@@ -0,0 +1,52 @@
...
+context('Links on site', () => {
+  const routes = [
+    "/",
+    "explore/Human-GEM/gem-browser",
+    "/explore/Human-GEM/gem-browser/metabolite/MAM01282c",
+    "/explore/Human-GEM/map-viewer/golgi_apparatus?dim=2d&panel=0&sel=&search=&coords=-4012.67,-15817.92,0.2,0,0,500&dataTypes=None&dataSources=None&dataSets=None",
+    "/explore/Human-GEM/interaction-partners",
+    "/explore/Human-GEM/interaction-partners/MAM01632e",
+    "/gems/repository",
+    //"/gems/repository/Fruitfly-GEM",
```
mihai-sysbio, line 12: Is there any special reason this is commented out?

frontend/index.html
```diff
@@ -1,5 +1,8 @@
 <!DOCTYPE html>
 <html lang="en">
+<head>
+  <meta charset="UTF-8">
+</head>
```
mihai-sysbio, line 5: This already exists in the production site - it's probably added dynamically. Are you sure it should be added here?

frontend/src/components/about/Citation.vue
```diff
@@ -65,23 +67,21 @@ import { doiref } from '@/helpers/utils';
...
-      doiref,
-    };
-  },
   props: {
     entry: {
       type: Object,
     },
   },
-  methods: {
+  computed: {
```
mihai-sysbio, line 75: 👍🏻

mihai-sysbio (commented):
Very nice cleanup of the links!

There's really one comment that I think needs follow-up, the addition of `<head>` in `index.html`. I'd like to make sure this is not breaking how the current `<head>` is set up.

frontend/index.html
pdmnyberg: Good question. I was relying on the old cypress tests which seemed to want that change at least in my local dev environment. I'll have a look and see if I can find how it is added in the production site. (Could be that the test is just old and needs to be removed I guess)

frontend/cypress/e2e/links.cy.js
pdmnyberg: Thank you for pointing that out. I should have reported that as a bug but forgot about it while looking into other problematic pages. I think there is a problem when navigating to that URL directly without passing `/gems/repository`. I think it should be reported as a bug separately since it seems to be a problem in `main` already.

mihai-sysbio: +1 for reporting that bug separately (with a note on amending the cypress test).

frontend/index.html
pdmnyberg: This was a mistake. I have reverted the change and updated the test to work with the current site.

mihai-sysbio: Thanks for fixing it!

mihai-sysbio:
While I haven't tested this PR myself, all my comments have been addressed, so from my view it may be merged.

PR #1337, opened 2023-04-03
feat: Use brotli

jonandernovella:
This PR closes https://github.com/MetabolicAtlas/private-issues/issues/22

**Type of change**  

- [ ] Other
 
**List of changes made**  

- Updated dockerfiles to use brotli plugin
- Added build and installation scripts

**Testing**  

- Rebuild and restart nginx (dev and prod)

### DIFF
--- a/nginx/Dockerfile
+++ b/nginx/Dockerfile
@@ -1,3 +1,13 @@
-FROM nginx:stable-alpine
+FROM nginx:mainline-alpine as builder
+
+COPY build-nginx-modules.sh .
+RUN chmod +x build-nginx-modules.sh
+RUN ./build-nginx-modules.sh "brotli"
+
+FROM nginx:mainline-alpine
+COPY --from=builder /tmp/packages /tmp/packages
+COPY install-nginx-modules.sh .
+RUN chmod +x install-nginx-modules.sh
+RUN ./install-nginx-modules.sh
 
 COPY nginx.conf /etc/nginx/nginx.conf
--- a/nginx/build-nginx-modules.sh
+++ b/nginx/build-nginx-modules.sh
@@ -1,0 +1,53 @@
+# Taken from https://raw.githubusercontent.com/nginxinc/docker-nginx/master/modules/Dockerfile.alpine
+set -ex
+
+apk update
+apk add linux-headers openssl-dev pcre2-dev zlib-dev openssl abuild \
+    musl-dev libxslt libxml2-utils make mercurial gcc unzip git \
+    xz g++ coreutils
+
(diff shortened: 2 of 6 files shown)


mihai-sysbio (commented):
Nice work!

The single acceptance criteria in the issue was to "Propose compression strategy to be used in the project", so since this PR is already implementing changes, I have a few discussion points before going further - perhaps easiest to take in a team discussion: 
- what is gain of brotli vs gzip
  - what do the numbers say?
- concerns about browser support:
  - are we going to drop performance for some users?
  - can multiple compression algorithms be used if one is not supported by the client?
- changes to the build process:
  - increase in build time?
  - increase in image size?
  - publishing our own base image with ghcr.io?
  - differences between `stable-` and mainline-`?
  - security implications?
- what other approaches have been considered:
  - Cloudfare?
  - any other compression service?
  - any other algorithm?

jonandernovella:
>     * what is gain of brotli vs gzip
>    * what do the numbers say?

 https://tech.oyorooms.com/how-brotli-compression-gave-us-37-latency-improvement-14d41e50fee4    

>     * concerns about browser support:
>       * are we going to drop performance for some users?
>       * can multiple compression algorithms be used if one is not supported by the client?

 I doubt there will be significant drops after looking at the benchmarks
>     * changes to the build process:
>       * increase in build time?
>       * increase in image size?
>       * publishing our own base image with ghcr.io?
The image is the same except it includes new features as well, plus the brotli binary
>       * differences between `stable-` and mainline-`?
The image is the same except it includes new features as well.
The official docs reccommend using mainline: https://www.nginx.com/blog/nginx-1-6-1-7-released/#Which-Version-Should-I-Use
>       * security implications?
Of what?
>     * what other approaches have been considered:
>       
>       * Cloudfare?
I considered that doing this on our end was better in case we decided dropping cloudflare in the future
>       * any other compression service?
No.
>       * any other algorithm?
Gzip was also tested (the previous solution)

perjo (approved):
Great work, LGTM

mihai-sysbio:
Notes from the discussion (feel free to edit):

Brotli generally performs better than gzip, and is more modern. For our users, it seems faster to use some compression. For more info, see [this](https://tech.oyorooms.com/how-brotli-compression-gave-us-37-latency-improvement-14d41e50fee4). There doesn't seem to be any standardized compression algorithm besides brotli.

In terms of browser support, [Brotli is well supported](https://caniuse.com/?search=brotli).

The compression part is slower than with gzip, but thanks to caching, our static assets will not incur a meaningful penalty.

Q: Are we using CPU time for compression?
Cloudfare seems to automatically run Brotli compression for us. Follow-up question: which assets?

Q: How does decompression compare for the users?
The decompression is run few times, so even if it would be a difference, it doesn't have a meaningful part.

PR #1339, opened 2023-04-04
Google Forms-based contact form

perjo:
This PR closes https://github.com/MetabolicAtlas/private-issues/issues/49

Embedded contact form on the platform page. Form responses are saved to a dedicated [Google Forms page](https://docs.google.com/forms/d/1_DvRfHqu1A9hzdb6M958lTINourAFIrNQlzDJCksIZY/edit).

**Type of change**  

**List of changes made**  

Addition of IFrame embedding contact form. 

**Screenshot of the fix**  

Skärmavbild 2023-04-04 kl  10 50 26

**Testing**  

1. Fill in form 
2. Verify it appears in the [Google Docs spread sheet](https://docs.google.com/spreadsheets/d/1creTLbJUK_JkkMn1x2VEBI932sp2RXPKrUzPv70AHOU/edit?resourcekey#gid=2025034565) for this form (you need to be logged in to your SciLifeLab account). 

**Further comments**  

### DIFF
--- a/frontend/src/components/about/Platform.vue
+++ b/frontend/src/components/about/Platform.vue
@@ -246,6 +246,11 @@
 
       <h3 id="contact-us" class="title is-3 mt-6">Contact us</h3>
 
+      <iframe
+        title="Contact Form"
+        src="https://docs.google.com/forms/d/e/1FAIpQLScnVFKd5OK9zIM9jR4qnnOr0QsZKZnhFql6saZuJaQYYsMs5g/viewform?embedded=true"
+        width="640" height="750" frameborder="0" marginheight="0" marginwidth="0">Loading…
+      </iframe>
       <p>
         For any questions, comments or suggestions, or if you find any bugs or vulnerabilities
         regarding Metabolic Atlas, please email us at
--- a/frontend/src/components/about/Platform.vue
+++ b/frontend/src/components/about/Platform.vue
@@ -252,17 +252,14 @@
           >contact [at] metabolicatlas [dot] org</a
         >, or use the contact form below.
       </p>
-
-      <iframe
-        title="Contact Form"
-        src="https://docs.google.com/forms/d/e/1FAIpQLScnVFKd5OK9zIM9jR4qnnOr0QsZKZnhFql6saZuJaQYYsMs5g/viewform?embedded=true"
-        width="640"
-        height="750"
-        frameborder="0"
-        marginheight="0"
-        marginwidth="0"
-        >Loading…
-      </iframe>
+      <div class="contact-form-wrapper">
+        <iframe
+          class="contact-form-iframe"
+          title="Contact Form"
+          src="https://docs.google.com/forms/d/e/1FAIpQLScnVFKd5OK9zIM9jR4qnnOr0QsZKZnhFql6saZuJaQYYsMs5g/viewform?embedded=true"
+          >Loading…
+        </iframe>
+      </div>
       <p>
         For any management related issues, such as collaborations or funding, please email us at
         <a href="mailto:management@metabolicatlas.org" target="_blank" rel="noopener noreferrer"
@@ -448,3 +445,22 @@ export default {
   },
 };
 </script>
+<style lang="scss" scoped>
+.contact-form-wrapper {
+  position: relative;
+  width: 100%;
+  overflow: hidden;
+  padding-top: 100%; /* 1:1 Aspect Ratio */
+}
+
+.contact-form-iframe {
+  position: absolute;
+  top: 0;
+  left: 0;
+  bottom: 0;
+  right: 0;
+  width: 100%;
+  height: 100%;
+  border: none;
+}
+</style>
--- a/frontend/src/components/about/Platform.vue
+++ b/frontend/src/components/about/Platform.vue
@@ -245,19 +245,18 @@
       </p>
 
       <h3 id="contact-us" class="title is-3 mt-6">Contact us</h3>
+      <p>
+        For any questions, comments or suggestions, or if you find any bugs or vulnerabilities
+        regarding Metabolic Atlas, please email us at
+        <a href="mailto:contact@metabolicatlas.org" target="_blank" rel="noopener noreferrer"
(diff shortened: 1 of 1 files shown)


jonandernovella (approved):
LGTM. I added a commit in order to make the google form responsive.

mihai-sysbio:
I'd like to review this as well.

mihai-sysbio:
Following up on some items from the _Acceptance criteria_ in the original issue:

- [ ] Form should have: name (optional), email (optional), description:
  - as noted previously, the description should be a required field
- [ ] Form should have spam protection
  - what is the conclusion here: is there some spam protection baked in? is there something from Cloudfare that could/should be used?
  - the form seem to have a fixed height (at least on mobile), which leads to an unnecessary scrollable area for the form (while the page is obviously also scrollable)
  - will the form work if the user disables "external" JS, i.e. not originating from the same domain? if not, a note should be added
  - in terms of page layout, it would somehow feel more consistent with the webpage if the width would be the same as that of the text above the form; what do others think?
  - it seems the loading happens only when the iframe is loaded in the html, which is great
  
Other feedback:
- [ ] while the green color chosen is the primary of MetAtlas, this color is sparingly used for buttons; has any other project-defined color been tried? there is a vars.css with these
- [ ] the horizontal green line in the form is just for styling, but it's inconsistent with the webpage in terms of how strong that color is; has any other color been tried?
- [ ] it would be more user-friendly if the default text in the form would be more verbose (e.g. please share your email to allow us to reach back to you with follow-up questions)
- [ ] the "Contact us" text should include a link that pops up the form in an external tab
- [ ] the form description should be used
- [ ] the name "Contact form" sounds duplicate to the section's "Contact us", try something else e.g. "Share your feedback"
- [ ] in the form settings, email notifications should be turned on
- [ ] the form should be shared in the right folder in GDrive
- the form ownership should be transferred to me

perjo:
> * [ ]  Form should have: name (optional), email (optional), description:
>   
>   * as noted previously, the description should be a required field

Fixed now!

> * [ ]  Form should have spam protection
>   
>   * what is the conclusion here: is there some spam protection baked in? is there something from Cloudfare that could/should be used?

Good question. According to [this answer](https://support.google.com/docs/thread/204773711?hl=en&msgid=205012024), Google Forms uses reCAPTCHA to avoid spam. Looking at it now I am not sure though, since there is no reCAPTCHA logo in the form. This is not well documented and hard to find information about since the behaviour seems to have changed over the years (first there was no CAPTCHA in Google Forms, then it was opt-in, later opt-out and now it seems it can't be turned off). It might be the case the reCAPTCHA is only on if you turn on the _collect email addresses_ option. We do not want this as it makes the email field required. Do not know if we can test this locally. 

> * [ ]  The changes are tested on multiple models and browsers
>   
>   * the form seem to have a fixed height (at least on mobile), which leads to an unnecessary scrollable area for the form (while the page is obviously also scrollable)

Yes, an IFrame needs a fixed height set (this will apply to any IFrame-based solution). Originally this was set to 750 px but was removed in the [last commit](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1339/commits/25e403b2a46e6d88397313e494709999920edea7) which should then give a [default height of 150 px](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#attributes). This seems to still work and the browser will expand the IFrame to fit the content, at least when I test it. To sum up: We can not avoid different scrollable areas, but the scroll bar should almost never need to be there. 

>   * will the form work if the user disables "external" JS, i.e. not originating from the same domain? if not, a note should be added

An IFrame is its own browsing context so JS executed within this is not considered cross-origin. I do not know if there is such a setting in modern browsers, but the proposed solution works in Chrome and Firefox even is security level is set to the strictest option. 

>   * in terms of page layout, it would somehow feel more consistent with the webpage if the width would be the same as that of the text above the form; what do others think?

I do not think we can change this. Even if the IFrame would take full width, the form itself (from Google) is this width. 

> * [ ]  while the green color chosen is the primary of MetAtlas, this color is sparingly used for buttons; has any other project-defined color been tried? there is a vars.css with these
> * [ ]  the horizontal green line in the form is just for styling, but it's inconsistent with the webpage in terms of how strong that color is; has any other color been tried?

There is a main theme color for the form, which will set the color of the top border and the submit button. I originally used the color of the "hero" section for the home page and GotEnzymes page (#25543c). Changed this now to the project link color (#00549e). This is easily changed in Google Forms. 

> * [ ]  it would be more user-friendly if the default text in the form would be more verbose (e.g. please share your email to allow us to reach back to you with follow-up questions)

It seems the placeholder text can't be changed but added a field description stating this

> * [ ]  the "Contact us" text should include a link that pops up the form in an external tab

Will be simple to fix, will implement if going further with this solution

> * [ ]  the form description should be used

Added a text which (easily changed in Google Forms)

> * [ ]  the name "Contact form" sounds duplicate to the section's "Contact us", try something else e.g. "Share your feedback"

Changed this

> * [ ]  in the form settings, email notifications should be turned on

Turned this on 

> * [ ]  the form should be shared in the right folder in GDrive

Where should it be? 

> * the form ownership should be transferred to me

Done!

mihai-sysbio:
super neat follow-up @perjo 🦄 I really appreciate it!

The answers are very clear to me. I will take it upon myself to expand on the text in the Google Form even further, and will ask everyone to proof it. A neat benefit of this GForms implementation is that the form can be changed without needing any deployments.

> > * the form should be shared in the right folder in GDrive
> 
> Where should it be?

The is a `Metabolic Atlas` folder, I think under `Maple` (feel free to move it outside this folder).

> will implement if going further with this solution

Good point actually. I don't recall a formal agreement that GForms will be the solution. My impression was that this seemed good enough (from a PO view) and easiest (least coding). We should touch base with the team to see if there are other opinions @jonandernovella @Jessica-Sandler @pdmnyberg @KattisLej

mihai-sysbio:
> will take it upon myself to expand on the text in the Google Form even further

Done! Please review.

mihai-sysbio:
Thanks fixing the spaces @perjo. There was a typo in there "shaking" -> "sharing", now fixed. From my side, feel free to merge 💯

PR #1344, opened 2023-04-06
feat: Add Global search button

### LINKED ISSUE #918: Improve discoverability of global search from gem search

As a user, I want to find global search more conveniently from gem search so that I can easier discover the global search page.

**Acceptance Criteria**

**Additional context**

- [ ] The current hint ("Limited to 50 results per type. Click here to search all integrated GEMs") could be made more obvious. Similar to the button for when no matches are found.
- [ ] Clicking "Enter/Return" on the keyboard should perform a global search with the current search term.

jonandernovella:
Fixes #918 

This button is now added to the GEM search section and gives access to the global search feature.

## Screenshots
image
image


pdmnyberg (changes requested):

pdmnyberg (commented):
While I think this solves the problem of making the global search easier to find I am not sure I like the way it affects the space available in the mobile view. Could we find a way that still maintains the old size of the text-input?

I think the issue had some suggestions where the button was show as a `part` of the text-input where I would expect it to disappear when selecting the input or when typing starts.

jonandernovella:
> While I think this solves the problem of making the global search easier to find I am not sure I like the way it affects the space available in the mobile view. Could we find a way that still maintains the old size of the text-input?
> 
> I think the issue had some suggestions where the button was show as a `part` of the text-input where I would expect it to disappear when selecting the input or when typing starts.

Pros with this solution:
- Global search button always shown, great for accessibility

Cons
- Shrunk input field

-----

Pros with suggested solutions in the issue
- Better use of space

Cons
- Limited accessibility: button won't be accessible when user is typing/has typed
- Button inside an input field that was designed for something else. Could confuse the user.

frontend/src/components/explorer/gemBrowser/GemSearch.vue
```diff
@@ -1,6 +1,17 @@
 <template>
   <div v-if="model" id="gem-search-wrapper">
     <div class="field has-addons m-0">
+      <a
+        href="/search"
+        id="globalSearchButton"
+        class="button is-rounded is-outlined is-success"
+        title="Global GEM search"
+      >
```
perjo, line 9: Surely we should use a `routerLink` instead of an ordinary `a` element to avoid full page reload?

jonandernovella:
@perjo and I did pair programming and we did some accessibility improvements:

- No page reload when clicking on global search
- Accessibility labels
- Moved button to the right end
- Improved quick search notification text about global search

perjo (approved):
Looks good (was pair programming)

pdmnyberg (approved):

PR #1351, opened 2023-04-14
feat: add overlay status indicator panel

### LINKED ISSUE #975: indicator when data overlay is applied

As a user, I want an indication when data overlay is applied so that I understand what is shown on the map.

**Acceptance Criteria**

- [ ] there are bonus points if the implementation is applied to the Interaction Partners (not a hard requirement)
- [ ] An indicator is shown when overlay is applied:
  - one idea is the button group in the map viewer area (same for 2D/3D)

**Additional context**

 [Example](https://github.com/CreativeBulma/bulma-badge/). See also #935

pdmnyberg:
This PR closes #975

Adds a panel indicating the overlay data status

**Type of change**  

**List of changes made**  
- Add an indicators getter to dataOverlay
- Add an indicator panel component
- Add indicator panel to mapviewer

**Screenshot of the fix**  
Screen Shot 2023-04-13 at 18 00 11

**Testing**  
- Go to: http://localhost/explore/Human-GEM/map-viewer/acylglycerides_metabolism?dim=2d&panel=0&sel=&search=&coords=-1973.35,-3529.21,0.18,0,0,500&dataTypes=gene,reaction&dataSources=hpaRna.tsv,HPA_single-cell_reactions.tsv&dataSets=adrenal%20gland,adipose%20tissue_fibroblasts_0
- Look for the panel in the lower right corner
- Change the overlays and verify that it is updated

### DIFF
--- a/frontend/cypress/e2e/homepage.cy.js
+++ b/frontend/cypress/e2e/homepage.cy.js
@@ -37,4 +37,22 @@ context('Homepage', () => {
     })
   })
 
+  describe('DataOverlay', () => {
+    const targets = [
+      {
+        url: '/explore/Human-GEM/map-viewer/acyl_coa_hydrolysis?dim=2d&panel=1&sel=&search=&coords=-3034.36,-3215.63,0.25,0,0,500&dataTypes=gene&dataSources=hpaRna.tsv&dataSets=adipose%20tissue',
+        indicator: 'gene'
+      },
+      {
+        url: '/explore/Human-GEM/map-viewer/acyl_coa_hydrolysis?dim=2d&panel=1&sel=&search=&coords=-3034.36,-3215.63,0.25,0,0,500&dataTypes=reaction&dataSources=HPA_single-cell_reactions.tsv&dataSets=adipose%20tissue_adipocytes_11',
+        indicator: 'reaction'
+      }
+    ]
+    for (const {url, indicator} of targets) {
+      it(`It has the ${indicator} indicator`, () => {
+        cy.visit(url)
+        cy.get('.map-indicators .indicator-id').contains(indicator)
+      })
+    }
+  })
 })
--- a/frontend/src/components/explorer/MapViewer.vue
+++ b/frontend/src/components/explorer/MapViewer.vue
@@ -96,6 +96,7 @@
             @un-select="unSelect"
             @update-panel-selection-data="updatePanelSelectionData"
           />
+          <IndicatorPanel class="map-indicators" :indicators="dataOverlayIndicators" />
           <ErrorPanel
             :message="loadMapErrorMessage"
             :hide-error-panel="(loadMapErrorMessage = '')"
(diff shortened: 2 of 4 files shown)


jonandernovella (changes requested):
Looks great! I think we would benefit from having:

- A cypress test
- More descriptive name for:
  * the indicator panel component. I'd suggest DataOverlayIndicator
  * the groupsWithItems function could convey better what's really returning

pdmnyberg:
> Looks great! I think we would benefit from having:
> 
>     * A cypress test
> 
>     * More descriptive name for:
>       
>       * the indicator panel component. I'd suggest DataOverlayIndicator
>       * the groupsWithItems function could convey better what's really returning

I do not agree about the naming of the IndicatorPanel. I think it gets overly specific.

jonandernovella:
> > Looks great! I think we would benefit from having:
> > ```
> > * A cypress test
> > 
> > * More descriptive name for:
> >   
> >   * the indicator panel component. I'd suggest DataOverlayIndicator
> >   * the groupsWithItems function could convey better what's really returning
> > ```
> 
> I do not agree about the naming of the IndicatorPanel. I think it gets overly specific.

It should be specific in this case, I think. Otherwise it's very difficult to guess what this component is about and/or find it. Besides, if it shall be more generic that may be done at a later stage, at a later increment

pdmnyberg:
> > > Looks great! I think we would benefit from having:
> > > ```
> > > * A cypress test
> > > 
> > > * More descriptive name for:
> > >   
> > >   * the indicator panel component. I'd suggest DataOverlayIndicator
> > >   * the groupsWithItems function could convey better what's really returning
> > > ```
> > 
> > 
> > I do not agree about the naming of the IndicatorPanel. I think it gets overly specific.
> 
> It should be specific in this case, I think. Otherwise it's very difficult to guess what this component is about and/or find it. Besides, if it shall be more generic that may be done at a later stage, at a later increment when there's a use case for it.

I am still not convinced that making such specific components is a great approach. It makes it hard to know if reuse is possible/relevant. I think the component is easily found since it has a proper class in the DOM and it is referenced from the MapViewer component.

I am ok with renaming it but since we seem to disagree I would like a second opinion on this . @perjo @KattisLej

KattisLej (approved):
I think it looks very good and makes it easier to see. My only thought is the placement,  I am thinking if the right upper corner would make it more visible at least on larger screens. If you already tried that out and it didn't work, don't mind my comment.

As for the naming I think IndicatorPanel is a good name, I also think the other name suggestion could absolutely work but I don't think it's crucial to the understanding of the code.

perjo:
> > > >   * the indicator panel component. I'd suggest DataOverlayIndicator

So, the question is whether this is a general IndicatorPanel or a specific DataOverlayIndicator(Panel). I do not really have a strong preference. I myself would probably have used something like Jons proposal, but then I also have a feeling that I might often use too specific names...

jonandernovella:
> > > > > * the indicator panel component. I'd suggest DataOverlayIndicator
> 
> So, the question is whether this is a general IndicatorPanel or a specific DataOverlayIndicator(Panel). I do not really have a strong preference. I myself would probably have used something like Jons proposal, but then I also have a feeling that I might often use too specific names...

We discussed the component naming and decided the context of the component (folder location) is good enough for understanding the component.

jonandernovella (approved):

PR #1355, opened 2023-04-18
Update team and previous contributors

### LINKED ISSUE #1291: update about-development section

As a person interested in the team, I want to see an updated list of persons in a compact way.

**Acceptance Criteria**

- [ ] add Thomas Svensson to the advisors list
- [ ] update current development team (start with a placeholder while waiting for approval)
- [ ] update previous contributors (Maple)
- [ ] compact cards on previous contributors (hard requirement): aiming for 30%-75% of the current surface area

perjo:
This PR closes #1291 

**Type of change**  

- [x] Data update
 
**List of changes made**  

- Updated members of team and previous contributors
- Made previous contributors cards smaller
- Extracted person info to its own file
- Extracted cards to components (there is more that could be done here. I thought this could suffice, but feel free to suggest improvements) 

**To be done**

- [ ] Update info about team members 

**Screenshot of the fix**  

**Testing**  

- Instructions on how to test

**Further comments**  

- [ ] I have performed a self-review of my own code and commented any hard-to-understand areas


pdmnyberg:
Now that we are updating this. Could I suggest that we use a `role` property on each `member/person` instead of separating the data into different sets? Dynamically filtering the members should not cause any measurable slowdown but would reduce the size of a diff when the team or a single team member move in/out of rotation. What do you think @perjo

perjo:
> Now that we are updating this. Could I suggest that we use a `role` property on each `member/person` instead of separating the data into different sets? Dynamically filtering the members should not cause any measurable slowdown but would reduce the size of a diff when the team or a single team member move in/out of rotation. What do you think @perjo

Yes, that could be a good idea. The only downside I see now that we do not use typescript is the lack of type safety for the new `role` property (could be prone to typos) but I guess this is nicer anyway.

perjo:
Thinking of extracting components for the person cards. To me that feels natural, what do you think?

pdmnyberg:
> Thinking of extracting components for the person cards. To me that feels natural, what do you think?

I think that makes sense. Would probably make the Platform component a bit cleaner. :)

perjo:
> > Now that we are updating this. Could I suggest that we use a `role` property on each `member/person` instead of separating the data into different sets? Dynamically filtering the members should not cause any measurable slowdown but would reduce the size of a diff when the team or a single team member move in/out of rotation. What do you think @perjo
> 
> Yes, that could be a good idea. The only downside I see now that we do not use typescript is the lack of type safety for the new `role` property (could be prone to typos) but I guess this is nicer anyway.

Hmm...also the order of the members is determined by the order in the array and that will not be as easily perceived if they are all in one array. Was thinking one could order the members alphabetically but I guess they should not be since the professor is ordered first.

pdmnyberg:
> Hmm...also the order of the members is determined by the order in the array and that will not be as easily perceived if they are all in one array. Was thinking one could order the members alphabetically but I guess they should not be since the professor is ordered first.

Fair point.  If there's an editorial decision needed for the ordering then it might be hard to achieve a nice data set without making it harder to understand? Or maybe it would make more clear that the order in intentional if we have some kind of ordering parameter more than alphabetical order. If it gets to complicated I guess we it might be a good idea to delay/skip this idea.

perjo:
> Fair point. If there's an editorial decision needed for the ordering then it might be hard to achieve a nice data set without making it harder to understand? Or maybe it would make it more clear that the order in intentional if we have some kind of ordering parameter more than alphabetical order. If it gets to complicated I guess we it might be a good idea to delay/skip this idea.

I am leaning towards keeping the original structure. An index/order parameter would just make it overly complicated I think.

pdmnyberg:
> > Fair point. If there's an editorial decision needed for the ordering then it might be hard to achieve a nice data set without making it harder to understand? Or maybe it would make it more clear that the order in intentional if we have some kind of ordering parameter more than alphabetical order. If it gets to complicated I guess we it might be a good idea to delay/skip this idea.
> 
> I am leaning towards keeping the original structure. An index/order parameter would just make it overly complicated I think.

Agreed. Thank you for looking into it :)

pdmnyberg:
I really like how you structured `personnel.js` @perjo :)

mihai-sysbio:
Some quick thoughts:
- it's useful to keep images small in resolution for faster transfer, and also in equal size
- there is a `content` folder that imho would be a better fit for `personnel.js`, especially if it could be turned into a plain iterable json rather than an object array
- wherever possible, I would opt to restrict the personnel fields to be unique strings rather than full links (e.g., only orcid id, and the link to be built by the component)
- the cards seem to be so aimed at persons that `PersonCard` might be more descriptive
- without looking into the code, my concern about Large and Small card components is if they adding vs reducing maintainability

perjo:
Discussion: 

> * there is a `content` folder that imho would be a better fit for `personnel.js`, especially if it could be turned into a plain iterable json rather than an object array

Ok, did not realize that. I personally very much prefer a [_folder/package-by-feature_](https://www.techyourchance.com/popular-package-structures/)  structure and since `personnel.js` only is only used by the `Platform.vue` component you could argue that it belongs in the same folder as this file (the other files in content is used by multiple files). Still, since this is just data I am open to the idea of keeping it together with other similar files. What do the rest of the team think? @MetabolicAtlas/developers 

Regarding the structure used as of now, it seems the project is mostly folder by type (`api`, `components`, `content`, `helpers`, `layouts` etc) but the `components` subfolder is organized by feature. I personally would like to go towards to a more consistent folder by feature structure. 

It seems Vue does not officially prescribe any type of structure. I could be biased coming from Angular (which [style guide clearly states that a folder-by-feature structure should by used](https://angular.io/guide/styleguide#folders-by-feature-structure)) and Java (in which [package-by-feature](http://www.javapractices.com/topic/TopicAction.do?Id=205) is clearly superior, thanks to the presence of package private visibility) but if there are no technical reasons to folder by type (like there were in next.js), I do not see why we should keep a structure with gives less cohesion and more focus implementation details. 

> * wherever possible, I would opt to restrict the personnel fields to be unique strings rather than full links (e.g., only orcid id, and the link to be built by the component)

That is a very good idea, will change that. 

> * the cards seem to be so aimed at persons that `PersonCard` might be more descriptive

Yes, it is not a generic card but specialized in presenting persons. My first idea was that it was scoped to the Platform.vue component and there are no other cards there, but I realize now that there are many components in the same folder. Will change this. 

> * without looking into the code, my concern about Large and Small card components is if they adding vs reducing maintainability

Could you elaborate further? This change removed duplication and split a very large component. How would that reduce maintainability?

perjo:
Perhaps make a `personnel` subfolder to `about` and put `personnel.js` and card components in there, would that be a good idea? (won't change or move `personnel.js` until you're done editing it. @MetabolicAtlas/developers

pdmnyberg:
I think I agree about the `LargeCard` and `SmallCard`. Those component share a lot of code and structure and thus I think it would make sense to instead use a `PersonCard` with a `size` option maybe?

When it comes to where to put the `personnel.js` file I think `data` could make sense as well but in that case I would be tempted to create a `store/modules/personnel.js` which would be the supplier of the data making sure to load `data/personnel.js` at some point in the code. But I guess that might be going a bit too far in regards to what i we actually want to achieve here.

KattisLej (approved):
Looks very nice, the structure is better, the small cards look good and seems reasonable to have them smaller.

haowang-bioinfo:
just noticed that my GH account should be updated to the latest, should I just edit by myself?

jonandernovella:
The links to the NBIS website might soon need to be updated after the migration. @mihai-sysbio Is it a requirement to keep them?

perjo:
Fixed: 

1. Created a generic `PersonCard` component with size attribute instead of two different components
2. Only store `orcid` in personnel.js

Not moved pesonnel.js, this could be deferred to new issue.

Jessica-Sandler (approved):
Looks nice! Good structure with the PersonCard component and personnel.js

Issue #1328, opened 2023-03-24
Accessing Human-Gem from URL is broken


KattisLej:
**Describe the bug**
Going to the URL directly for Human-Gem doesn't work in main:
http://localhost/gems/repository/Human-GEM

Compare with production:
https://metabolicatlas.org/gems/repository/Human-GEM

It's possible to access the repository by clicking on Human-Gem on main as before

**Steps to reproduce**
Got to URL http://localhost/gems/repository/Human-GEM
![image.png](https://images.zenhubusercontent.com/627bbee5c0197a3932d6815e/342b44a8-d517-48df-aac4-8d0eef0b4008)

**Expected behavior**
Expected behaviour in https://metabolicatlas.org/gems/repository/Human-GEM:
![image.png](https://images.zenhubusercontent.com/627bbee5c0197a3932d6815e/df8dce0a-efc8-411c-adf3-901dff837f63)

**Additional context**
Was found in another issues PR #1304 but has not been introduced by that change bc it is reverted.
[Comment of bug](https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1304/files#r1135435985)

mihai-sysbio:
Closing as duplicate.

PR #1362, opened 2023-04-21
fix: load initial state of gems repository properly

### LINKED ISSUE #1326: Directly navigating to `/gems/repository/<model>` will fail to load the correct state
**Describe the bug**
Links to `/gems/repository/<model>` do not load properly.  When going directly to `/gems/repository/<model>` the component throws a javascript error (`Uncaught (in promise) TypeError: $data.selectedModel is null`) and fails to load properly. This only happens when navigating via the browser address bar.

This error also prevents cypress tests for `/gems/repository/<model>` from working properly. As described here: https://github.com/MetabolicAtlas/MetabolicAtlas/pull/1325#discussion_r1145459545

**Steps to reproduce**
- Open your browser and show terminal
- Load `http://localhost/gems/repository/Fruitfly-GEM` from address bar
- Verify that the error shows up in the console
- Verify that no `modal` with model information in showing

**Expected behavior**
- [ ] No error is thrown by the responsible component when loading `/gems/repository/<model>` directly
- [ ] The cypress test for `/gems/repository/<model>` works as expected

pdmnyberg:
This PR closes #1326 

This PR tries to clean up some of the state management related to models and gems in order to make it easier to load an initial state without fail in the `Repository` component.

**Type of change**  

**List of changes made**  
- Adjust `gems`-store
- Adjust `models`-store
- Add compatibility fixes for other components
- Clean up state of `Repository` component
- Add compatibility fixes for `Explorer` component
- Rename old `selectModel` to `trySelectModel`

**Testing**  
- In the folder `frontend`, run `npx cypress run cypress`
- All tests should pass
- Go to http://localhost/gems/repository/Human-GEM
- The page should load properly and show the popup described in #1328

### DIFF
--- a/api/src/neo4j/queries/idInModels.js
+++ b/api/src/neo4j/queries/idInModels.js
@@ -46,6 +46,7 @@ RETURN { identifier: properties(r), components: COLLECT(DISTINCT({component: r, 
     if (dbName === 'MetabolicAtlas') {
       identifier.dbName = dbName;
       identifier.externalId = identifier.id;
+      identifier.url = `https://identifiers.org/metatlas:${identifier.id}`;
     }
 
     return { components, identifier };
--- a/api/src/swagger/config.yaml
+++ b/api/src/swagger/config.yaml
@@ -607,12 +607,12 @@ paths:
           in: 'path'
           required: true
           type: 'string'
-          description: 'the name of the database, e.g. **Protein Atlas**'
+          description: 'the name of the database, e.g. **Protein Atlas** or **MetabolicAtlas**'
         - name: 'externalId'
           in: 'path'
           required: true
           type: 'string'
-          description: 'the external ID, e.g. **ENSG00000120915**'
+          description: 'the external ID, e.g. **ENSG00000120915** or **MAR02279** for *Protein Atlas* and *MetabolicAtlas* respectively'
       responses:
         '200':
           description: 'Successful query'
--- a/api/test/identifier.test.js
+++ b/api/test/identifier.test.js
@@ -10,4 +10,26 @@ describe('identifier', () => {
 
     expect(models).toEqual(sortedModels);
   });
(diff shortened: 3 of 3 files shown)


pdmnyberg:
After some time to reflect I would suggest adding a number of jest-tests in order to verify functionality of this before merging. At least testing the `vuex` modules

jonandernovella:
So there are two main things I learned after reviewing the PR:

1- There are a number of changes in this PR about refactoring the use of vue state with better names etc.
2- The error described in https://github.com/MetabolicAtlas/MetabolicAtlas/issues/1326 seems to be an internal Vue error which only seems to happen in local development and when navigating using the address bar. Possibly some hot reload issue.
3. It is unclear to me what solves the above issue in local development and whether we should actually try to fix it ourselves or instead send in a ticket to the Vue project

KattisLej (approved):
Looks good to me, well done.

pdmnyberg:
> So these are the main things I learned after reviewing the PR:
> 
>     1. There are a number of changes in this PR about refactoring the use of vue state with better names etc.
> 
>     2. The error described in [Directly navigating to `/gems/repository/<model>` will fail to load the correct state #1326](https://github.com/MetabolicAtlas/MetabolicAtlas/issues/1326) seems to be an internal Vue error which only seems to happen in local development and when navigating using the address bar. Possibly some hot reload issue.
> 
>     3. It is unclear to me what solves the above issue in local development and whether we should actually try to fix it ourselves or instead send in a ticket to the Vue project

I think the main problem is that the original state management is pretty unclean duplicating some state elements and making it hard to assess where the actual problem lies. I think there might be a timing issue due to the local-dev-api generally being slower to respond than the prod-api. I can try to reduce the number of changes when I get back. But I do think this change is necessary while I think it might be reasonable to hold off on merging this branch atm.

Thank you for your feedback @jonandernovella

KattisLej:
Converting to draft bc of the conversation in the comments

PR #1366, opened 2023-04-24
Validate API input to prevent injection

### LINKED ISSUE #1317: API is open to Cypher injection
**Describe the bug**

The API is open to Cypher injection. 

**Steps to reproduce**

**Simple example**: Enter `HumanGem)-[]-(x`  as model in the [compartment API](http://localhost/api/v2/#/Compartments/compartmentInfo) (or visit http://localhost/api/v2/compartments/golgi_apparatus?model=HumanGem)-%5B%5D-(x&full=true
 directly. Now we've not only inserted data into the Cypher query, we have also changed the query that is executed. 

**More advanced example**: Enter `HumanGem)-[]-(r)-[]-(s)-[]-(ss)-[]-(cm)-[]-(g)-[]-(e)-[]-(csvg) // ` as model in the [compartment API](http://localhost/api/v2/#/Compartments/compartmentInfo)  (or visit http://localhost/api/v2/compartments/golgi_apparatus?model=HumanGem)-%5B%5D-(r)-%5B%5D-(s)-%5B%5D-(ss)-%5B%5D-(cm)-%5B%5D-(g)-%5B%5D-(e)-%5B%5D-(csvg)%20%2F%2F&full=true
 directly). In this example we inject code and comment out other parts of the query. This request times out in my environment but that is not really relevant. 

These examples could of course be tested with the production API as well. A simple sign of that is that if one requests https://metabolicatlas.org/api/v2/compartments/golgi_apparatus?model=%22 we get an error message directly from Neo4J not being able to parse the query. 

It is not only the compartments endpoint that is affected, it seems most of them except the GotEnzymes ones are. 

**Expected behavior**

The examples above should return error codes 400 or 404. 

**Disclaimer**: This issue is public after team and project management discussions since cypher injection issue is already disclosed in for example https://github.com/MetabolicAtlas/MetabolicAtlas/pull/557

perjo:
This PR closes #1317 .

**Type of change**  

**List of changes made**  

Validate all parameters passed to Neo4J. 

**Screenshot of the fix**  

**Testing**  

1. Test the examples in the issue. Expected: 400 Bad Request with error message ("Malicious char detected")
2. Make any other API call and use single quotes or double quotes in any of the params. Expected: 400 Bad Request with error message ("Malicious char detected"). The API should **never** give a Neo4J parse error in the response ( i.e. something like

> Invalid input '{': expected whitespace, comment, '.', node labels or rel types, '[', '^', '*', '/', '%', '+', '-', "=~", IN, STARTS, ENDS, CONTAINS, IS, '=', "<>", "!=", '<', '>', "<=", ">=", AND, XOR, OR, ',' or ')' (line 3, column 51 (offset: 74)) " MATCH (cs:CompartmentState)-[]-(c:Compartment:" {id: 'golgi_apparatus'})" ^

)

**Further comments**  

The validation can be optimized a little bit (i.e. by using a regular expression) but this is so small that I felt it is not worth the escaping-hell and low readability of RegExps. 


api/src/endpoints/neo4j.js
```diff
@@ -24,9 +24,24 @@
...
 
+function validatePayload(payload) {
+  for (const value of Object.values(payload)) {
+    validateInput(value);
+  }
+}
+
+function validateInput(value) {
+  MALICIOUS_CHARACTERS.forEach(malicious_char => {
+    if (value && value.includes(malicious_char)) {
```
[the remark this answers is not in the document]
perjo: I would like dismiss this as false positive as the rows above makes sure that is in fact not an array, but I am unsure of the process we use for these alerts. Maybe it is better that reviewers make these calls.

api/src/endpoints/neo4j.js
```diff
@@ -143,23 +162,36 @@ neo4jRoutes.get('/3d-network', async (req, res) => {
...
     const network = await get3dNetwork(payload);
     res.json(network);
   } catch (e) {
     res.status(400).send(e.message);
   }
 });
 
+function validateJson(json) {
+  if (typeof json === 'object') {
+    Object.values(json).forEach(val => validateJson(val));
```
jonandernovella, line 176: I assume that here you wanted to call validateInput() ?

api/src/endpoints/neo4j.js
```diff
@@ -191,11 +223,16 @@ neo4jRoutes.get('/identifier/:dbName/:externalId', async (req, res) => {
   const { referenceType } = req.query;
 
   try {
+    validateInput(dbName);
+    validateInput(externalId);
+    validateInput(referenceType);
```
jonandernovella, line 228: This looks a bit ugly. What about an iterative control statement?

jonandernovella (changes requested):
Great work. The tests seem reasonable and they do pass most of the times.

I have pointed at some possible issues with my comments.


In addition to that I found a new file called frontend/cypress.json which was probably commited by mistake.

api/src/endpoints/neo4j.js
perjo: No, I think I meant it to be recursive.

api/src/endpoints/neo4j.js
perjo: Hmm...it may be a matter of taste but I do not see what is ugly about it. How would you like to re-write it?

api/src/neo4j/queries/idInModels.js
```diff
@@ -1,5 +1,6 @@
 import querySingleResult from 'neo4j/queryHandlers/single';
 import { crossReferencesDict } from 'data/identifiers';
+import db from '../../gotEnzymes/db';
```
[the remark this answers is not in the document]
perjo: Not intentional, should be removed.

api/src/endpoints/neo4j.js
perjo: Of course, if you pass a circular structure to this function it will not terminate but JSON structures can't be circular. I do not know if this should be considered a problem.

api/src/endpoints/neo4j.js
jonandernovella: I was thinking of iterating using forEach. But if you think it's fine then OK!

KattisLej (approved):
Looks good and the response is as expected

